PR #23, opened 2022-08-18
Download all plugins from repository

eigengravy:


app/src/main/java/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -257,6 +259,26 @@ object AppUtils {
...
+            downloadAllPluginsDialog(url)
+        }
+    }
+
+    fun Activity.downloadAllPluginsDialog(repositoryUrl: String) {
+        runOnUiThread {
+            val context = this
+            val builder: AlertDialog.Builder = AlertDialog.Builder(this)
+            builder.setTitle(
+                "Download all plugins from this repo?"
```
Blatzar, line 271: R.string :skull:

Blatzar: Probably good to include the repo name if that's available without another request :thinking:

eigengravy: Done 🚀

eigengravy:
i certainly did not

LagradOst:
You can use Download and Cancel instead of yes and no. that is more readable for the user impo

eigengravy:
Updated it

PR #48, opened 2022-08-21
[Feature] Copy settings values from Preferences to providers APIs

Jacekun:
ready for merge.

Reason for PR: Allow extensions to behave differently based on the values of settings on Main App. Can be extended to include other settings without changing it much.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
@@ -367,7 +367,7 @@ abstract class MainAPI {
         }
     }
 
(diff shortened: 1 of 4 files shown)


LagradOst:
Dude what? just use a companion object for this

app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -356,13 +360,16 @@ abstract class MainAPI {
         var overrideData: HashMap<String, ProvidersInfoJson>? = null
     }
 
-    fun init() {
+    fun init(settingsForProvider: SettingsJson) {
```
LagradOst, line 363: companion object

LagradOst (changes requested):

PR #58, opened 2022-08-24
Added Zorofile (extractor)

hexated:


app/src/main/java/com/lagradost/cloudstream3/extractors/Zorofile.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.lagradost.cloudstream3.utils.ExtractorApi
+import com.lagradost.cloudstream3.utils.ExtractorLink
+import com.lagradost.cloudstream3.utils.M3u8Helper
+
+class Zorofile : ExtractorApi() {
+    override val name = "Zorofile"
+    override val mainUrl = "https://zorofile.com"
+    override val requiresReferer = true
+
+    override suspend fun getUrl(url: String, referer: String?): List<ExtractorLink> {
```
LagradOst, line 17: No need to change, but be aware that the new api you should use is https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/ExtractorApi.kt#L390

LagradOst (commented):
👍

hexated:
ready to merge

PR #68, opened 2022-08-29
Made Language into provider

KillerDogeEmpire:


C10udburst:
maybe you should remove all translations of `category_preferred_media_and_lang`

KillerDogeEmpire:
True, I will do that

KillerDogeEmpire:
True, I will do that

KillerDogeEmpire:
I will do that when I have time. I think I can do it now

PR #69, opened 2022-08-29
Add audio and video track support

Blatzar:
Missing preferred quality 

@LagradOst Please check


app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -106,6 +113,8 @@ class GeneratorPlayer : FullScreenPlayer() {
 
     override fun embeddedSubtitlesFetched(subtitles: List<SubtitleData>) {
```
LagradOst, line 114: Might be more valid to have 2 different callbacks or one callback called onEmbbededDataFetched or smth

LagradOst: other than that, everything looks fine. if you feel like this is not an issue this can be merged now

PR #70, opened 2022-08-29
[Refactor] Override domainUrl on select Extractor for changing hostname from extensions

Jacekun:
Ready for merge.

Some providers uses extractors but with different hostname to work.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/extractors/XStreamCdn.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/extractors/XStreamCdn.kt
@@ -42,7 +42,7 @@ open class XStreamCdn : ExtractorApi() {
     override val name: String = "XStreamCdn"
     override val mainUrl: String = "https://embedsito.com"
     override val requiresReferer = false
-    open var domainUrl: String = "embedsito.com"
+    override var domainUrl = "embedsito.com"
 
     private data class ResponseData(
         @JsonProperty("file") val file: String,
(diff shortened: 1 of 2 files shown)


Blatzar:
might as well make mainurl overridable then?

Jacekun:
> might as well make mainurl overridable then?

hmm idk. Maybe? DomainUrl is mainly used by XstreamCdn extractor AFAIK.

Blatzar:
> > might as well make mainurl overridable then?
> 
> hmm idk. Maybe? DomainUrl is mainly used by XstreamCdn extractor AFAIK.

I think they do that to sidestep making new classes, tbh we need an entirely new system for that.

Jacekun:
> > > might as well make mainurl overridable then?
> > 
> > 
> > hmm idk. Maybe? DomainUrl is mainly used by XstreamCdn extractor AFAIK.
> 
> I think they do that to sidestep making new classes, tbh we need an entirely new system for that.

Uh kinda. I'm the one who added domainUrl to XtreamCdn. Some extractor sites using XstreamCdn as base only works if the domain for the request is different from the mainUrl. Some uses the site url as domain for XstreamCdn

LagradOst:
> Uh kinda. I'm the one who added domainUrl to XtreamCdn. Some extractor sites using XstreamCdn as base only works if the domain for the request is different from the mainUrl. Some uses the site url as domain for XstreamCdn

I dont get why every api should have it, cant you just use a superclass or interface?

Jacekun:
> > Uh kinda. I'm the one who added domainUrl to XtreamCdn. Some extractor sites using XstreamCdn as base only works if the domain for the request is different from the mainUrl. Some uses the site url as domain for XstreamCdn
> 
> I dont get why every api should have it, cant you just use a superclass or interface?

You're right. It can definitely be accessed thru a superclass of it. I just made it so its easier to access. If you think its not worth the code change, I'll understand.

LagradOst:
Will be closed as stale and not really important

PR #71, opened 2022-08-29
[Feature] Add new extractor - PlayLtXyz

Jacekun:
ready for merge

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/extractors/PlayLtXyz.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/extractors/PlayLtXyz.kt
@@ -2,10 +2,9 @@ package com.lagradost.cloudstream3.extractors
 
 import android.util.Log
 import com.fasterxml.jackson.annotation.JsonProperty
-import com.fasterxml.jackson.module.kotlin.readValue
 import com.lagradost.cloudstream3.app
-import com.lagradost.cloudstream3.mapper
 import com.lagradost.cloudstream3.utils.*
+import com.lagradost.cloudstream3.utils.AppUtils.tryParseJson
 
 class PlayLtXyz: ExtractorApi() {
     override val name: String = "PlayLt"
@@ -13,7 +12,7 @@ class PlayLtXyz: ExtractorApi() {
     override val requiresReferer = true
 
     private data class ResponseData(
-        @JsonProperty("data") val data: String?
+        @JsonProperty("data") val data: String? = null
     )
 
     override suspend fun getUrl(url: String, referer: String?): List<ExtractorLink> {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/extractors/PlayLtXyz.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    override suspend fun getUrl(url: String, referer: String?): List<ExtractorLink> {
+        val extractedLinksList = mutableListOf<ExtractorLink>()
+        //Log.i(this.name, "Result => (url) $url")
+        var idUser = ""
+        var idFile = ""
+        var bodyText = ""
+        val doc = app.get(url, referer = referer).document
+        //Log.i(this.name, "Result => (url, script) $url / ${doc.select("script")}")
+        doc.select("script").forEach {
```
Blatzar, line 26: how about using something like 
`val bodyText =  doc.select("script").firstOrNull { ... } ?: ""`

app/src/main/java/com/lagradost/cloudstream3/extractors/PlayLtXyz.kt
```diff
@@ -0,0 +1,84 @@
...
+                ?.groupValues?.get(0) ?: ""
+
+            idFile = "(?<=var idfile = \")(.*)(?=\";)".toRegex().find(bodyText)
+                ?.groupValues?.get(0) ?: ""
+        }
+        //Log.i(this.name, "Result => (idUser, idFile) $idUser / $idFile")
+        if (idUser.isNotBlank() && idFile.isNotBlank()) {
+            //val sess = HttpSession()
+            val ajaxHead = mapOf(
+                Pair("Origin", mainUrl),
```
Blatzar, line 47: you can use 
"Origin" to mainUrl instead of Pair() just FYI :+1:

Jacekun: I know but Pair() is more readable to me. (Curse of C# and Python)

app/src/main/java/com/lagradost/cloudstream3/extractors/PlayLtXyz.kt
Jacekun: The first ``script`` does not always contain the required info so iterating over them is kinda needed. Though for optimization, I could skip the other elements once ``bodyText`` is not empty. Or, I could ``.filter{ }`` the ``doc.select``

Blatzar: firstOrNull iterates over the list and stops the iteration once a result is found, i think this would be the equivalent code.

```kotlin
val bodyText = doc.select("script").firstOrNull {
    it?.toString()?.contains("var idUser") == true
}?.toString()
```

Jacekun: Ah yeah that could work. Will change it

Blatzar (approved):
Looks good

PR #88, opened 2022-09-05
add voting system

C10udburst:


Blatzar (approved):
I don't think I would change a single line, it looks great!

LagradOst:
Everything looks fine, however I would rename Votes: 1 to rated or likes. Also sorting by votes, however that is really not needed

LagradOst:
Also it defaults to 10K if you dont have any internet

C10udburst:
> Everything looks fine, however I would rename Votes: 1 to rated or likes. Also sorting by votes, however that is really not needed

what about karma :trollface:

Blatzar:
> what about karma :trollface:

:skull:

Blatzar:
Bananas would unironically work :flushed:

C10udburst:
> Bananas would unironically work 😳

all right :shipit:

PR #95, opened 2022-09-07
croatian language added - real

Anarchydr:


Blatzar:
TODO, this is missing the changes in app to make the language selectable and readme, I can fix that tomorrow if you don't wanna dig around to fix that :+1:

app/src/main/res/values-hr/strings.xml
```diff
@@ -0,0 +1,628 @@
+<!--https://newbedev.com/concatenate-multiple-strings-in-xml-->
+<resources>
+    <!-- KEYS DON'T TRANSLATE -->
+    <string name="search_providers_list_key" translatable="false">search_providers_list</string>
```
C10udburst, line 4: please remove all of the entries which have `translatable="false"` as they shouldn't be translated

C10udburst (changes requested):

PR #97, opened 2022-09-08
removed "Provider" from extension name

hexated:


app/src/main/java/com/lagradost/cloudstream3/ui/settings/extensions/PluginAdapter.kt
```diff
@@ -135,6 +135,7 @@ class PluginAdapter(
         ) {
             val metadata = data.plugin.second
             val disabled = metadata.status == PROVIDER_STATUS_DOWN
+            val name = if (metadata.name.contains("Provider")) metadata.name.replace("Provider", "") else metadata.name
```
Blatzar, line 138: The if statement isn't required and I think replaceSuffix is more appropriate, but good thinking :+1:

hexated: Do i need to change it?

Blatzar (approved):
Looks good to me 👍

PR #128, opened 2022-09-30
GitHub account

antonydp:


Blatzar:
Looks good, @LagradOst ?

Is there anything missing or incomplete? I can help if there's anything :pray:

antonydp:
The main functions (backupGithub & restorePromptGithub) are working for me, 
we now need to decide if it's best to add a dedicated "backup data on cloud" button or to make it do that automatically or both.
Also before merging I think someone else should compile it and try all the features to see if it's actually working fine.

venkat2305:
Can you explain how it works in layman terms?

antonydp:
Basically the current backup feature already coded in cloudstream works thanks to two main functions:
Backup () writes and stores in the phone memory a txt file with the user data
RestorePrompt() use the txt file the user selects and convert back the txt into cloudstream data

This cloud solutions mimics this procedure
BackupGithub() creates a gist in your gist.github account and uploads there the txt file that backup().
In order to do that you need to provide a GitHub token for authentication
RestorePromptGithub() simply downloads the gist and uses the current RestorePrompt()

venkat2305:
After logging in , every time we open CS3 , it downloads and uploads the text(back up data) right?

antonydp:
I only coded backupGithub & restorePromptGithub, the logic regarding how to automate the process is not there yet.
I was in fact asking if it's best to add a dedicated "backup data on cloud" button or to make it do that automatically or both.

venkat2305:
I think it should be automatic. We use multiple devices(me personally tv, laptop, phone). We can just continue from another device without doing anything.
Or maybe options for automatic and and dedicated button???

Blatzar:
Okay for automatic restoration I'm thinking something like 

```kt
ioSafe {
    initAll()
    // No duplicates (which can happen by registerMainAPI)
    apis = allProviders.distinctBy { it }

    if (githubApi.getLatestLoginData() != null && settingsManager.getBoolean(getString(R.string.automatic_cloud_backups), false)){
        this.restorePromptGithub()
    }
}
```
at https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt#L584
and something like
```kt
override fun onDestroy() {
    val broadcastIntent = Intent()
    broadcastIntent.action = "restart_service"
    broadcastIntent.setClass(this, VideoDownloadRestartReceiver::class.java)

    if (githubApi.getLatestLoginData() != null && settingsManager.getBoolean(getString(R.string.automatic_cloud_backups), false)) {
        this.backupGithub()
    }

    this.sendBroadcast(broadcastIntent)
    afterPluginsLoadedEvent -= ::onAllPluginsLoaded
    super.onDestroy()
}
```
At https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt#L400
I am not 100% this will cause a race condition since its threaded, but if you ever see that it does not backup the data to github try moving it to the onStop or lastly onPause

Blatzar:
This would assume you had a switch in settings with the key R.string.automatic_cloud_backups :+1:

antonydp:
There's something to fix in the restoreprompt: it looks like it's restoring cached data before it gets the backup from cloud.
does it make sense?

Blatzar:
I am not quite sure what you mean :confused:

antonydp:
I wasn't sure what I meant either 🤣 
anyway we made some changes, it still needs testing, if someone else can compile it and test it, that would be nice

venkat2305:
Can it backup and restore app settings?

C10udburst:
hm. it still doesnt seem to properly push the settings to the gist

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -607,6 +615,10 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
             initAll()
             // No duplicates (which can happen by registerMainAPI)
             apis = allProviders.distinctBy { it }
+
+            if (githubApi.getLatestLoginData() != null && settingsManager.getBoolean(getString(R.string.automatic_cloud_backups), false)){
```
Blatzar, line 619: You cannot have false as a default value here, but true in the xml
That will make it look like you have it on while it evaluates to false

Blatzar, line 619: This is also a race condition as initAll() inits the providers, not apis
Move this check to after `inAppAuths.amap`

Blatzar:
To make the home page refresh you can do the following:
In the HomeFragment companion object:

`val reloadStoredDataEvent = Event<Unit>()`

in the onstop and onresume in homefragment.kt inlude the new method like the following

```kt
    override fun onResume() {
        super.onResume()
        reloadStored()
        afterPluginsLoadedEvent += ::firstLoadHomePage
        mainPluginsLoadedEvent += ::firstLoadHomePage
        reloadStoredDataEvent += ::reloadStoredEvent
    }

    override fun onStop() {
        afterPluginsLoadedEvent -= ::firstLoadHomePage
        mainPluginsLoadedEvent -= ::firstLoadHomePage
        reloadStoredDataEvent -= ::reloadStoredEvent
        super.onStop()
    }

    private fun reloadStoredEvent(input: Unit) {
        reloadStored()
    }
```
Now you can call `HomeFragment.reloadStoredDataEvent.invoke(Unit)` from wherever you want and it'll reload all the local data @antonydp

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -498,6 +506,15 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
         }
     }
 
+    override fun onStart() {
+        val settingsManager = PreferenceManager.getDefaultSharedPreferences(this)
+        super.onStart()
+        ioSafe {
+            if (githubApi.getLatestLoginData() != null && settingsManager.getBoolean(getString(R.string.automatic_cloud_backups), true)){
```
Blatzar, line 513: This can still cause a race condition :pensive: 
I have thought a while and the best way I can think of resolving this is

1. Create a private variable `val accountsLoginLock = Mutex()`
2. [Here](https://github.com/recloudstream/cloudstream/blob/b6868217a03b0e0f8755fc933b65302fc9e60095/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt#L610) do `githubLoginLock.withLock { ` after the ioSafe
3. Do the same [here](https://github.com/recloudstream/cloudstream/blob/b6868217a03b0e0f8755fc933b65302fc9e60095/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt#L512)

The idea behind this is that when you do:

`githubLoginLock.withLock { /* first code here*/ }`

on one thread and then call another piece of code like this while the first runs:

`githubLoginLock.withLock { /* second code here*/ }`

then the second code will only be launched when the first code is completed.
You want to do this as otherwise the restore code can be launched before the app is fully logged in to github, which will fail the `githubApi.getLatestLoginData() != null` check and therefore not restore anything only because it was too fast.

C10udburst (approved):
works perfectly for me

LagradOst:
I tested this now on a fresh dummy account on a fresh cs3 install and got some things to say.
1. the toast when exiting the app saying "saved data" is super annoying, it is better to have a toast if it *fails* 
2. It did not work for me and instead created many gist with the same name?????
image
3. The account got fucked because it looked sus af to have an 20min old acc creating many gists. please add a not that that might happend if ppl create new accounts 
image

I can take a look at the code too, but these are kinda important impo

rokejulianlockhart:
What did this PR do?

rokejulianlockhart:
> https://github.com/recloudstream/cloudstream/pull/128#issuecomment-1675069301

https://codeberg.org/cloudstream/cloudstream/issues/426#issuecomment-849020

PR #145, opened 2022-10-09
update strings.xml Greek

thanasistrisp:
Add new translations to greek language (el) and fix typos.


LagradOst:
Error: Found item String/synopsis more than one time

PR #155, opened 2022-10-15
Change link to web archive

DoYouNotice:
normal link is not working anymore


C10udburst:
Yes, yes it is.
Because of a complaint from SkyUK & Kopjra we're unable to provide links to repositories as some extensions can provide access to some uninteresting TV show which is not in public domain.

PR #158, opened 2022-10-16
Switch to kotlin build scripts

JFronny:
AGP and Gradle support writing build scripts in kotlin, which allows for better IDE support.
Since this project is written in kotlin, doing so just makes sense.
Additionally, I have implemented auto-incrementing version codes for the prerelease flavor.


C10udburst:
[build](https://github.com/recloudstream/cloudstream/actions/runs/3259203831/jobs/5351859350#step:6:9)
No files were found with the provided path: app/build/outputs/apk/debug/*.apk. No artifacts will be uploaded.

.github/workflows/pull_request.yml
```diff
@@ -15,9 +15,9 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Run Gradle
-      run: ./gradlew assembleDebug
```
C10udburst, line 18: hm why is the pull request build a `prerelease` and not `debug`?

C10udburst (changes requested):

.github/workflows/pull_request.yml
JFronny: It is a debug build using the prerelease flavor and is therefore debuggable, automatically version and uses the prerelase id, which seemed like a sensible setup, but a debug build with the release flavor (which would better match the current setup) is also an option, if that is preferred.

C10udburst: why cant it be a debug flavor as it was prior to this pr?

JFronny: release and debug are build types, not flavors, and doing things like modifying version codes is impossible from build types, so two flavors (release: a normal cloudstream build and prerelease: autogenerated versions and the prerelease suffix) were added as a second dimension of configurations.
Presented as a table:

<table>
<tr>
	<td>
	<td>prerelease
	<td>release
<tr>
	<td>release
	<td>No debugging, autogenerated version, application suffix
	<td>No debugging, normal version, no suffix
<tr>
	<td>debug
	<td>Debugging, autogenerated version, application suffix
	<td>Debugging, normal version, no suffix
</table>

JFronny: Technically, just giving up on either automatic version codes for prereleases or manual version codes for stable releases would remove the need for them, but its up to you to decide what option of these three you prefer

C10udburst: all right 🙏. lgtm as is

C10udburst (approved):

Blatzar:
@LagradOst :whip:

LagradOst (approved):

PR #163, opened 2022-10-18
added sub to streamsb & xtreamCdn

hexated:


app/src/main/java/com/lagradost/cloudstream3/extractors/XStreamCdn.kt
```diff
@@ -59,44 +70,66 @@ open class XStreamCdn : ExtractorApi() {
...
                     }
                 }
             }
+
+            val userData = sources?.player?.poster_file?.split("/")?.get(2)
+            sources?.captions?.map {
+                subtitleCallback.invoke(
+                    SubtitleFile(
+                        it.language,
+                        "$mainUrl/asset/userdata/$userData/caption/${it.hash}/${it.id}.srt"
```
Blatzar, line 129: Is .srt really hardcoded? I thought they had an "extension" field in the json response

Blatzar:
I'll merge this, but please use .forEach{} instead of .map{} when the result of said map is not used.

PR #167, opened 2022-10-23
strings.xml

ardoslav:
Translation in Bulgarian language.


LagradOst:
Very good, however you also need to add it here :+1: https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L57

ardoslav:
Hello I did what you asked of me me.and there is a mistake at  g i t  h u b
. I am illiterate in coding so please help me resolve the issue

C10udburst:
added manually in 044822040f7da8c777958057a12d1f7ee3a4222c

PR #172, opened 2022-10-24
[Feature] Automatically download plugin, based on language setting

Jacekun:
~~Todo: Waiting for setting re- implementation~~

Ready for merge.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
@@ -220,6 +220,9 @@ object PluginManager {
      * 4. Else load the plugin normally
      **/
     fun updateAllOnlinePluginsAndLoadThem(activity: Activity) {
+        // Load all plugins as fast as possible!
+        loadAllOnlinePlugins(activity)
+        afterPluginsLoadedEvent.invoke(true)
 
         val urls = (getKey<Array<RepositoryData>>(REPOSITORIES_KEY)
             ?: emptyArray()) + PREBUILT_REPOSITORIES
--- a/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
@@ -317,7 +317,7 @@ object PluginManager {
             //Omit lang not selected on language setting
             val lang = sitePlugin.language ?: return@outer null
             //If set to 'universal', don't skip any language
-            if (!providerLang.contains("universal") && !providerLang.contains(lang)) {
+            if (!providerLang.contains(AllLanguagesName) && !providerLang.contains(lang)) {
                 return@outer null
             }
             //Log.i(TAG, "sitePlugin lang => $lang")
--- a/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
@@ -30,7 +30,6 @@ import com.lagradost.cloudstream3.plugins.RepositoryManager.downloadPluginToFile
 import com.lagradost.cloudstream3.plugins.RepositoryManager.getRepoPlugins
 import com.lagradost.cloudstream3.ui.settings.extensions.REPOSITORIES_KEY
 import com.lagradost.cloudstream3.ui.settings.extensions.RepositoryData
-import com.lagradost.cloudstream3.utils.Coroutines.ioSafe
 import com.lagradost.cloudstream3.utils.Coroutines.main
 import com.lagradost.cloudstream3.utils.ExtractorApi
 import com.lagradost.cloudstream3.utils.UIHelper.colorFromAttribute
--- a/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
@@ -267,8 +267,8 @@ object PluginManager {
         }
 
         main {
-            val uitext = UiText.StringResource(R.string.plugins_updated, updatedPlugins)
-            createNotification(activity, uitext)
+            val uitext = txt(R.string.plugins_updated, updatedPlugins.size)
+            createNotification(activity, uitext, updatedPlugins)
         }
 
         // ioSafe {
@@ -348,8 +348,8 @@ object PluginManager {
         }
 
         main {
-            val uitext = UiText.StringResource(R.string.plugins_downloaded, newDownloadPlugins)
-            createNotification(activity, uitext)
+            val uitext = txt(R.string.plugins_downloaded, newDownloadPlugins.size)
+            createNotification(activity, uitext, newDownloadPlugins)
         }
 
         // ioSafe {
@@ -611,15 +611,14 @@ object PluginManager {
 
     private fun createNotification(
         context: Context,
-        uitext: UiText.StringResource
+        uitext: UiText,
+        extensions: List<String>
     ): Notification? {
         try {
-            val extensionNames = uitext.args.filterIsInstance<String>()
-            val resIdFormat = uitext.resId
 
-            if (extensionNames.isEmpty()) return null
+            if (extensions.isEmpty()) return null
 
-            val content = extensionNames.joinToString(", ")
+            val content = extensions.joinToString(", ")
 //        main { // DON'T WANT TO SLOW IT DOWN
             val builder = NotificationCompat.Builder(context, EXTENSIONS_CHANNEL_ID)
                 .setAutoCancel(false)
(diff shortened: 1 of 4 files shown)


C10udburst:
ngl i don't like that you're forcing users to download all the plugins of a given language w/o any option to prevent it

Jacekun:
> ngl i don't like that you're forcing users to download all the plugins of a given language w/o any option to prevent it

Its only if 'Auto update plugin' is on

C10udburst:
> > ngl i don't like that you're forcing users to download all the plugins of a given language w/o any option to prevent it
> 
> Its only if 'Auto update plugin' is on

it should be a separate setting

Jacekun:
That's doable. Though I'd argue that updating plugins is the same as downloading missing ones. And less processing time for the user

C10udburst:
> That's doable. Though I'd argue that updating plugins is the same as downloading missing ones. And less processing time for the user

then make it on by default but leave it as an option. its not the same (eg it would force me to download all english movie providers when i use one or two)

LagradOst:
🤨

Jacekun:
> 🤨

Whenever a new extension is added to my repo, I wanna have my cloudstream to download it. More plugins = more options = less overload.

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -527,7 +615,8 @@ object PluginManager {
 
     private fun createNotification(
         context: Context,
-        extensionNames: List<String>
+        extensionNames: List<String>,
+        title: Int
```
Blatzar, line 619: @StringRes title: Int clarifies it's a string resource

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -265,16 +267,102 @@ object PluginManager {
...
+    /**
+     * Automatically download plugins not yet existing on local
+     * 1. Gets all online data from online plugins repo
+     * 2. Fetch all not downloaded plugins
+     * 3. Download them and reload plugins
+     **/
+    fun downloadNotExistingPluginsAndLoad(activity: Activity, isLoaded: Boolean) {
+        // Load all plugins as fast as possible, if its not loaded yet.
+        if (!isLoaded) {
+            loadAllOnlinePlugins(activity)
```
Blatzar, line 289: Not a big fan of making the method doing two things at once like this :pensive: 
Why not move the code in this if statement to the MainActivity code? That would be more legible imo

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -265,16 +267,102 @@ object PluginManager {
...
+            //If set to 'universal', don't skip any language
+            if (!providerLang.contains(AllLanguagesName) && !providerLang.contains(lang)) {
+                return@outer null
+            }
+            //Log.i(TAG, "sitePlugin lang => $lang")
+
+            //Omit NSFW, if disabled
+            sitePlugin.tvTypes?.let { tvtypes ->
+                if (!settingsForProvider.enableAdult) {
+                    if (tvtypes.contains("NSFW")) {
```
Blatzar, line 327: I can see why you did this as we do have another reference in the codebase doing this, but I'd rather have TvType.NSFW.name for some kind of future proofing :pensive:

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Blatzar: I am a guilty of doing it myself which is probably why you did, but I've had real issues debugging the race conditions afterwards :skull:

Blatzar (commented):
Overall good PR!

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Jacekun: The method runs when auto plugin is enabled (cause it is run inside it). This is just a checker to not run it when its already executed on the auto plugin update method. If not, it runs it like how it is before

Jacekun: I could move the code, and have it run before checking for plugin updates

Blatzar: Why not something like this?
```kt
                    if (settingsManager.getBoolean(
                            getString(R.string.auto_update_plugins_key),
                            true
                        )
                    ) {
                        PluginManager.updateAllOnlinePluginsAndLoadThem(this@MainActivity)
                    } else {
                        PluginManager.loadAllOnlinePlugins(this@MainActivity)
                    }
                    // Now all online plugins are loaded
                    if (....) PluginManager.downloadNotExistingPluginsAndLoad(this@MainActivity)
```
All plugins are loaded before the downloadNotExistingPluginsAndLoad
This would be much easier to read

Jacekun: hmm that could work too. But honestly, the 2 methods* gets executed regardless, so not much difference.

* ``loadAllOnlinePlugins()`` is always executed first. ``afterPluginsLoadedEvent.invoke(true)`` runs on plugin update and download.

Blatzar:
Looks good, I will test it later today :+1:

Blatzar:
Looks good :pray:

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -527,7 +608,8 @@ object PluginManager {
 
     private fun createNotification(
         context: Context,
-        extensionNames: List<String>
+        extensionNames: List<String>,
+        @StringRes title: Int
```
LagradOst, line 612: if you are going to do this but also support formatting use UiText instead, https://github.com/recloudstream/cloudstream/blob/28b4456dfd6e808553a7bbffbd06c8ef4e6dfc9a/app/src/main/java/com/lagradost/cloudstream3/ui/result/UiText.kt#L43

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -265,16 +266,96 @@ object PluginManager {
...
+            ?: emptyArray()) + PREBUILT_REPOSITORIES
+        val onlinePlugins = urls.toList().apmap {
+            getRepoPlugins(it.url)?.toList() ?: emptyList()
+        }.flatten().distinctBy { it.second.url }
+
+        val providerLang = activity.getApiProviderLangSettings()
+        //Log.i(TAG, "providerLang => ${providerLang.toJson()}")
+
+        // Iterate online repos and returns not downloaded plugins
+        val notDownloadedPlugins = onlinePlugins.mapNotNull outer@{ onlineData ->
```
LagradOst, line 297: not sure why you have to label this

LagradOst (commented):
tbh I dont know much about the dex system, but check out UiText

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Jacekun: You meant the ``outer`` loop? My prev implementation required 2 ``.mapNotNull`` but now I've changed it to only use 1. It can be removed.

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Jacekun: The string resource needs a decimal input (list size), and the UI text flattens the list entries so it doesn't match

ahmedtn2021:
Salut
es qu'il ya des serveurs en français 
Merci

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Jacekun: Yeah Ik, I meant inside UIText.toString(), the method flattens the list and not doing ``extensionNames.size``. It will always error cause wrong parameter is used. Will have to check more about UiText().

C10udburst (approved):

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -265,16 +267,98 @@ object PluginManager {
...
+            //If set to 'universal', don't skip any language
+            if (!providerLang.contains(AllLanguagesName) && !providerLang.contains(lang)) {
+                return@mapNotNull null
+            }
+            //Log.i(TAG, "sitePlugin lang => $lang")
+
+            //Omit NSFW, if disabled
+            sitePlugin.tvTypes?.let { tvtypes ->
+                if (!settingsForProvider.enableAdult) {
+                    if (tvtypes.contains(TvType.NSFW.name)) {
```
Blatzar, line 322: pretty much everything is great besides unnecessary nested if statements here 
`if (!settingsForProvider.enableAdult && tvtypes.contains(TvType.NSFW.name)) {`
does the same :pray:

C10udburst:
> @LagradOst you have one day to object before i merge +1

image

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Jacekun: Micro optimization. If the toggle is off, no need to check if tvtype contains the NSFW type.

Blatzar: No, the compiler already does that

Simpson65:
Hey, I'm new to this. Can you please help me with cloud storage repositories?

PR #187, opened 2022-10-31
Initialize imdb functionality for more accurate searching of movies' subtitles

thanasistrisp:
By getting an API from https://www.themoviedb.org/ with unlimited uses and by using the https://github.com/holgerbrandl/themoviedbapi/ implementation for Java/Kotlin, now searching subs is far more accurate (for now only movies is implemented for subs, series are untouched), as the search inside app returned most ways inaccurate results. Now the search is accomplished with imdb id and year. This was already implemented, however not the imdb id as it was a TODO.
For example when I was searching Dual before, the results were bad, however when searching it now we this implementation the correct subs returned!

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -113,6 +113,11 @@ android {
         checkReleaseBuilds = false
     }
     namespace = "com.lagradost.cloudstream3"
+    packagingOptions {
+        resources {
+            excludes += "META-INF/DEPENDENCIES"
+        }
+    }
 }
 
 repositories {
@@ -217,6 +222,9 @@ dependencies {
 
     // Library/extensions searching with Levenshtein distance
     implementation("me.xdrop:fuzzywuzzy:1.4.0")
+    implementation("com.github.holgerbrandl:themoviedbapi:1.12") {
+        exclude(group = "commons-logging", module = "commons-logging")
+    }
 }
 
 tasks.register("androidSourcesJar", Jar::class) {
--- a/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/OpenSubtitlesApi.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/OpenSubtitlesApi.kt
@@ -6,6 +6,7 @@ import com.lagradost.cloudstream3.*
 import com.lagradost.cloudstream3.AcraApplication.Companion.getKey
 import com.lagradost.cloudstream3.AcraApplication.Companion.removeKey
 import com.lagradost.cloudstream3.AcraApplication.Companion.setKey
+import com.lagradost.cloudstream3.metaproviders.TmdbProvider
 import com.lagradost.cloudstream3.mvvm.logError
 import com.lagradost.cloudstream3.subtitles.AbstractSubApi
 import com.lagradost.cloudstream3.subtitles.AbstractSubtitleEntities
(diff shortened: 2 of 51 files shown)


Jacekun:
Use query.year. Check if there's a UI on subtitle search for the year

Blatzar:
Why cannot the tmdb api already present in app be used?

app/build.gradle.kts
```diff
@@ -217,6 +222,9 @@ dependencies {
 
     // Library/extensions searching with Levenshtein distance
     implementation("me.xdrop:fuzzywuzzy:1.4.0")
+    implementation("com.github.holgerbrandl:themoviedbapi:1.12") {
```
Blatzar, line 225: The issue lies with using another API, not that the code is in the wrong file

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/OpenSubtitlesApi.kt
```diff
@@ -163,17 +164,21 @@ class OpenSubtitlesApi(index: Int) : InAppAuthAPIManager(index), AbstractSubApi
...
 
-        val imdbId = query.imdb ?: 0
         val queryText = query.query.replace(" ", "+")
         val epNum = query.epNumber ?: 0
         val seasonNum = query.seasonNumber ?: 0
         val yearNum = query.year ?: 0
+        val tmdbprov = TmdbProvider()
+        var imdbId = ""
+        if (seasonNum == 0 && epNum == 0) { // applicable only for movies for now
+            imdbId = tmdbprov.getImdb(queryText, yearNum)
```
Blatzar, line 174: This should be moved up the chain as not only opensubtitles could benefit from IMDb id
See https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt#L551

This would of course require adding a get IMDb id method to the viewmodel which would allow for easy caching via an hashmap.

app/build.gradle.kts
Blatzar: You never want unnecessary dependencies

Blatzar (changes requested):
Please change to the current tmdb api dependency

Blatzar:
> Well, I used the defined inapp api. However, my question remains: how can I get the year of the movie? By using `query.year` I get null. This need to be fixed. 

See https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt#L340 
The year is simply never set, this is simply because we have not bothered to pass over the data from the info page yet

app/build.gradle.kts
Blatzar: See https://github.com/recloudstream/cloudstream/blob/master/app/build.gradle.kts#L197
We already have a dependency to handle the JSON that the API returns.

Blatzar: But that method uses com.github.holgerbrandl:themoviedbapi, the unnecessary dependency

Blatzar: Because the other API already has a sizable codebase in the app which works, not to mention https://github.com/UweTrottmann/tmdb-java has more recent commits and releases

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/OpenSubtitlesApi.kt
Blatzar: You do not want to fetch imdbId twice if you click the load subtitles twice, hence why you want cache and to do that you move the getImdbId method to https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerGeneratorViewModel.kt 

Viewmodels handle all the business logic and allows you to keep a local internal state, e.g a hashmap of previous search results. See https://developer.android.com/topic/architecture/ui-layer/stateholders

Jacekun:
> The year is simply never set, this is simply because we have not bothered to pass over the data from the info page yet

And now it will be possible with PR #232

Luna712:
@thanasistrisp @fire-light42 Is it okay if I try to do this PR again using current APIs or do you want me to leave this and it'll be completed again?

thanasistrisp:
> @thanasistrisp @fire-light42 Is it okay if I try to do this PR again using current APIs or do you want me to leave this and it'll be completed again?

You can try it of course.

PR #202, opened 2022-11-07
Add zh_TW

jhihyulin:
Add Traditional Chinese (Taiwan)


C10udburst:
Fails to build: `values-zh-TW: Error: Invalid resource directory name`

Sdarfeesh:
It seems like the Traditional Chinese translation cannot load, could you fix that?
https://user-images.githubusercontent.com/50188628/201454476-c077b64e-697f-4275-a87f-ac2e0007bc8d.mp4

jhihyulin:
I don't know how to fix this bug, but now I can go into the application and skip the language setting and it will display normally, so the problem should be in the language code in `app\src\main\java\com\lagradost\cloudstream3\ui\settings\SettingsGeneral.kt` line 79, or in the other settings of the language setting file?

Sdarfeesh:
> I don't know how to fix this bug, but now I can go into the application and skip the language setting and it will display normally, so the problem should be in the language code in `app\src\main\java\com\lagradost\cloudstream3\ui\settings\SettingsGeneral.kt` line 79, or in the other settings of the language setting file?

`https://github.com/recloudstream/cloudstream/blob/3fc63a216548a93baf74209dd814c682e17f9a7c/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79`

The language code should be same as value-XXXX !!!

jhihyulin:
> > I don't know how to fix this bug, but now I can go into the application and skip the language setting and it will display normally, so the problem should be in the language code in `app\src\main\java\com\lagradost\cloudstream3\ui\settings\SettingsGeneral.kt` line 79, or in the other settings of the language setting file?
> 
> https://github.com/recloudstream/cloudstream/blob/3fc63a216548a93baf74209dd814c682e17f9a7c/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> 
> The language code should be same as values-* !!!

https://github.com/recloudstream/cloudstream/blob/0a8cb1d10c8bd35394cd0f208671046e1399cdeb/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79

https://github.com/recloudstream/cloudstream/blob/809693eca5dbccc7099bda57f281de28ebad8dd4/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79

Tried it, doesn't work

Sdarfeesh:
> > > I don't know how to fix this bug, but now I can go into the application and skip the language setting and it will display normally, so the problem should be in the language code in `app\src\main\java\com\lagradost\cloudstream3\ui\settings\SettingsGeneral.kt` line 79, or in the other settings of the language setting file?
> > 
> > 
> > https://github.com/recloudstream/cloudstream/blob/3fc63a216548a93baf74209dd814c682e17f9a7c/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> > 
> > The language code should be same as values-* !!!
> 
> https://github.com/recloudstream/cloudstream/blob/0a8cb1d10c8bd35394cd0f208671046e1399cdeb/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> 
> https://github.com/recloudstream/cloudstream/blob/809693eca5dbccc7099bda57f281de28ebad8dd4/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> 
> Tried it, doesn't work

I found the problem, because ISO 639-1 codes only have zh code for Chinese, so it's not possible to add another Chinese language

Sdarfeesh:
> > > I don't know how to fix this bug, but now I can go into the application and skip the language setting and it will display normally, so the problem should be in the language code in `app\src\main\java\com\lagradost\cloudstream3\ui\settings\SettingsGeneral.kt` line 79, or in the other settings of the language setting file?
> > 
> > 
> > https://github.com/recloudstream/cloudstream/blob/3fc63a216548a93baf74209dd814c682e17f9a7c/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> > 
> > The language code should be same as values-* !!!
> 
> https://github.com/recloudstream/cloudstream/blob/0a8cb1d10c8bd35394cd0f208671046e1399cdeb/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> 
> https://github.com/recloudstream/cloudstream/blob/809693eca5dbccc7099bda57f281de28ebad8dd4/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt#L79
> 
> Tried it, doesn't work

You can try to use a code of ISO 639-1 codes that isn't used
(For example, tw)

jhihyulin:
I don't think using other code is the best option, which may cause the language to not be set automatically.

BTW, `tw` is already used in the ISO 639-1
Screenshot_2022-11-12-17-23-09-48_40deb401b9ffe8e1df2f1cc5ba480b12

Blatzar:
Java Locale has Traditional Chinese but that cannot be instantiated given only a language code.

I think creating an exception hashmap and just manually mapping the locales there would be the least messy solution:

Right below _appLanguages_:
```kt
/**
 * Not all languages can be fetched from locale with a code.
 * This map allows sidestepping the default Locale(languageCode)
 * when setting the app language.
 **/
val appLanguageExceptions = hashMapOf(
    "zh-Hant-TW" to Locale.TRADITIONAL_CHINESE
)
```

and in https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt#L113 you can add the exception like this:

```kt
val locale = appLanguageExceptions[languageCode] ?: Locale(languageCode)
```

jhihyulin:
The language can now be switched normally, but there is still a problem with the part where the language selection is displayed.

Blatzar:
> The language can now be switched normally, but there is still a problem with the part where the language selection is displayed.

What do you mean?

jhihyulin:
> > The language can now be switched normally, but there is still a problem with the part where the language selection is displayed.
> 
> What do you mean?

Like this
Select Traditional Chinese, the language of the interface is also Traditional Chinese, but the language menu shows that Simplified Chinese is selected.

https://user-images.githubusercontent.com/74854953/201515081-5cc582d7-3d7e-4a10-86aa-005e336c0c6d.mp4

Blatzar:
Weird, I'll get to investigating soon

Blatzar:
Sorry for the wait :pensive:

PR #221, opened 2022-11-24
feat(subtitle): add addic7ed.com

daarkdemon:


Blatzar:
nice!

Blatzar:
image
Great PR but the names are way too weird and there is no hearing impairment support :pensive: 
image

Even the main series name is possible to get, although in an url, but it would be way more readable than `S1E3 NF.WEBRip-STARZ+AMRAP+ION10`

Testing on [the witcher s1e3](https://www.addic7ed.com/ajax_loadShow.php?show=7659&season=1&langs=&hd=undefined&hi=undefined) :+1:

bachig26:
@Blatzar 
does this PR works as indented?
will it be added soon?

PR #223, opened 2022-11-27
[Feature] Use subtitle filename, instead of movie title for Opensubtitles search.

Jacekun:
ready for merge


app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/OpenSubtitlesApi.kt
```diff
@@ -200,8 +200,12 @@ class OpenSubtitlesApi(index: Int) : InAppAuthAPIManager(index), AbstractSubApi
             it.data?.forEach { item ->
                 val attr = item.attributes ?: return@forEach
                 val featureDetails = attr.featDetails
+                //Use filename as name, if its valid
+                val filename = attr.files?.mapNotNull { subfile ->
```
Blatzar, line 204: You can just use firstOrNull {} instead of mapping everything

Jacekun: If the first item have null ``file_name``, it would return null immediately. I want it to return the first item with non-null ``file_name`` value.

CyberLykan:
Thank you for implementing this!

PR #232, opened 2022-12-04
[Feature] Allow input of Year on Subtitle search

Jacekun:
Added UI for year Textbox to be used on search queries.

Added so PR #187 can be continued.

TODO: Automatically set Year on Textbox if the current loaded movie has year metadata. Will probably do this on another PR.

For now, ready for merge.


Blatzar (approved):
ill fix the rest :+1:

PR #242, opened 2022-12-12
Add new playback speed option of 2.5

hairowits:
Add another option to set playback speed to 2.5
Hopefully one more option wouldn't hurt.
Thanks


LagradOst:
😕 I am not too keen on this

hairowits:
> 😕 I am not too keen on this

So sorry for the inconvenience. I will close the pr then. Can you at least tell me why you don't like this?

LagradOst:
> Can you at least tell me why you don't like this?
Watching media at anything faster than 2x is just wasteful, dont watch it if it is that much of a bother to watch it, read a summery if you have to catch up with your friends. You cant experience a movie or tv-series any good at such high speed. Adding more speed options is just bloat and ruins the watching experience. If smth then I would like to limit it to 1.5x max or smth, but toooo many ppl watch at 2x

PR #274, opened 2022-12-24
Separate loading view from homepage

tuyafeng:
Thanks for your project, it helped me a lot.

I have a device with poor performance, when cloudstream is running on it, some operations cause stuttering, so I want to optimize its performance.

I first found that the home page has a lot of unnecessary loading status views, and we can remove them completely when the loading is over. I added a framelayout as the fragment's container, and loaded the loading state as a fragment into the container.

In this way we can separate the loading part from the home page, which will make the code more maintainable.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
@@ -4,9 +4,7 @@ import android.annotation.SuppressLint
 import android.app.Activity
 import android.content.Context
 import android.content.DialogInterface
-import android.content.Intent
 import android.content.res.Configuration
-import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.view.LayoutInflater
@@ -33,7 +31,6 @@ import com.google.android.material.button.MaterialButton
 import com.google.android.material.chip.Chip
 import com.google.android.material.chip.ChipGroup
 import com.lagradost.cloudstream3.*
-import com.lagradost.cloudstream3.APIHolder.apis
 import com.lagradost.cloudstream3.APIHolder.filterProviderByPreferredMedia
 import com.lagradost.cloudstream3.APIHolder.getApiFromNameNull
 import com.lagradost.cloudstream3.APIHolder.getApiProviderLangSettings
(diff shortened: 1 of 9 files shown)


LagradOst:
Intresting

LagradOst:
Does it actually improve performance for you?

tuyafeng:
I'm still working on other parts. This is just a small part. I did some simple measurements and below are my data.

Before the change, it takes about 350ms to load the layout (inflate view) on the homepage (on my [xiaomi mi6 sagit](https://wiki.lineageos.org/devices/sagit/)), after separation, it takes 300ms to inflate the view, which is still time-consuming but helpful.

tuyafeng:
In the next step, I plan to use constraint layout to reduce layout nesting, and use View+Drawable to replace the placeholder CardView, which should also help.

LagradOst:
Ok, sounds nice. I was planning on doing a tv homepage UI rework, but I will hold off on that while you do this

Blatzar:
Upon investigation it looks like the main culprit of slow load times is the main recyclerview not recycling as it's in a nested scroll view. This will get fixed soon.

tuyafeng:
You can set the reuse pool of these recyclerviews to be the same. Going a step further, you can even use recyclerview nested recyclerview to implement the main page (of course, you also need to set their reuse pool to the same, which will make some Views enter the recycling stack).

Referer: [RecyclerView#setRecycledViewPool](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView#setRecycledViewPool(androidx.recyclerview.widget.RecyclerView.RecycledViewPool))

tuyafeng:
But let's still pick up some cigarette butts and do small optimizations like splitting the home page state, which I think are good for maintaining the code and are not as heavy as the above mentioned optimizations and can be done easily. This PR is complete, you can merge it directly, of course, if you have a better way, you can also modify it after closing it. This is your project, feel free to do it.

Blatzar:
Don't get me wrong, this pr will get merged and I'm very thankful for it, I'll do a proper check tomorrow 🙏
I'm just informing you to get opinions and prevent double work

And I mean the vertical recyclerview, it's currently inside a nested scroll view to allow the top stuff to be displayed. If you got any ideas how to implement that without dragging the header logic inside the recyclerview itself it'd be a godsend 🙏

Rendering 18 rows at once is extremely heavy on the UI thread 💀

tuyafeng:
I think the current approach is acceptable if you don't want nesting to optimize performance. ViewPager2 itself is a RecyclerView, sharing the reuse pool will help, and [RecyclerView#setHasFixedSize](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView#setHasFixedSize(boolean)) will also help.

Blatzar:
Ill try to fix the conflicts soon

LagradOst:
closing as stale and outdated with the new binding system

PR #337, opened 2023-01-26
[extractor] added ByteShare

LiJu09:
Added ByteShare extractor.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/extractors/ByteShare.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/extractors/ByteShare.kt
@@ -15,15 +15,17 @@ open class ByteShare : ExtractorApi() {
                 if (script.data().contains("'use strict';")) {
                     val data = script.data()
                         .substringAfter("sources: [").substringBefore("]")
-                        .replace(" ","")
+                        .replace(" ", "")
                         .substringAfter("src:\"").substringBefore("\",")
-                    sources.add(ExtractorLink(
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/extractors/ByteShare.kt
```diff
@@ -0,0 +1,35 @@
...
+    override val mainUrl = "https://byteshare.net"
+    override val requiresReferer = false
+
+    override suspend fun getUrl(url: String, referer: String?): List<ExtractorLink> {
+        val sources = mutableListOf<ExtractorLink>()
+        with(app.get(url).document) {
+            this.select("script").map { script ->
+                if (script.data().contains("'use strict';")) {
+                    val data = script.data()
+                        .substringAfter("sources: [").substringBefore("]")
```
LagradOst, line 17: you sure you cant parse as json or smth?

LiJu09: found even better way

LagradOst:
damn, LGTM

LagradOst:
assuming the download link is a raw link

LiJu09:
yeah, its same as in `src: ...`

LagradOst:
> yeah, its same as in `src: ...`

but can you play that link in vlc for example

LiJu09:
> but can you play that link in vlc for example

yes

PR #342, opened 2023-01-28
Use user setting for seeking on Android TV

TerHano:


LagradOst:
@doteq

bachig26:
@TerHano I don't understand much (on coding), but looking at the changes. it looks same.
It looks like both left and right key executes the forward seek.

TerHano:
> @TerHano I don't understand much (on coding), but looking at the changes. it looks same.
> It looks like both left and right key executes the forward seek.

Sorry about that, I had done some edits prior to checking in and it looks like the minus was deleted. Should be fixed now.

doteq:
When I was designing TV UI I wanted to have two ways of seeking.
 - If you couldn't hear something you can just rewind 10 seconds by clicking left button without interface opened
 - If you want to find some part in video you can seek by 30 seconds when interface is opened
 I feel that my design is very intuitive because it gives you more control when seeking. You could add second "TV seeking time" setting, so users can tweak that and use it here, but using only one value is a bad idea.

TerHano:
I agree that having two settings would be nice. I made the change and added the settings, let me know if it's what you had in mind.

doteq:
I would remove semicolons, because you don't need them in kotlin and it looks weird when only your lines have them, but overall looks good to me, great work :)

LagradOst:
LGTM, ready to merge?

PR #365, opened 2023-02-10
Sendvid extractor

Stormunblessed:


app/src/main/java/com/lagradost/cloudstream3/extractors/Sendvid.kt
```diff
@@ -0,0 +1,32 @@
...
+        val urlString = doc.select("head meta[property=og:video:secure_url]").attr("content")
+        val sources = mutableListOf<ExtractorLink>()
+        if (urlString.contains("m3u8"))  {
+                generateM3u8(
+                    name,
+                    urlString,
+                    mainUrl,
+                    headers = app.get(url).headers.toMap()
+                ).forEach {link ->
+                    sources.add(link)
```
Blatzar, line 27: Adding sources is what we've done a lot in our extractors, but that is kinda deprecated. It would be much cleaner to use the new https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/ExtractorApi.kt#L456-L463
and then just do `.forEach(callback)`

app/src/main/java/com/lagradost/cloudstream3/extractors/Sendvid.kt
```diff
@@ -0,0 +1,32 @@
...
+    override suspend fun getUrl(url: String, referer: String?): List<ExtractorLink>? {
+        val doc = app.get(url).document
+        val urlString = doc.select("head meta[property=og:video:secure_url]").attr("content")
+        val sources = mutableListOf<ExtractorLink>()
+        if (urlString.contains("m3u8"))  {
+                generateM3u8(
+                    name,
+                    urlString,
+                    mainUrl,
+                    headers = app.get(url).headers.toMap()
```
Blatzar, line 25: This looks like an exceedingly unnecessary request :sob: 
Please do something like `val response = app.get(url)` and then get the headers and doc separately from that response instead of making the request twice.

Blatzar: We dont have that good caching in cs3 :skull:

app/src/main/java/com/lagradost/cloudstream3/extractors/Sendvid.kt
```diff
@@ -0,0 +1,32 @@
+package com.lagradost.cloudstream3.extractors
+
+import com.lagradost.cloudstream3.utils.*
+import com.lagradost.cloudstream3.app
+import com.lagradost.cloudstream3.utils.M3u8Helper.Companion.generateM3u8
+
+class SendvidHttps: Sendvid() {
+    override val mainUrl: String = "https://www.sendvid.com"
```
Blatzar, line 8: This is not required because of:
https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/ExtractorApi.kt#L94-L98
But good thinking :pray:

Blatzar (changes requested):
Good extractor, some hard to catch issues tho :pensive: 
We lack good dev docs :skull:

app/src/main/java/com/lagradost/cloudstream3/extractors/Sendvid.kt
Blatzar: ^ not a required change

PR #377, opened 2023-02-16
update VideoDownloadService

MhmdIbrahim1:
In this pull request, I've replaced the IntentService with a Service that uses Kotlin coroutines to perform the download operation asynchronously. Here are the main changes:

Replaced the IntentService with a regular Service that uses coroutines for asynchronous execution.
Added a coroutine scope and context to manage the coroutines.
Launched a coroutine to perform the download operation asynchronously, ensuring that it runs on a separate thread and does not block the main thread.
Added a onDestroy method to cancel the coroutine context when the service is destroyed, to avoid any potential memory leaks.
Using coroutines in this way can potentially make the code more efficient and easier to read, while also ensuring that the user experience is not impacted by long-running operations


Blatzar:
Merging this as IntentService is deprecated

PR #384, opened 2023-02-19
Replace notifyDataSetChanged() with notifyItemRangeChanged() in setList(), use requireActivity().onBackPressedDispatcher.onBackPressed()

MhmdIbrahim1:
## Description

This pull request makes the following changes to the code:

- In the `setList()` function, we replaced the `notifyDataSetChanged()` method call with `notifyItemRangeChanged()` to use a more specific change event.

 I replaced the `activity?.onBackPressed()` call with `requireActivity().onBackPressedDispatcher.onBackPressed()` to use the new `onBackPressedDispatcher` API. Also updated the `notifyDataSetChanged()` method call with `notifyItemRangeChanged()` for more efficient updates.


LagradOst:
nononono never ever do requestactivity, that shit will crash harder than a plane with no wings, you notify shit is also the wrong fix, check out diffutill that is used legit for every other list

MhmdIbrahim1:
Ooooh! I didn't think it was that bad :D

PR #385, opened 2023-02-19
Fix StreamSB stream break: support audiotracks

LikDev-256:
The previous method,
`M3u8Helper.generateM3u8()`
caused the stream break, which caused the extractor to not be able to support multi-audio tracks.

My method involves not using `M3u8Helper.generateM3u8()` it uses `ExtractorLink()` (Psychopath Method)
Which allows the player to automatically detect audio tracks.

The code change,
```
safeApiCall {
            callback.invoke(
                ExtractorLink(
                    name,
                    name,
                    mapped?.streamData?.file ?: return@safeApiCall,
                    url,
                    Qualities.Unknown.value,
                    true,
                    headers
                )
            )
        }
```
Any my crappy code fixes are welcomed 🙏🏼

**A proof of concept can be found [Here](https://github.com/LikDev-256/likdev256-tamil-providers/blob/master/ShowFlixProvider/src/main/kotlin/com/likdev256/ShowFlixProvider.kt#L447)**


LagradOst:
I dont exactly get why you have safeApiCall, remove that

LikDev-256:
This PR is useless @hexated 's SoraStream heavily relies on M3u8Helper() and It works better.
I'm closing this.

About ,
> I dont exactly get why you have safeApiCall, remove that

I actually don't remember why I  used it

PR #402, opened 2023-02-28
Intertal Player: Added MPD support

pokerface-bad:


Blatzar:
Don't you need the exoplayer dash dependency too?

Blatzar:
I will merge after I fix manifestversion in extensions to not break every single extension when this merge goes live :pensive:

Blatzar:
Cba fixing manifestversion so I added a constructor I made sure would work with the old extensions :smiley:

PR #443, opened 2023-04-04
Right-To-Left language UI fix

MudssirAhmed:


LagradOst:
This does not fix it, it only makes it worse

PR #448, opened 2023-04-09
CS3IPlayer: fix buffer lost when seeked to backward

Shif-Jess:
https://user-images.githubusercontent.com/117321707/230777753-a5919d32-50c7-4e61-9134-8f3ab8304b68.mp4

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -670,11 +670,6 @@ class CS3IPlayer : IPlayer {
              **/
             maxVideoHeight: Int? = null
         ): ExoPlayer {
-            val bufferDurationMs = if (videoBufferMs <= 0) {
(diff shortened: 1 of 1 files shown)


LagradOst:
uhhh, uniform back and forwards buffer, that is like a lot of waste. I would hardcode it at 30ish sec or half of regular and you often watch movies/tvseries with minimal timeskipping backwards. there is already a lot of Out Of Memory errors, and this will just cause a lot more if we just double the buffer size like this. good pr tho

asl97:
> uhhh, uniform back and forwards buffer, that is like a lot of waste. I would hardcode it at 30ish sec or half of regular and you often watch movies/tvseries with minimal timeskipping backwards. there is already a lot of Out Of Memory errors, and this will just cause a lot more if we just double the buffer size like this. good pr tho

PIP seek by 30 sec, maybe PIP seek should be reduce or a bigger buffer than 30 sec.

https://github.com/recloudstream/cloudstream/blob/94e7eb8e9d3e4603ab6d1956a8003b36a9d25095/app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerPipHelper.kt#L52-L60

PR #495, opened 2023-07-01
Updated some of the extractor Domains

Nexus-Gits:


LagradOst:
urls should not end with /

PR #505, opened 2023-07-11
getTracker: add matching with year & type as fallback

Sofie99:
[tested] I noticed that some anime still missing when using getTracker, so i added matching with year & type as fallback to increase accuracy. idk about non english site but as long as that website provide romanji/english title it should work fine and anilist search is awesome..


LagradOst:
That can return VERY bad results and I much rather have anilist have less matches than false positives. if you could match more strictly then it will be merged, but as it stands no

Issue #386, opened 2023-02-19
Black bar in PIP when using internal player


asl97:
### Steps to reproduce

1. Check that PIP (Picture In Picture) is enable
2. Check internal player is used
3. Find content that isn't the same aspect ratio as the screen, in most cases, it be 4:3 (640x480)
4. Play the content in "Fit to Screen" and then press the home button to enter PIP mode
5. See black bar still showing in PIP mode

### Expected behavior

No black bar, such as when using VLC or Chrome (web browser)

### Actual behavior

Black bar is still shown despite in PIP mode

### Cloudstream version and commit hash

4.0.0 33aecfb

### Android version

Android 12

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [X] If related to a provider, I have checked the site and it works, but not the app.
- [X] I will fill out all of the requested information in this form.

LagradOst:
exo issue

LagradOst:
The problem is that the vid probs has black bars in its encoding and aspec ratio change is not preserved in pip

asl97:
> The problem is that the vid probs has black bars in its encoding
That doesn't really explain why using the long press > play in VLC/Web browser's PIP doesn't have black bar, I did test it without VLC custom PIP

To ensure both player are playing the same file, I tired using the download feature and it shows black bar in both internal and VLC, it did not play in chrome, just result in a .ts download, but there was no black bar in Just Player and Mi Video, I am guessing that's a separate issue due to combining the source streams from the m3u8 together into an mp4

>and aspec ratio change is not presearved in pip
I assume you mean the aspect ratio wasn't updated for the PIP to reflect the source aspect ratio, which I agree is most likely the issue

asl97:
tested with the same video I had issue with, the black bar issue is fixed.

I assume #525 fixed it.

PR #508, opened 2023-07-14
Migrated to Media3

self-similarity:
Exoplayer is deprecated as per its latest [release](https://github.com/google/ExoPlayer/releases). To keep up with exoplayer releases a Media3 migration is required.

Sound, subtitles, video, PiP and headphone control all works as it should with the new changes.
There are some small changes relating to PiP to make it utilize the latest Android features such as smooth resizing.

The setting enableJetifier was disabled as no library was using it.

Closes https://github.com/recloudstream/cloudstream/issues/386


app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerPipHelper.kt
```diff
@@ -26,7 +26,7 @@ class PlayerPipHelper {
                     activity,
                     code,
                     Intent("media_control").putExtra("control_type", code),
-                    0
+                    PendingIntent.FLAG_IMMUTABLE
```
LagradOst, line 29: todo remove

LagradOst (changes requested):
todo fix black bars

LagradOst:
moved to #525

PR #518, opened 2023-07-26
Feature: Refactor autodownload plugin to have multiple modes.

Jacekun:
[Feature] Refactor auto-download plugin to choose 1 of the following modes:

- Disable
- Filter by Preferred Language,
- All
- NSFW Only

### DIFF
--- a/app/src/main/res/values/array.xml
+++ b/app/src/main/res/values/array.xml
@@ -41,7 +41,14 @@
         <item>0</item>
         <item>1</item>
     </array>
-
+    
+    <!-- MainAPI enum: AutoDownloadMode -->
+    <array name="auto_download_plugin">
+        <item>Disable</item>
+        <item>Filter by Preferred Language</item>
+        <item>All</item>
+        <item>NSFW Only</item>
+    </array>
 
     <array name="player_pref_names">
         <item>@string/player_settings_play_in_app</item>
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/array.xml
```diff
@@ -41,7 +41,14 @@
...
     </array>
-
+    
+    <!-- MainAPI enum: AutoDownloadMode -->
+    <array name="auto_download_plugin">
+        <item>Disable</item>
+        <item>Filter by Preferred Language</item>
+        <item>All</item>
+        <item>NSFW Only</item>
+    </array>
```
LagradOst, line 51: This should be res/strings as only strings are translated, not array.

you can use 
@string/nsfw
@string/all 
@subtitles_filter_lang
and add
@string/disable = disable

LagradOst (commented):
I am a bit unsure how useful this feature is tbh, feels just a bit bloaty

Jacekun:
> I am a bit unsure how useful this feature is tbh, feels just a bit bloaty

I agree its a pretty niche feature. but it'd be nice to have for *isolated* apps

app/src/main/res/values/array.xml
Jacekun: Done changes

PR #523, opened 2023-07-31
 Extractor: fix Gofile and added Userscloud 

Sofie99:


LagradOst:
This site has cloudflare, so you might need to add some cloudflare killer. otherwise it looks fine even if I have not tested it

PR #528, opened 2023-08-02
Use ffmpeg library

self-similarity:
Closes https://github.com/recloudstream/cloudstream/issues/464

PR #533, opened 2023-08-02
fix: can not switch subtitle after integrate ffmpeg decoder.

### LINKED ISSUE #531: Subtitles not working/ Downloading
It appears that after the recent pre-release update, subtitles are not working/ downloading. Can you please check. Also can you provide the way to revert to previous apk.

d4rkk3y:
fix #531 without revert #528

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -7,16 +7,14 @@ import android.os.Looper
 import android.util.Log
 import android.util.Rational
 import android.widget.FrameLayout
-import androidx.media3.common.C
-import androidx.preference.PreferenceManager
 import androidx.media3.common.C.*
 import androidx.media3.common.Format
 import androidx.media3.common.MediaItem
-import androidx.media3.common.TrackSelectionOverride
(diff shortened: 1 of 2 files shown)


self-similarity:
:partying_face:

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -662,12 +663,7 @@ class CS3IPlayer : IPlayer {
 
         private fun getTrackSelector(context: Context, maxVideoHeight: Int?): TrackSelector {
             val trackSelector = DefaultTrackSelector(context)
-            trackSelector.parameters = DefaultTrackSelector.ParametersBuilder(context)
-                // .setRendererDisabled(C.TRACK_TYPE_VIDEO, true)
-                .setRendererDisabled(C.TRACK_TYPE_TEXT, true)
```
LagradOst, line 667: This seems to be the culprit and your patch works without the setTrackTypeDisabled
Are there any reasons to have those too?

d4rkk3y: Without those two setTrackTypeDisabled when I choose `No Subtitles` in subtitle selection, player still show subtitle (maybe default subtitle).

PR #534, opened 2023-08-02
fix player session id

self-similarity:
Fixes the session id error which prevented two concurrent players, like trailers and normal videos or local file and PiP.
Maybe UUID could be used instead.


LagradOst:
LGTM 👍

PR #537, opened 2023-08-04
More meaningful errors when adding repositories

self-similarity:
When adding repositories the app now checks the repository validity and gives error messages to help the user.

The download all prompt does not appear if there are no plugins to download.


LagradOst (approved):
LGTM 👍

PR #544, opened 2023-08-09
fix voting api

C10udburst:
Few caveats:
- There's no ability to edit votes
- They may have some weird sus analytics (see [this](https://counterapi.com/stats/cs-codeberg.org-cloudstream-cloudstream-extensions/vote))


LagradOst:
why change it?

C10udburst:
> why change it?

the old one doesn't work

PR #538, opened 2023-08-05
TV UX improvements

self-similarity:
- Added auto focus to the skip timestamp popup on TV

- Added a checkmark to the Dub/Sub/Season/Episode selector

PR #548, opened 2023-08-11
Added Simkl

self-similarity:
Unfortunately also includes stuff from https://github.com/recloudstream/cloudstream/pull/538


app/src/main/java/com/lagradost/cloudstream3/syncproviders/SyncAPI.kt
```diff
@@ -59,14 +60,24 @@ interface SyncAPI : OAuth2API {
         override var id: Int? = null,
     ) : SearchResponse
 
+    abstract class AbstractSyncStatus {
```
self-similarity, line 63: Required change for passing around data not present in the base class

KillerDogeEmpire:
@self-similarity if it includes stuff from #538 close that PR

app/build.gradle.kts
```diff
@@ -54,17 +55,27 @@ android {
...
         buildConfigField(
             "String",
             "BUILDDATE",
             "new java.text.SimpleDateFormat(\"yyyy-MM-dd HH:mm\").format(new java.util.Date(" + System.currentTimeMillis() + "L));"
         )
-
+        buildConfigField(
+            "String",
+            "SIMKL_CLIENT_ID",
+            "\"" + (System.getenv("SIMKL_CLIENT_ID") ?: localProperties["simkl.id"]) + "\""
```
C10udburst, line 72: could you add these env vars to `.github/workflows/build_to_archive.yml` and `.github/workflows/prerelease.yml`?

self-similarity: Added but untested :+1:

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SimklApi.kt
```diff
@@ -0,0 +1,843 @@
...
+
+        val code = uri.getQueryParameter("code") ?: return false
+        val token = app.post(
+            "$mainUrl/oauth/token", json = TokenRequest(code)
+        ).parsedSafe<TokenResponse>() ?: return false
+
+        switchToNewAccount()
+        setKey(accountId, SIMKL_TOKEN_KEY, token.access_token)
+
+        val user = getUser()
```
LagradOst, line 830: if getUser throws an error switchToOldAccount will not get called, will lead to invalid state, add normalsafeapicall

LagradOst (approved):
just add one line and then LGTM 👍

PR #555, opened 2023-08-15
Fix episode removal in simkl

self-similarity:


LagradOst (approved):
LGTM, ready to merge?

PR #565, opened 2023-08-18
Add Native Crash Handler

self-similarity:
Arbitrary dex files can lead to arbitrary errors, even low level SIGSEGV. Instead of killing the app this throws an exception which can be handled by the normal crash handlers, properly starting safe mode.

This catches SIGSEGV and sets a variable which can be polled from Kotlin, notifying a crash.
It is [not possible to do much from a signal handler](https://stackoverflow.com/a/12434262), least of all run Kotlin code which is why polling is required. It may be possible to do something better but this seems reliable enough with minimal overhead.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/NativeCrashHandler.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/NativeCrashHandler.kt
@@ -1,6 +1,7 @@
 package com.lagradost.cloudstream3
 
 import com.lagradost.cloudstream3.MainActivity.Companion.lastError
+import com.lagradost.cloudstream3.mvvm.logError
 import com.lagradost.cloudstream3.plugins.PluginManager.checkSafeModeFile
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/lagradost/cloudstream3/NativeCrashHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+            // Do not crash in safe mode!
+            if (lastError != null) continue
+            if (checkSafeModeFile()) continue
+
+            throw RuntimeException("Native crash with code: $signal. Try uninstalling extensions.\n")
+        }
+    }
+
+    fun initCrashHandler() {
+        System.loadLibrary("native-lib")
```
LagradOst, line 31: try catch this shit

LagradOst (approved):
Very nice 👍

PR #567, opened 2023-08-19
chore: better subtitle decoding

IndusAryan:


LagradOst:
This pr just looks useless. 2.4.0 is the latest version and I dont want libs to be bleeding edge in case of bugs. Only newpipe requires this commit level version

PR #571, opened 2023-08-23
Lower targetSdk to get all installed packages

CranberrySoup:
To get Aniyomi Extension Support you must be able to query all installed packages. This sadly requires lowering the targetSdk (like Aniyomi does). The code should not be impacted by this change.


edl2:
Whar merg

PR #578, opened 2023-08-27
fix: speedostream extractor

IndusAryan:
fixes yomovies


LagradOst:
🤨duplicate extractors?

LagradOst:
If both redirect to the same, remove one

LagradOst:
> actually did that first,but they didn't remove the server choice from yowebsite, i think its just temporary offline and and their current speedo server has only 360p and 480p support. another one had 720p. If they remove that one then i will remove too. i will wait for a week,this can be merged as of now

I will just add it quickly myself
image

PR #582, opened 2023-08-28
dont pause playback unfocused if no video present

C10udburst:


LagradOst:
smells good to me, ready to merge 👍

Luna712:
I wonder if we could have a "Play as audio" button for background video play as well, similar to VLC?

PR #590, opened 2023-09-04
Upgrade SDK

CranberrySoup:
The previous reason for downgrading the SDK was that the app could not properly query all installed apps even with `QUERY_ALL_PACKAGES`. It turns out that https://github.com/recloudstream/cloudstream/commit/c92ac3e8b3502f26ed812647134dc0977218831e#diff-7fa6aef292187a049f7a4d6060d8df3ba212d838789c78940bd363344b1c38cdR18 actually fixed `QUERY_ALL_PACKAGES` not working. This makes it possible to upgrade the SDK again :smiley: 

Note: targetSdk not updated to 34 because I have not tested it against https://developer.android.com/about/versions/14/behavior-changes-14

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -50,13 +50,13 @@ android {
         }
     }
 
-    compileSdk = 34
+    compileSdk = 33
     buildToolsVersion = "34.0.0"
 
     defaultConfig {
         applicationId = "com.lagradost.cloudstream3"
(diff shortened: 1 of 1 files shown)


CranberrySoup:
I think this will fix the updater not working too

LagradOst:
looks good to me 👍

Issue #591, opened 2023-09-05
Automatic Backup


ghost:
### Describe your suggested feature

Not like what everyone I saw is talking about, what about scheduled backups to a specific folder at a specific time of the day, while deleting the older one (like signal does), this will be very useful and easy to do and we can just sync the file to the cloud using tools like Rclone.

Or just change the data folder of the app from Android/data to Android/media to be easily accessible and updated in real time.

I know there have been many tries to do this with GitHub or Google but I don't want to give this kind of data to any of them (and I'm not a Google user). The possibility of an offline backup gives me the control to do whatever I want.

Thank you.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #592, opened 2023-09-06
Automatic backups

self-similarity:
Based on SubscriptionWorkManager.
Refactored some backup code to handle context better.

Resolves https://github.com/recloudstream/cloudstream/issues/591


C10udburst:
> I'm asking because the build has failed so there is no apk to test.

build failed because we are being held hostage by JitPack 😭

Blatzar:
There is no fixed folder, all backups are tied to whatever your video downloading path is (which is changeable). While it is not that clean it leaves room for user options. Ideally it would be in a backup folder in the download path.

PR #593, opened 2023-09-07
getTracker: switched to anilist api

Sofie99:
seems like consumet is dying.. :p


LagradOst (approved):
Code looks good, however I have not tested it yet

LagradOst:
Tried it now, and it works fine. Just wondering before I merge, what was wrong with the consumet api as that still works for me

bachig26:
> Tried it now, and it works fine. Just wondering before I merge, what was wrong with the consumet api as that still works for me

from what i know, Consumet API (Base URL: https://api.consumet.org/) getting lot of DDOS attacks frequently and recently the dev's are moving to other projects due to this; the development and maintenance are slowed down and there's a mention that they might completely shut down anytime soon.

PR #598, opened 2023-09-10
Fix opensubtitles

self-similarity:
New specifications require all API calls to contain a user-agent with app name and app version. 
https://opensubtitles.stoplight.io/docs/opensubtitles-api/51ea4c484e6e5-changelog


LagradOst (approved):
any reason to make it and headerInterceptor instead of just appending the headers?

PR #606, opened 2023-09-15
Many UI fixes

CranberrySoup:
1. Resized episode/season/dub width to the largest button to make it more standardized
2. Made the shadow always dynamically cover the episode selector. Always readable text regardless of width.
3. Fixed rare bug where an episode group would end in -2147483648
4. Removed download tab on TV
5. Reordered sub/dub to be on the left side of episodes
6. Made account homepage persistent
image
image
image


brezelman:
> Removed download tab on TV
Could you please revert that change? I always preload two to three episodes, so I don't have to care about streaming conditions when I watch them.

There was no problem with that on fire TV light. And it's possible to extend the available space by attaching an external drive via usb otg.

PR #608, opened 2023-09-15
Potential fix for crash loops

CranberrySoup:
Untested but should theoretically fix unrecoverable plugin exceptions.
The .cs3 files are compiled against an old version of the app with pointers to methods which may have changed, leading to a low level NPE.
Recompiling the dex files hopefully fixes outdated pointers.


omjhon8422:
Tv android v 6.0 still crash for me

PR #618, opened 2023-09-20
Add homepage search on TV

CranberrySoup:
image
image


KillerDogeEmpire:
@CranberrySoup can you add mal and anilist/simkl sync to tv layout? When you have time? Would love to watch anime on TV and it sync

LagradOst (approved):

PR #623, opened 2023-09-22
Remote sync draft

CranberrySoup:
Continuation of https://github.com/recloudstream/cloudstream/pull/442

- [X] Viewbindings
- [ ] ???

*more to come here when I get going*


Jerico3241:
So hyped for this feature.

bachig26:
@CranberrySoup only the Google Drive supported?
is there any possibility for other cloud services to be included?

Luna712:
Some of the documentation for this is wrong if it is the first time using google cloud console. You need to create OAuth consent screen in order to create the credentials and some of the names of links the documentation says appears to either be inaccurate or just outdated.

Luna712:
Screenshot_20230922_171952_CloudStream Debug
It should also probably be able to pull name by default (and allow to be charged to a custom name) and profile icon (and allow the icons like local accounts to be used if desired) from authenticated Google account rather than named like this and an empty icon. Additionally it should probably be the OAuth setup separately from account login, that way multiple google accounts can be synced using the same Auth application so it only has to be done once. Finally, the client secret field should probably be a password-type field to hide its value as that should not usually be public for security purposes. That is just my thoughts and opinions on it, maybe not all necessary or good ideas.

CranberrySoup:
For anyone wondering whats happening: this is not dead, there is just some massive changes going on :+1:

CranberrySoup:
This is now functional, but requires testing and better documentation.

app/src/main/res/layout/add_account_input_oauth.xml
```diff
@@ -0,0 +1,153 @@
...
+            android:layout_width="wrap_content"
+            android:layout_gravity="center_vertical|end"
+            android:text="@string/login" />
+
+        <com.google.android.material.button.MaterialButton
+            android:id="@+id/cancel_btt"
+            style="@style/BlackButton"
+            android:layout_width="wrap_content"
+            android:layout_gravity="center_vertical|end"
+            android:text="@string/sort_cancel" />
```
Luna712, line 151:
```suggestion
            android:text="@string/cancel" />
```
Very minor and just a suggestion, because `cancel` is the main key used for this, and makes more sense, `sort_cancel` just used `cancel` anyway.

Luna712:
This works pretty good now, the one thing I would suggest still is maybe some sort of prefill select to reuse keys from one Google Cloud Project when adding another account, chances are one Google Cloud Project can be used for all accounts using sync with.

Luna712:
Also I wonder if we could hook Google accounts into the local account system to use Google accounts to save data to them, and have name/picture etc in account select for example, and show profile picture instead of the default account icon also. Could work for some data account provider, but is not needed for this PR I don't think, just a thought for later.

app/src/debug/res/drawable/ic_baseline_add_to_drive_24.xml
Luna712, line 1: Shouldn't this be in main not debug?

CranberrySoup: You are correct, thank you for finding this :+1:

TNTwise:
After testing, I have found that if a file has been downloaded on a phone, and then is synced to a computer, resume from the home menu doesn't work. Idk if this is a known issue, but I think its to do with this block of code located in 

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchHelper.kt

`if (card.isFromDownload) {
                                handleDownloadClick(
                                    DownloadClickEvent(
                                        DOWNLOAD_ACTION_PLAY_FILE,
                                        VideoDownloadHelper.DownloadEpisodeCached(
                                            card.name,
                                            card.posterUrl,
                                            card.episode ?: 0,
                                            card.season,
                                            id,
                                            card.parentId ?: return,
                                            null,
                                            null,
                                            System.currentTimeMillis()
                                        )
                                    )
                                )
                            }
                        } else {
                            loadSearchResult(card, START_ACTION_LOAD_EP, id)`

Idk kotlin, or much about how cloudstream functions, but there might be a way to try catch this?

KillerDogeEmpire:
@CranberrySoup this dead?

Luna712:
Any updates on this?

Fruehwirth:
Hey. Just wondering how's it going with that PR? Any updates?

KingLucius:
I hope to see this one finished

Luna712:
@CranberrySoup I just realized my merge accidentally reverted 5270467 so that may need reapplied. Apologies.

Issue #627, opened 2023-09-22
Support a preview seekbar for showing previews when seeking


Luna712:
### Describe your suggested feature

It would be nice if the player supported a preview seekbar, for showing thumbnail previews while you are seeking. Something like https://github.com/rubensousa/PreviewSeekBar which has [examples for media3 implementation](https://github.com/rubensousa/PreviewSeekBar#how-to-use-with-media3) as well. It is also fairly maintained.

If not the library, there may also be a solution using something similar to this: https://stackoverflow.com/a/62006472

Thumbnail sprite could probably be created as a temporary file using something like ffmpeg for online videos and perhaps something like [MediaMetadataRetriever](https://developer.android.com/reference/kotlin/android/media/MediaMetadataRetriever) for local videos. It could be generated in the loading step of a video, or even after while the video is playing or on first attempted seek and only stored while the video is playing, then removed. You could also have it save under a single file name and overwrite when a new one is generated so you only ever have one thumbnail sprite file generated and to look at.

In my opinion this would make the built in cloudstream player so much better and alot more friendly usage, and consistently with popular streaming services that support something like this, YouTube, Netflix, Google TV, Etc...

This is just an idea and I understand it might be complex, but I do believe it would be a fairly worthwhile feature improvement to the media player. If this is not something that will even be consider then I guess I understand, but it does seem to me like something at least feasible.

My technical knowledge of this is limited at best so if my theory here is totally not feasible then I apologize as well.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Blatzar:
I am well aware of both that exact stackoverflow question and that library as I have investigate this before. Actually doing the UI is not that difficult, getting proper thumbnails is the harder part.

MediaMetadataRetriever actually looks promising, but downloaded videos is pretty low priority (less used + no buffering).
Ffmpeg is pretty much the most viable method to get thumbnails in the least shitty way, but that itself requires a lot of work to implement. It is simply not worth it time wise compared to other requests, even though it is a good request. 

Developing all boils down to usefulness/time and this is much time and not that high usefulness :pensive:

Luna712:
I understand it might be quite complex, I probably wouldn't know how to implement it into the existing app code, but I can try to learn the codebase and implement something like this if it is something you'd actually be interested in having merged in?

[...]

PR #637, opened 2023-09-30
[DRAFT] Add ThumbnailSpriteGenerator

Luna712:
This is probably not all the best way to do this and is just a rough proof of concept

Per https://github.com/recloudstream/cloudstream/issues/627#issuecomment-1741881369

NOTE: this will not support some types of videos such as M3u8 playlists, without FFmpeg/FFprobe support implementation.

Usage example:
```kotlin
val spriteGenerator = ThumbnailSpriteGenerator(videoPath, object : ThumbnailSpriteCallback {
	override fun onThumbnailSpriteGenerated(spriteBitmap: Bitmap) {
		// We can do stuff after thumbnail sprite is generated
		runOnUiThread {
			val imageView = findViewById<ImageView>(R.id.imageView)

			// Set the Bitmap to the ImageView
			imageView.setImageBitmap(spriteBitmap)
		}
	}

	override fun onThumbnailSpriteGenerationError(error: Exception) {
		// We can do stuff when thumbnail sprite generation fails, i.e. log to logcat
		Log.e("Error", error.toString())
	}
})

AsyncTask.execute {
	spriteGenerator.generateThumbnailSprite()
}
```

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/ThumbnailSpriteGenerator.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/ThumbnailSpriteGenerator.kt
@@ -17,13 +17,15 @@ internal class ThumbnailSpriteGenerator(
 	private val callback: ThumbnailSpriteCallback
 ) {
 
-	private val maxLines = 10
-	private val maxColumns = 10
+	private val maxLines: Int = 10
+	private val maxColumns: Int = 10
+
+	private val minFrameIntervalSeconds: Int = 10
 
 	// By setting this here we can use setDataSource once
 	// This will improve performance, in particular for online videos
 	// It means that we don't have to seek through the video multiple times, only once
-	private val retriever = MediaMetadataRetriever()
+	private val retriever: MediaMetadataRetriever = MediaMetadataRetriever()
 
 	internal fun generateThumbnailSprite() {
 		retriever.setDataSource(videoPath)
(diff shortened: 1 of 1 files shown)


KingLucius:
What do you think about using the seeking step in settings to get the thumb generating step?
I think it will be easier and effective. To avoid generating useless thumbs.

Luna712:
> What do you think about using the seeking step in settings to get the thumb generating step? I think it will be easier and effective. To avoid generating useless thumbs.

Do you mean the "Player seek amount (Seconds)" setting? If so, the reason it determines here by lines and columns is if you use to many you may encounter some issues, or it can take a long time to generate as well.

The one thing I do need to do though is make max be max not min and max, so short videos don't make potentially multiple frames per second, by setting a minimum interval so it doesn't generate to many. We could get rid of max lines and columns and use a defined interval such as from settings, but then we could have an issue where we then have //to many// frames being generated as well. I tried to prioritize performance a lot when creating this, and found this to be the most effective way so it doesn't take a long time to load the frames.

If you meant something else, do you mind elaborating? Thanks!

KingLucius:
I got your point, thanks

LagradOst:
if this was implemented then smth like a recursive div 2 might be implemented to quickly generate good enough frames really quickly. (Eg first generate at 50%, then 25+75, then 12.5+37.5 +62.5+87.5 ect)
so if you try to seek while it is still generating you will still get a close match

KingLucius:
@LagradOst maybe 2 tasks, quick fetch for seeking feedback to get just 1 thumb for current seeking position and a slower one for total duration generation.
the quick process should check for thumb availability first if not there (from the slow process) then generate.

PR #653, opened 2023-10-05
Added new primary colors for comfort

KillerDogeEmpire:
This PR can be ignored if you want, just added these colors since a few of them were requested (specifically Dandelion Yellow), but also some of these are just comfortable to look at when you use cloudstream at night. I will add more to this branch  for new features.


Rayforest:
Yes! Please do add them. New colour choices are always welcome 🌱

PR #656, opened 2023-10-05
Reload library when reloading home

### LINKED ISSUE #650: Library doesn't reload when switching accounts
### Steps to reproduce

* Open library (with account1)
* Switch account
* Open Library again (with account2)

### Expected behavior

The library should now be from account2

### Actual behavior

The library is from account1 still, only restarting the app correctly gives library from account2 again.

4.2.0 3f51195

### Android version

Android 13

### Logcat

_No response_

### Other details

_No response_

### Acknowledgements

Luna712:
So that library is reloaded when switching accounts.

Fixes #650

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
@@ -131,7 +131,6 @@ object DataStoreHelper {
 
                             // update UI
                             setAccount(getDefaultAccount(context), true)
-                            MainActivity.bookmarksUpdatedEvent(true)
                             dialog?.dismissSafe()
                         }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -131,7 +131,6 @@ object DataStoreHelper {
 
                             // update UI
                             setAccount(getDefaultAccount(context), true)
-                            MainActivity.bookmarksUpdatedEvent(true)
```
Luna712, line 134: NOTE: This is removed because it is also done in setAccount so it was doing it twice. If that was incorrect, I will revert this part.

Luna712:
I forgot to mention this as well, the reason I just used the reloadHomeEvent is to prevent having another one for library when it would only ever be used at the same time. I can update to create a different one if you want though.

PR #663, opened 2023-10-07
Library on TV

KingLucius:
that's a draft as I need some help from you guys with the issues below:

Implemented:
- Library appears on Nav menu.
- Sort button in the AppBar.
- Next Focus implementation

Issues:
- page_recyclerview is focusable, I have tried everything I know without success.
- search_result_root is focusable empty tabs.


LagradOst:
add a testcase for the layout in the testing file

KingLucius:
> add a testcase for the layout in the testing file

Done

KingLucius:
Now the Tablayout is on top, also viewpager focus issues 90% fixed

KingLucius:
Everything related to Library on TV layout is ready to be merged, still only the "Sync" part in result TV layout I can't decide how it can fit in this layout.
I think we should merge this till deciding what is the best way to fit the sync in TV layout

At least the users would have the library on TV, still setting the progress via TV

PR #674, opened 2023-10-08
Fix issue where DownloadedPlayerActivity interferes with MainActivity

### LINKED ISSUE #605: Opening an external video in "Open with..." using CloudStream breaks things
### Steps to reproduce

1) Open an external video using CloudStream Beta
2) Click the back button to close the video
3) Open the CloudStream Beta app
4) Search for a movie on the search panel or click one from the home screen

Note: to reproduce CloudStream Beta must already be opened when you open the external video

### Expected behavior

It should be able to handle this case, and still work like normal when the external video is closed.

### Actual behavior

Nothing happens; click event is not functioning, you can not play anything at all. You have to restart CloudStream to get it working properly.

4.1.8 2bed79b

### Android version

Android 13

### Logcat

_No response_

### Other details

_No response_

### Acknowledgements

Luna712:
Fixes #605

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -61,7 +61,9 @@
             android:exported="true"
             android:resizeableActivity="true"
             android:screenOrientation="userLandscape"
-            android:supportsPictureInPicture="true">
+            android:supportsPictureInPicture="true"
+            android:taskAffinity="com.lagradost.cloudstream3.downloadedplayer"
+            android:launchMode="singleTask">
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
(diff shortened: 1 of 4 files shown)


Luna712:
This might not be the best way to do this, but it works. If there is a better way, then I apologize this can be closed and that can be tried.

LagradOst (approved):
bro, I honestly don't see how this fixes the problem, but yet it does??????

Luna712:
> bro, I honestly don't see how this fixes the problem, but yet it does??????

@LagradOst 

I updated this for a more efficient solution (seemingly), I realized there was an issue where if you opened download activity then main activity, it tried opening the MainActivity in a PiP window then crashes, this separates the tasks more efficiently now (from testing anyway).

Luna712:
Thanks for merging, this does make the DownloadedPlayerActivity a lot more proper implementation now, allowing it to be split into a second window altogether so you can say play something from file manager, pause it, go to main CloudStream, and return to the paused window later on without having to reply it. It's when reason I went with that approach and it solved one of the most annoying bugs in the app to me, with this now done, and preview seekbar implemented, I finally was able to make CloudStream the default media player on my phone. Thanks for all your hard work with the app as well!

PR #675, opened 2023-10-09
Delete Focus frame from empty Downloads list & Search TV Layout

KingLucius:


LagradOst:
🤨 if you want to actually remove it from the chips then you need to add it to the style

KingLucius:
Ok will update the PR

LagradOst:
I am actually on the fence about removing the white square entirely as too many ppl complain

KingLucius:
So there is no meaning for this PR, i think we can close it then.

KingLucius:
Removed the white square frame with tag via style & added rounded stroke as on TV it's hard to recognize the focused chip

PR #676, opened 2023-10-09
Upgrade androidx.navigation:navigation-fragment-ktx to 2.7.4

### LINKED ISSUE #625: Settings sometimes unresponsive
### Steps to reproduce

Sometimes if you go to downloads.and then settings, settings becomes unresponsive (nothing can be clicked) and tab still appears as if downloads is open. You have to restart the app to get it working again.

### Expected behavior

Settings should work as expected.

### Actual behavior

Settings becomes unresponsive.

4.2.0 d4fff7c

### Android version

Android 13

### Logcat

_No response_

### Other details

The screenshot is from CloudStream Debug but happens on CloudStream Beta and CloudStream apps.as well.
Screenshot_20230922_133836_CloudStream Debug~2

### Acknowledgements

Luna712:
This also required bumping targetSdk to 34

I am not sure if androidx.navigation:navigation-ui-ktx should be bumped as well or not.

Fixes #625

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -50,13 +50,13 @@ android {
         }
     }
 
-    compileSdk = 33
+    compileSdk = 34
     buildToolsVersion = "34.0.0"
 
     defaultConfig {
         applicationId = "com.lagradost.cloudstream3"
         minSdk = 21
-        targetSdk = 33
+        targetSdk = 34
 
         versionCode = 62
         versionName = "4.2.0"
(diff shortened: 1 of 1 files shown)


LagradOst:
are you sure this fixes the issue?

Luna712:
> are you sure this fixes the issue?

it appears to, I looked at commits, and confirmed one of the issues it mentions is at least similar, double clicking the back button is what breaks the navigation graph, consistently reproducable, doing that after upgrading this does not break it. Whether every cause is that I'm not sure 100% but this does fix that one issue at least.

Luna712:
Basically I can't confirm 100%, because I could never figure out the exact circumstances, I did figure out some exact circumstances it breaks now and this prevents that from happening. Whether there is another way it breaks, I'm unsure. @LagradOst

LagradOst:
im already fixing this

PR #677, opened 2023-10-10
feat(ui): material design upgrade

IndusAryan:
bottom sheets became real transparent so changed their attr to primary gray background that changes according to layout theme.
kept everything same. except bottom nav ripple


LagradOst:
this makes the UI ugly in lib

PR #679, opened 2023-10-10
feat(ui): upgrade material design but not library

IndusAryan:


LagradOst:
I am already fixing this

Issue #614, opened 2023-09-18
Support custom bookmark lists and watch history


Luna712:
### Describe your suggested feature

It would be nice to support adding custom bookmark lists. If that can not be done, it would alternatively be nice to have a list such as "Favorites". Additionally it would be nice if you could have proper watch history. Perhaps inspired by QuickNovel, which has a history/activity tab in navigation, rather than having to manually mark each one as "Completed", though this watch history wouldn't replace the "Completed" list either, it would just be nice to have.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Luna712:
@LagradOst I can do a PR to add a "Favorites" list if your okay with that. I found some past issues where you said you don't want to allow custom, which I do understand.

Luna712:
With favorites done I don't think this is needed anymore, at least not for me.

Luna712:
I closed this because #541 is watch history also and the bookmark list part of this is what isn't needed anymore for me anywah

PR #682, opened 2023-10-11
Add favorites

Luna712:
Refs #614

This also hides subscriptions on TV since they can not be used on TV so no reason to show them in library


Luna712:
This adds favorites to the top bar somewhat based on how subscriptions is done. This way you can add things to specific and favorite them separately, so you can keep better track of shows in the library. Since favorites has a different usage than bookmarks also, it isn't added on the homepage, only to the library. For now, it doesn't work on TV as I couldn't find a good layout implementation for it, so I hid both it and subscriptions in the library since neither can be used on TV, there is no reason for it to be there.

Screenshot_20231012_101418_CloudStream Debug
Screenshot_20231012_101433_CloudStream Debug

LagradOst (commented):
code looks fine, however I have not tested it, and no tv support?

Luna712:
> code looks fine, however I have not tested it, and no tv support?

@LagradOst 

No TV support *yet* I might do one on a follow-up, I just could not find a proper place for the UI element on TV layout just yet.

Luna712:
@LagradOst I was able to do TV support now afterall

PR #689, opened 2023-10-14
feat: make cloudstream fast boi, ksp migration

IndusAryan:
https://developer.android.com/build/migrate-to-ksp

recommended

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -5,6 +5,7 @@ import java.net.URL
 
 plugins {
     id("com.android.application")
+    id("com.google.devtools.ksp")
     id("kotlin-android")
     id("kotlin-kapt")
     id("org.jetbrains.dokka")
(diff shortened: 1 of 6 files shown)


LagradOst:
bro, you cant simply remove glideapp???? it has overrides we use

Blatzar:
GlideApp gets generated from [this](https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/GlideApp.kt#L23-L55), which features critical functionality like DOH and cache size :+1: 
Just changing it to the default glide removes this. Although it is very easy to miss this code since it is not directly referenced anywhere.

Blatzar:
Wow! That actually works
I am seldom surprised using kotlin, but damn
I checked that requests were sent correctly using a custom interceptor:

```kt
        val client =
            Requests().apply {
                defaultHeaders = mapOf("user-agent" to USER_AGENT)
            }.initClient(context)
                .newBuilder()
                .addInterceptor {
                    val request = it.request()
                    println("Request: ${request.url}")
                    it.proceed(request)
                }
                .addInterceptor(DdosGuardKiller(false))
                .build()
```
And it did work :pray:

Blatzar:
I am just not used to annotation magic, other languages can probably do this too :shrug: 
I just thought com.bumptech.glide.Glide would not use our GlideApp, it looks like any other innocent import.
Previously the class was generated and that was more intuitive, but this is better since they wont be undefined before building :+1: 

Good job :pray:

fire-light42:
@LagradOst Did you test this pr?

fire-light42:
It build for me at least

Issue #678, opened 2023-10-10
Library movies/series order


EHMilon:
### Steps to reproduce

In local library to short movies there are only tow option, first is Alphabetic (A to Z) or second Alphabetic (Z to A). But there is no option to short movies by updated time ascending descending order. It should be default setting. Which movie add last, it should be show first.

In SIMKL library, short movies by update time(new to old) and update (old to new) is not working. It's not changing the order of the movies.

### Expected behavior

If I save multiple movies on library. Last save movie should take the first position and last second will take the second place and go on. It should be the default library sitting. The chances are higher that last save movie is my most updated and priorities movie.

### Actual behavior

1. In local library short option limitations
2. SIMKL bugs in short by update time.

### Cloudstream version and commit hash

4.2.0 33eb3a3

### Android version

Android 12

### Logcat

```java
--------- beginning of main
10-10 20:24:29.947 18039  4886 I gralloc4: register: id=28000008609, vaddr=0x795370a000
10-10 20:24:29.948 18039  4886 I gralloc4: register: id=2800000860a, vaddr=0x794ae2e000
10-10 20:24:29.950 18039  4886 I gralloc4: register: id=2800000860b, vaddr=0x794a9b7000
10-10 20:24:29.951 18039  4886 I gralloc4: register: id=2800000860c, vaddr=0x7946560000
10-10 20:24:37.640 18039 28801 I gralloc4: unregister: id=2800000860a, vaddr=0x794ae2e000
10-10 20:24:37.640 18039 28801 I gralloc4: unregister: id=2800000860b, vaddr=0x794a9b7000
10-10 20:24:37.650 18039 18039 I gralloc4: unregister: id=2800000860c, vaddr=0x7946560000
10-10 20:24:37.651 18039 18039 I gralloc4: unregister: id=28000008609, vaddr=0x795370a000
10-10 20:24:37.894 18039 18039 W MediaRouter: Ignoring invalid provider descriptor: null
10-10 20:25:19.238 18039  5031 I gralloc4: register: id=2800000865d, vaddr=0x79b7e20000
10-10 20:25:19.243 18039  5031 I gralloc4: register: id=2800000865e, vaddr=0x795370a000
10-10 20:25:19.245 18039  5031 I gralloc4: register: id=2800000865f, vaddr=0x794ae2e000
10-10 20:25:19.248 18039  5031 I gralloc4: register: id=28000008660, vaddr=0x794a9b7000
[...]

PR #690, opened 2023-10-14
Enable sorting by updated date in LocalList

Luna712:
Refs #678


fire-light42:
@Blatzar why was it removed in the first place

EHMilon:
SIMKl Update (new to old) and Update (old to new) short is not fixed.

Luna712:
> SIMKl Update (new to old) and Update (old to new) short is not fixed.

This was not intended to fix that part yet. I didn't know the immediate issue there but did for the locallist sorting so just did that for now.

PR #691, opened 2023-10-14
Warn of potential duplicates when adding to library

### LINKED ISSUE #622: Duplicate title warning when adding a title to library (from another source)

If i have a title already in library like '"the blacklist" & i try to add it again from different source than add option to skip it and replace it (with existed) other than adding another one with same title.

A screenshot of same thing added from tachiyomi manga app

### Other details

Screenshot_20230922-144320_Tachiyomi

### Acknowledgements

### LINKED ISSUE #684: Utilize IMDb ID when available when saving items to the library

If it is available for specific entries, I think using the IMDb ID to save items to bookmarks/library would be a good idea. This would help prevent duplicates if they are the same entry. For providers that don't use the IMDb ID something like #622 would be a very useful.

### Other details

_No response_

### Acknowledgements

Luna712:
* Resolves #622
* Resolves #684

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
@@ -26,7 +26,6 @@ import com.lagradost.cloudstream3.CommonActivity.getCastSession
 import com.lagradost.cloudstream3.CommonActivity.showToast
 import com.lagradost.cloudstream3.LoadResponse.Companion.addTrailer
 import com.lagradost.cloudstream3.LoadResponse.Companion.getAniListId
-import com.lagradost.cloudstream3.LoadResponse.Companion.getImdbId
 import com.lagradost.cloudstream3.LoadResponse.Companion.getMalId
 import com.lagradost.cloudstream3.LoadResponse.Companion.isMovie
 import com.lagradost.cloudstream3.metaproviders.SyncRedirector
@@ -135,7 +134,7 @@ data class ResultData(
 data class CheckDuplicateData(
     val name: String,
     val year: Int?,
-    val imdbId: String?
+    val syncData: Map<String, String>?
 )
 
 enum class LibraryListType {
@@ -851,7 +850,7 @@ class ResultViewModel2 : ViewModel() {
             CheckDuplicateData(
                 name = response.name,
                 year = response.year,
-                imdbId = response.getImdbId(),
+                syncData = response.syncData,
             ),
             bookmarkedData
         ) { shouldContinue: Boolean, duplicateIds: List<Int?> ->
@@ -927,7 +926,7 @@ class ResultViewModel2 : ViewModel() {
                 CheckDuplicateData(
                     name = response.name,
                     year = response.year,
-                    imdbId = response.getImdbId(),
+                    syncData = response.syncData,
                 ),
                 getAllSubscriptions(),
             ) { shouldContinue: Boolean, duplicateIds: List<Int?> ->
@@ -994,7 +993,7 @@ class ResultViewModel2 : ViewModel() {
                 CheckDuplicateData(
                     name = response.name,
                     year = response.year,
-                    imdbId = response.getImdbId(),
+                    syncData = response.syncData,
                 ),
                 getAllFavorites(),
             ) { shouldContinue: Boolean, duplicateIds: List<Int?> ->
@@ -1044,6 +1043,7 @@ class ResultViewModel2 : ViewModel() {
         data: List<DataStoreHelper.LibrarySearchResponse>,
         checkDuplicatesCallback: (shouldContinue: Boolean, duplicateIds: List<Int?>) -> Unit
     ) {
+        val whitespaceRegex = "\\s+".toRegex()
         fun normalizeString(input: String): String {
             /**
              * Trim the input string and replace consecutive spaces with a single space.
@@ -1051,15 +1051,22 @@ class ResultViewModel2 : ViewModel() {
              * and one provider has the title with an extra whitespace. This is minor enough that
              * it should still match in this case.
              */
-            val regex = "\\s+".toRegex()
-            return input.trim().replace(regex, " ")
+            return input.trim().replace(whitespaceRegex, " ")
         }
 
         val duplicateEntries = data.filter {
-            checkDuplicateData.imdbId != null &&
-                    getImdbIdFromSyncData(it.syncData) == checkDuplicateData.imdbId ||
-                    normalizeString(it.name) == normalizeString(checkDuplicateData.name) &&
-                    it.year == checkDuplicateData.year
+            val syncData = checkDuplicateData.syncData
+            val librarySyncData = it.syncData
+
+            val checks = listOf(
+                { getImdbIdFromSyncData(syncData) != null && getImdbIdFromSyncData(librarySyncData) == getImdbIdFromSyncData(syncData) },
+                { getTMDbIdFromSyncData(syncData) != null && getTMDbIdFromSyncData(librarySyncData) == getTMDbIdFromSyncData(syncData) },
+                { syncData?.get(AccountManager.malApi.idPrefix) != null && librarySyncData?.get(AccountManager.malApi.idPrefix) == syncData[AccountManager.malApi.idPrefix] },
+                { syncData?.get(AccountManager.aniListApi.idPrefix) != null && librarySyncData?.get(AccountManager.aniListApi.idPrefix) == syncData[AccountManager.aniListApi.idPrefix] },
+                { normalizeString(it.name) == normalizeString(checkDuplicateData.name) && it.year == checkDuplicateData.year }
+            )
+
+            checks.any { it() }
         }
 
         if (duplicateEntries.isEmpty() || context == null) {
@@ -1125,6 +1132,14 @@ class ResultViewModel2 : ViewModel() {
             SimklApi.readIdFromString(
                 syncData?.get(AccountManager.simklApi.idPrefix)
             )[SimklApi.Companion.SyncServices.Imdb]
+        }
+    }
+
+    private fun getTMDbIdFromSyncData(syncData: Map<String, String>?): String? {
+        return normalSafeApiCall {
+            SimklApi.readIdFromString(
+                syncData?.get(AccountManager.simklApi.idPrefix)
+            )[SimklApi.Companion.SyncServices.Tmdb]
         }
     }
 
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
@@ -29,7 +29,6 @@ import com.lagradost.cloudstream3.LoadResponse.Companion.getMalId
 import com.lagradost.cloudstream3.LoadResponse.Companion.isMovie
 import com.lagradost.cloudstream3.metaproviders.SyncRedirector
 import com.lagradost.cloudstream3.mvvm.*
-import com.lagradost.cloudstream3.services.SubscriptionWorkManager
 import com.lagradost.cloudstream3.syncproviders.AccountManager
 import com.lagradost.cloudstream3.syncproviders.SyncAPI
 import com.lagradost.cloudstream3.syncproviders.providers.Kitsu
@@ -83,11 +82,6 @@ import kotlinx.coroutines.*
 import java.io.File
 import java.util.concurrent.TimeUnit
 
-interface AlertDialogResponseCallback {
-    fun onUserResponse(action: Boolean)
-
-    fun onUserResponseReplace(duplicateId: Int)
-}
 
 /** This starts at 1 */
 data class EpisodeRange(
@@ -463,33 +457,6 @@ class ResultViewModel2 : ViewModel() {
             return this?.firstOrNull { it.season == season }
         }
 
-        fun updateWatchStatus(currentResponse: LoadResponse, status: WatchType) {
-            val currentId = currentResponse.getId()
-
-            val currentWatchType = getResultWatchState(currentId)
-
-            setResultWatchState(currentId, status.internalId)
-            val current = getBookmarkedData(currentId)
-            val currentTime = System.currentTimeMillis()
-            setBookmarkedData(
-                currentId,
-                DataStoreHelper.BookmarkedData(
-                    current?.bookmarkedTime ?: currentTime,
-                    currentTime,
-                    currentId,
-                    currentResponse.name,
-                    currentResponse.year,
-                    currentResponse.url,
-                    currentResponse.apiName,
-                    currentResponse.type,
-                    currentResponse.posterUrl
-                )
-            )
-            if (currentWatchType != status) {
-                MainActivity.bookmarksUpdatedEvent(true)
-            }
-        }
-
         private fun filterName(name: String?): String? {
             if (name == null) return null
             Regex("[eE]pisode [0-9]*(.*)").find(name)?.groupValues?.get(1)?.let {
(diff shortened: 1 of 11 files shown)


Luna712:
> Could not resolve all files for configuration ':app:prereleaseDebugCompileClasspath'.
   > Could not find com.github.Blatzar:NiceHttp:0.4.3.
     Required by:
         project :app
         
not sure what is up with it failing sometimes.

Luna712:
* TODO: #684, but for now this is ready for review

This might be a very messy way to do this and might be much better ways, but I'm not sure. If this is a total mess, I apologize.

Luna712:
I updated this to be much more clean (in my opinion) by reducing duplicate code through use of better callbacks, and cleaning up messages in dialog to be more clear also.

Luna712:
I've tested almost every possibility I can think of on both TV and Phone for Subscriptions, Favorites, and each and every bookmark option (both in viewer and homepage preview), but there is still a chance I missed something. This works pretty consistently and accurately from my observations.

Luna712:
This now does the IMDb ID checking support as well.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -841,71 +882,188 @@ class ResultViewModel2 : ViewModel() {
...
+        checkDuplicatesCallback: (shouldContinue: Boolean, duplicateId: Int?) -> Unit
+    ) {
+        fun normalizeString(input: String): String {
+            /**
+             * Trim the input string and replace consecutive spaces with a single space.
+             * This covers some edge-cases where the title does not match exactly across providers,
+             * and one provider has the title with an extra whitespace. This is minor enough that
+             * it should still match in this case.
+             */
+            return input.trim().replace("\\s+".toRegex(), " ")
```
fire-light42, line 1024: this recreates the regex for every single run of the function, and that can be really expensive if you want to check multiple items. in this case it is about 30x faster to cache the regex in a persistent variable

app/src/main/res/values/strings.xml
```diff
@@ -686,13 +686,34 @@
...
     <string name="favorite_added">%s added to favorites</string>
     <string name="favorite_removed">%s removed from favorites</string>
     <string name="action_add_to_favorites">Add to favorites</string>
     <string name="action_remove_from_favorites">Remove from favorites</string>
+
+    <string name="duplicate_title">Potential Duplicate Found</string>
+    <string name="duplicate_add">Add</string>
+    <string name="duplicate_replace">Replace</string>
+    <string name="duplicate_cancel" translatable="false">@string/sort_cancel</string>
+    <string name="duplicate_message_bookmarks">
```
fire-light42, line 701: 3 nearly identical messages makes no sense, it takes a lot longer to translate that into like 50 languages. Keep it to 1 and use %s to fill the type, that way we can add this to future stuff and keep it easier to translate

fire-light42 (changes requested):
Very nice :+1:, however it is not very futureproof nor support any other service than imdb

Luna712:
> Very nice 👍, however it is not very futureproof nor support any other service than imdb

Thanks for your review!
Though, this does support even if not using Imdb, it then checks if title is an exact match, could support tmdb but I though that might complicate it as imdb and tmdb would contain fairlt similar stuff so didn't see why we would need to check both.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -352,20 +352,35 @@ object DataStoreHelper {
...
         @JsonProperty("apiName") override val apiName: String,
-        @JsonProperty("type") override var type: TvType? = null,
+        @JsonProperty("type") override var type: TvType?,
         @JsonProperty("posterUrl") override var posterUrl: String?,
-        @JsonProperty("year") val year: Int?,
-        @JsonProperty("quality") override var quality: SearchQuality? = null,
-        @JsonProperty("posterHeaders") override var posterHeaders: Map<String, String>? = null,
-    ) : SearchResponse {
+        @JsonProperty("year") open val year: Int?,
+        @JsonProperty("imdbId") open val imdbId: String?,
```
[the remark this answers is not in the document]
Luna712: That's not a bad idea, makes it decent for future, I didn't think about that. Thanks!

app/src/main/res/values/strings.xml
Luna712: The reason I kept it separate was the subtle differences between each of the 3, in particular subscriptions, originally had it as a combined message but depends on the action it didn't make to much sense to be combined, if you want I can go back to that again though and figure something out.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -822,9 +812,60 @@ class ResultViewModel2 : ViewModel() {
     val selectPopup: LiveData<SelectPopup?> = _selectPopup
 
 
-    fun updateWatchStatus(status: WatchType) {
-        updateWatchStatus(currentResponse ?: return, status)
-        _watchStatus.postValue(status)
+    fun updateWatchStatus(
+        status: WatchType,
+        context: Context?,
```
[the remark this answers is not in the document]
Luna712: Using a global context results in an exception: "java.lang.IllegalStateException: You need to use a Theme.AppCompat theme (or descendant) with this activity." unless you mean something else?

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -352,20 +352,35 @@ object DataStoreHelper {
...
+        @JsonProperty("lastSeenEpisodeCount") val lastSeenEpisodeCount: Map<DubStatus, Int?>,
+        override var id: Int?,
+        override val latestUpdatedTime: Long,
+        override val name: String,
+        override val url: String,
+        override val apiName: String,
+        override var type: TvType?,
+        override var posterUrl: String?,
+        override val year: Int?,
+        override val imdbId: String?,
```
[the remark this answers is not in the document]
Luna712: If you are always passing something to it doesn't it become null from what you are using if it does not exist? The value is always passed so I thought it would work, might be wrong as the whole json thing kinda confuses me, so I can do that if needed.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -572,6 +589,16 @@ object DataStoreHelper {
         return getKey("$currentAccount/$RESULT_WATCH_STATE_DATA", id.toString())
     }
 
+    fun getBookmarkedDataByWatchType(watchType: WatchType): List<BookmarkedData> {
```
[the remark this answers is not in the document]
Luna712: I'm confused, getBookmarkedData requires an ID, not a WatchType that we are using here to list all the entries of a specific watchtype in bookmarks unless I can get that some other way? Sorry, might just me being dumb here.

fire-light42:
> > Very nice 👍, however it is not very futureproof nor support any other service than imdb
> 
> Thanks for your review! Though, this does support even if not using Imdb, it then checks if title is an exact match, could support tmdb but I though that might complicate it as imdb and tmdb would contain fairlt similar stuff so didn't see why we would need to check both.

the point is not to do overlapping support like tmdb + imdb, but to have support for ANY service and in this case MAL would cover what imdb does not

Luna712:
> > > Very nice 👍, however it is not very futureproof nor support any other service than imdb
> > 
> > 
> > Thanks for your review! Though, this does support even if not using Imdb, it then checks if title is an exact match, could support tmdb but I though that might complicate it as imdb and tmdb would contain fairlt similar stuff so didn't see why we would need to check both.
> 
> the point is not to do overlapping support like tmdb + imdb, but to have support for ANY service and in this case MAL would cover what imdb does not

I understand, the issue I ran into supporting all syncData is Imdb is in simkl as a text of both tmdb and imdb, parsing through this while supporting non-simkl values makes the code much more complicated and I couldn't really figure it out. I'll try some more though.

app/src/main/res/values/strings.xml
Luna712: I went ahead and did this now, though it still has two: one for single duplicate found, and one for multiple

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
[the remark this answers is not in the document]
Luna712: Thanks! Done.

Luna712:
> Very nice 👍, however it is not very futureproof nor support any other service than imdb

@fire-light42 

Thanks again for the review, I think I got everything now!

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -113,6 +132,18 @@ data class ResultData(
     val plotHeaderText: UiText,
 )
 
+data class CheckDuplicateData(
+    val name: String,
+    val year: Int?,
+    val imdbId: String?
```
fire-light42, line 138: make this include the entire syncdata mapping

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -841,71 +905,227 @@ class ResultViewModel2 : ViewModel() {
...
+        checkDuplicatesCallback: (shouldContinue: Boolean, duplicateIds: List<Int?>) -> Unit
+    ) {
+        fun normalizeString(input: String): String {
+            /**
+             * Trim the input string and replace consecutive spaces with a single space.
+             * This covers some edge-cases where the title does not match exactly across providers,
+             * and one provider has the title with an extra whitespace. This is minor enough that
+             * it should still match in this case.
+             */
+            val regex = "\\s+".toRegex()
```
fire-light42, line 1054: this is not a persistent variable

fire-light42 (commented):
:pray: This is much better, but I would still like to see mal, anilist and tmdb id check before merge

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
Luna712: Thanks! Done.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
Luna712: Whoops, sorry. Done.

Luna712:
> 🙏 This is much better, but I would still like to see mal, anilist and tmdb id check before merge

Thanks again, done!

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -841,71 +904,243 @@ class ResultViewModel2 : ViewModel() {
...
+
+            return _subscribeStatus.value
+        }
     }
 
     /**
      * @return true if added to favorites, false if not. Null if not possible to favorite.
      **/
-    fun toggleFavoriteStatus(): Boolean? {
+    fun toggleFavoriteStatus(
```
fire-light42, line 975: this function should not return anything as it uses the callback

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -841,71 +904,243 @@ class ResultViewModel2 : ViewModel() {
...
+             * Trim the input string and replace consecutive spaces with a single space.
+             * This covers some edge-cases where the title does not match exactly across providers,
+             * and one provider has the title with an extra whitespace. This is minor enough that
+             * it should still match in this case.
+             */
+            return input.trim().replace(whitespaceRegex, " ")
+        }
+
+        val duplicateEntries = data.filter {
+            val syncData = checkDuplicateData.syncData
```
fire-light42, line 1058: move this outside the filter as well as all getImdbIdFromSyncData(librarySyncData) to reduce the amount of calls as this is quite a hot loop for many items this will improve the performance a bit

fire-light42: that way we dont have to call normalizeString, hashmap get and getImdbIdFromSyncData getTMDbIdFromSyncData on every single item

fire-light42 (commented):
Minor code quality things then merge. I am somewhat extra strict here with performance as we don't want a 300ms or so delay when clicking just because we have 1000 items in our library

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
fire-light42: it goes without saying that the comment needs updating as well

Luna712: Thanks! Done.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
Luna712: Thanks! Done.

Luna712:
> Minor code quality things then merge. I am somewhat extra strict here with performance as we don't want a 300ms or so delay when clicking just because we have 1000 items in our library

Thanks, and I appreciate that. It should be done again now.

fire-light42:
When testing this I found 2 UX issues

1. We just want to match on duplicates no matter the status. This might have been my confusing comment about it 3 days ago that made you change it, but just write smth like
`    fun getAllBookmarkedData(): List<BookmarkedData> = getKeys("$currentAccount/$RESULT_WATCH_STATE_DATA")?.mapNotNull(::getKey) ?: emptyList()
`
2. We only want to do the check when the watch status when it goes from None -> something else, we have already confirmed that we want a series bookmarked if we have on watching and move it to dropped for example.

Other than that it looks fine 👍 good job.

Luna712:
> When testing this I found 2 UX issues
> 
> 1. We just want to match on duplicates no matter the status. This might have been my confusing comment about it 3 days ago that made you change it, but just write smth like
>    `   fun getAllBookmarkedData(): List<BookmarkedData> = getKeys("$currentAccount/$RESULT_WATCH_STATE_DATA")?.mapNotNull(::getKey) ?: emptyList()`
> 2. We only want to do the check when the watch status when it goes from None -> something else, we have already confirmed that we want a series bookmarked if we have on watching and move it to dropped for example.
> 
> Other than that it looks fine 👍 good job.

Got it thanks, I'll finalize that i  a couple hours. Sorry about that.

Luna712:
> When testing this I found 2 UX issues
> 
> 1. We just want to match on duplicates no matter the status. This might have been my confusing comment about it 3 days ago that made you change it, but just write smth like
>    `   fun getAllBookmarkedData(): List<BookmarkedData> = getKeys("$currentAccount/$RESULT_WATCH_STATE_DATA")?.mapNotNull(::getKey) ?: emptyList()`
> 2. We only want to do the check when the watch status when it goes from None -> something else, we have already confirmed that we want a series bookmarked if we have on watching and move it to dropped for example.
> 
> Other than that it looks fine 👍 good job.

@fire-light42 

I *think* I got it none, had some issues with how NONE was done, which saved to data regardless (seems like even before it did) so counted as a duplicate but I think I fixed that, and made it remove properly, before it seems like it was removed and immediately readded, which I have no idea if that was intentional or not, but it does not do that anymore. If you want it to, I can figure something else out.

fire-light42 (approved):

Luna712:
Thanks again for all your reviewing and suggestions @fire-light42

PR #692, opened 2023-10-16
Old APIs focus fixes

KingLucius:
Fixes the below focus behavior for old APIs (tested for API 25) in TV layout:
- Movie button focus in result layout.
- Cast item focus in result layout.
- Media description focus in result layout.
- Switch account button focus, in Homepage during loading.


app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -575,6 +575,7 @@ class ResultFragmentTv : Fragment() {
         observeNullable(viewModel.movie) { data ->
             binding?.apply {
                 resultPlayMovie.isVisible = data is Resource.Success
+                resultPlaySeries.isVisible = data == null
```
fire-light42, line 578: Series button visible on null movie? why

KingLucius: I am hidding it for movie, as it takes the focus for movie

PR #693, opened 2023-10-16
Poster cropped at 20% from Top

KingLucius:


fire-light42 (approved):
Code looks ok, but not tested

PR #694, opened 2023-10-16
chore: update nicehttp and fix artifact push

IndusAryan:


Luna712:
Thanks! I was planning to do this also, as I had to do it in #691 because it kelt failing the build and when I noticed it failing for everything else I thought it would be a good idea to split it so thanks!

PR #695, opened 2023-10-16
Fix scroll for Library TV layout

KingLucius:


fire-light42 (changes requested):
NestedScrollView = very bad as it has to render every single item that exist as it cant recycle. Explain really well why this is nessessary and why you can fix it in any other way or else this will never be merged

KingLucius:
The layouts description says if you have multi scrolls in same layout then usr nestedscrollview, i am open to any other suggestion no issue

KingLucius:
Updated the fix without NestedScrollView

fire-light42 (approved):

PR #699, opened 2023-10-20
Enable PreviewGenerator on TV layout

Luna712:
Because I like using the full TV layout on emulator (like BlueStacks), and it looks and works a bit nicer for me, the sole reason I don't anymore is that the preview doesn't appear.

If you don't want it enabled that's fine I suppose, I can live without the TV layout

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/PreviewGenerator.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/PreviewGenerator.kt
@@ -9,7 +9,6 @@ import android.util.Log
 import androidx.annotation.WorkerThread
 import androidx.core.graphics.scale
 import com.lagradost.cloudstream3.mvvm.logError
-import com.lagradost.cloudstream3.ui.settings.SettingsFragment
 import com.lagradost.cloudstream3.utils.Coroutines.ioSafe
 import com.lagradost.cloudstream3.utils.ExtractorLink
 import com.lagradost.cloudstream3.utils.ExtractorLinkType
@@ -62,12 +61,7 @@ interface IPreviewGenerator {
 
     companion object {
         fun new(): IPreviewGenerator {
-            /** because TV has low ram + not show we disable this for now */
-            return if (SettingsFragment.isTrueTvSettings()) {
-                empty()
-            } else {
-                PreviewGenerator()
-            }
+            return PreviewGenerator()
         }
 
         fun empty(): IPreviewGenerator {


Luna712:
> i think blue guy decided to stop showing previews in tv as they were causing multiple crashes


Hmm, I have it on 3 different TVs, including Fire TV and Google TV and a few emulators and it always worked fine. But I wonder if it should be a preference, I can see a use case for it to be a preference even on phones due to potential for data usage as well.

fire-light42:
This solved nothing, and only makes the problem worse as isTrueTvSettings == is on a physical TV not emulator. the preview is still on the emulator with the current code, so if you are using an emulator it should work rn

Blatzar:
> > i think blue guy decided to stop showing previews in tv as they were causing multiple crashes
> 
> Hmm, I have it on 3 different TVs, including Fire TV and Google TV and a few emulators and it always worked fine. But I wonder if it should be a preference, I can see a use case for it to be a preference even on phones due to potential for data usage as well.

That's the wonders of android TV development, it works when you test it but then you get a lot of reports of crashing :sob: 
In this case I think it was OOM for very low-end TVs, nothing you would get on more recent android TVs or Emulators.
Everything for Android TV needs to be designed to be used by devices from the 1900s :sob:

Adding it as a preference was very likely decided against, because it was such a niche use-case (TV layout on emulator when there's an emulator layout). There's already too many settings, more would need good reason. At the very best you might add more emulation detection to override this behavior. Do note that you cannot use touch events (detecting a mouse) securely as I am not sure all TVs conform to anything regarding touch.

Luna712:
@Blatzar I completely understand that. I do think there is a valud use case for it to be a setting due to the potential data usage though. Maybe it doesn't use as much as I thought but my data usage sky rocketed after it was in CS. It doesn't matter to me as I have unlimited but for some that might be an issue. It mixing with potential OOMs provides a fairly decent use case some might want it disabled. But I completely understand your reasoning for keeping it disabled on TV for now.

Luna712:
> This solved nothing, and only makes the problem worse as isTrueTvSettings == is on a physical TV not emulator. the preview is still on the emulator with the current code, so if you are using an emulator it should work rn

@fire-light42 

I missed this, but that is incorrect from my understanding. IsTrueTvSettings is defined to check if getLayoutInt is 1, if it is, it is a TV, getLayoutInt is manual setting of layout also. -1 is auto, and it also checks that, but even if manual change of layout to TV it still disables. I opened #701 to only disable if it is really a detected TV and not layout setting also.

PR #698, opened 2023-10-20
Add more info on homepage to emulator layout

Luna712:
Fixes #662 

It does some manual changes to avoid having to duplicate the entire layout for minor changes

Eventually things like this, in particular visibility might do good by using DataBinding, but for now I just did some manual layout changes (outside XML) so that we did not have to duplicate 350 lines of layout code for very few changes. We can keep the same TV layout just make some changes for proper emulator support.

**Part 1 of emulator layout improvements**

PR #701, opened 2023-10-20
Only disable PreviewGenerator if it is really a TV

Luna712:
That way it can still be used if using the TV layout manually (IE on emulator)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/PreviewGenerator.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/PreviewGenerator.kt
@@ -9,7 +9,7 @@ import android.util.Log
 import androidx.annotation.WorkerThread
 import androidx.core.graphics.scale
 import com.lagradost.cloudstream3.mvvm.logError
-import com.lagradost.cloudstream3.ui.settings.SettingsFragment
+import com.lagradost.cloudstream3.ui.settings.SettingsFragment.Companion.isAutoTv
 import com.lagradost.cloudstream3.utils.Coroutines.ioSafe
 import com.lagradost.cloudstream3.utils.ExtractorLink
 import com.lagradost.cloudstream3.utils.ExtractorLinkType
@@ -63,7 +63,7 @@ interface IPreviewGenerator {
     companion object {
         fun new(): IPreviewGenerator {
             /** because TV has low ram + not show we disable this for now */
-            return if (SettingsFragment.isTrueTvSettings()) {
+            return if (isAutoTv()) {
                 empty()
             } else {
                 PreviewGenerator()
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsFragment.kt
@@ -33,6 +33,7 @@ class SettingsFragment : Fragment() {
         var beneneCount = 0
 
         private var isTv: Boolean = false
+        private var isAutoTv: Boolean = false
         private var isTrueTv: Boolean = false
 
         fun PreferenceFragmentCompat?.getPref(id: Int): Preference? {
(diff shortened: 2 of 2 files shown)


Luna712:
The other option is doing something like https://stackoverflow.com/a/21505193/22641654 but this just seemed cleaner, if you'd rather I could go the other route.

fire-light42:
NONONONO AutoTV is **ONLY** for setup as a hint. the worst tv devices wont get matched on autoTV and the layout settings are used to override **ALL** behavior no matter what type of device it is. If you really really want to have your previews you will have to use emulator layout and this wont change no matter what

Luna712:
> NONONONO AutoTV is **ONLY** for setup as a hint. the worst tv devices wont get matched on autoTV and the layout settings are used to override **ALL** behavior no matter what type of device it is. If you really really want to have your previews you will have to use emulator layout and this wont change no matter what

What if I make an emulator layout autodetect and then check that like @Blatzar suggested in #699? Again I understand, I just like TV layout on PC and I like to use the previews for it also.

The purose behind it not being enabled on TV is RAM, on emulator, this is unlikely an issue, therefore I do not see the logic of forcing it be disabled based on the TV layout when the layout itself is not the issue but the OS behind it. It should not be forced to disabled no matter what you do when using a different type of device with TV layout, that supports it, since another reason ir was even disabled is it doesn't show on TV due to how the seek works, this is again, not an issue on emulator and it is supposed to work fine there. If I have to I guess I can fork CS temporarily or use a custom debug version of one of these patches to enable it, but I iust wanted to avoid that for obvious issues. I do understand your reasoning for not wanting to though.

Blatzar:
> therefore I do not see the logic of forcing it be disabled based on the TV layout when the layout itself is not the issue but the OS behind it.

It goes by this logic:

1. The TV layout is supposed to be used by TVs using Dpad navigation.
2. Dpad navigation does not support preview (currently).
3. If users cannot see a preview why waste resources? Especially if it crashes.

It was disabled because of the OS, but remains disabled not only because of the OS.

Blatzar:
I think a satisfactory solution would be something like

```
return if (SettingsFragment.isTrueTvSettings() && !someIsEmulatorMethod())
```

This makes sure all TVs get covered, but some emulators get explicitly whitelisted, instead of the other way around.
You cannot guarantee that all TVs will be 'auto tv', but you can be pretty sure Emulators are Emulators  (no real TV will have device model bluestacks).

Blatzar:
@fire-light42 opinions?

Luna712:
> > therefore I do not see the logic of forcing it be disabled based on the TV layout when the layout itself is not the issue but the OS behind it.
> 
> It goes by this logic:
> 
> 1. The TV layout is supposed to be used by TVs using Dpad navigation.
> 2. Dpad navigation does not support preview (currently).
> 3. If users cannot see a preview why waste resources? Especially if it crashes.
> 
> It was disabled because of the OS, but remains disabled not only because of the OS.

@Blatzar 

Not all TVs use DPAD navigation either though. Some TVs including Google TV/Android TV and FireTV support a mouse and keyboard as well making it work properly for them, so my PoV is the primary reason it's required to be disabled regardless if using TV layout setting is OS. But I do understand, I'm not saying I don't, emulator layout is basically phone with some changes, and works even worse, phone layout just doesn't look right (but the one I am forced to use right now, though emulator layout works a bit better with #698), but TV layout itself, particularly the movie/series view layout screen, works and looks a lot nicer on emulator for PC in my opinion, but I just like the preview so I'm forced to not, because for me it's even more important. But again, I understand the reasoning.

Luna712:
> I think a satisfactory solution would be something like
> 
> ```
> return if (SettingsFragment.isTrueTvSettings() && !someIsEmulatorMethod())
> ```
> 
> This makes sure all TVs get covered, but some emulators get explicitly whitelisted, instead of the other way around. You cannot guarantee that all TVs will be 'auto tv', but you can be pretty sure Emulators are Emulators (no real TV will have device model bluestacks).

Bluestacks doesn't have device model bluestacks either, it is a bit tricky, manageable but tricky and then youd have to maintain a list and update models if they change, etc... so I also wanted to avoid that, as I thought it might make it even harder on you, but there might be some fairly reliable way to do it also.

Luna712:
Maybe an alternative could just be improvements to the emulator layout itself so it works good and then this wont matter.

fire-light42:
Emulator layout is quite literately made for emulator, I dont see the issue

Blatzar:
> Maybe an alternative could just be improvements to the emulator layout itself so it works good and then this wont matter.

Yeah that would be the best option :pray:

PR #702, opened 2023-10-21
Library/LocalList: enable subscriptions on emulator layout

Luna712:


fire-light42:
What is the point of subscriptions on an emulator? Does the background worker even work on an emulator?

Luna712:
> What is the point of subscriptions on an emulator? Does the background worker even work on an emulator?

It does indeed work. And sends desktop notifications for it, at least with BlueStacks

PR #703, opened 2023-10-21
Bump libraries

Luna712:
Unless we change targetSdk this doesn't enable runtime changes for android 14, only changes for compiling, never affecting actual runtime. However I can not be certain a specific library will not cause some issues, bumping compileSdk itself should not cause runtime issues as compilesdk is not used on runtime. This only really affects compiling and the primary reason for this is fixinf navigation issues:

* Double  backpress causing navigation graph to be lost
* Sometimes on TV clicking next episode or it automicilly going to next episode at the very end results in it creating a dual audio, the first one which can not be closed, without force stoping the app, as far as I can tell this happens fairly often before, but appears to not (or not as often/ I didn't properly reproduce this time). It seems the issue had something to do withglobal_to_navigation_playeI could be wrong), which this seemingly fixes.

This also reverts #688 since that had no effect, and personally that doesn't seem something to be global navigation, if you want me to undo reverting that I can.

If I am totally wrong in opening this, I apologize, please close it. Thanks!

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -50,14 +50,14 @@ android {
         }
     }
 
-    // https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading
-    compileSdk = 33 // android 14 is fucked
+    compileSdk = 34
     buildToolsVersion = "34.0.0"
 
     defaultConfig {
         applicationId = "com.lagradost.cloudstream3"
         minSdk = 21
-        targetSdk = 33
+        // https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading
+        targetSdk = 33 // android 14 is fucked
 
         versionCode = 62
         versionName = "4.2.1"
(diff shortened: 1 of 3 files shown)


Luna712:
> this might cause some conflict in my kapt to ksp pr, i m doing upstream bumps slowly to checlayer also. At the very least I feel this should bump navigation-fragment-ktx first, as it seems to cause a bunch of issues.k and identify any bugs easily, and reverting the navigation graph is not needed i ve already fixed the settings crash by nullsafing when nav forgets the past fragment will clean gradle and remove kapt after that merge and slowly upgrade

It doesn't just happen with settings it is possible to happen from main navigation and player also. At the very least I feel this should bump navigation-fragment-ktx first, as it seems to cause a bunch of issues.

I'm fine with closing this and leaving bumping to someone else, but I do feel navigation should be bumped first if possible.

Luna712:
Also I purposefully left glide alone as to try and avoid conflicts with you here.

PR #704, opened 2023-10-21
Show confirm exit dialog on emulator layout

Luna712:
**Part 3 of emulator layout improvements**


Luna712:
> 🤨 Is that not just more annoying?

This is not absolutely needed, it is just annoying when I accidentally click he esc button to many times and close the app.

PR #705, opened 2023-10-21
Support showing content ratings for TmdbProvider

### LINKED ISSUE #613: Support for displaying content rating

It would be nice if content rating could be added to display. That way it makes it easier than manually checking something like IMDB for it.

### Other details

_No response_

### Acknowledgements

Luna712:
Resolves #613


app/src/main/java/com/lagradost/cloudstream3/metaproviders/TmdbProvider.kt
```diff
@@ -264,6 +268,51 @@ open class TmdbProvider : MainAPI() {
...
+                if (!certification.isNullOrEmpty()) {
+                    return certification
+                }
+            }
+        }
+
+        return null
+    }
+
+    private fun findFirstNonEmptyCertification(releaseDates: List<ReleaseDate>?): String? {
```
Blatzar, line 305: Take advantage of kotlin functions :pray: 

`list.firstOrNull { !it.isNullOrBlank() }`

(.find is the same thing)

app/src/main/java/com/lagradost/cloudstream3/metaproviders/TmdbProvider.kt
```diff
@@ -264,6 +268,51 @@ open class TmdbProvider : MainAPI() {
...
+                null
+            }
+        }
+    }
+
+    private fun findFirstNonEmptyCertificationWithCountry(
+        releaseDatesResults: List<ReleaseDatesResult>?,
+        country: String
+    ): String? {
+        for (releaseDateResult in releaseDatesResults.orEmpty()) {
```
Blatzar, line 293: here's also a great place to use firstOrNull/find

Blatzar (commented):
don't mind me, just spreading the niceness of the built in kotlin functions :drooling_face: 

Nothing you need to change

app/src/main/java/com/lagradost/cloudstream3/metaproviders/TmdbProvider.kt
Luna712: Thanks! I was able to remove both these functions by doing that making it much more simpler!

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
```diff
@@ -700,6 +701,11 @@ open class ResultFragmentPhone : FullScreenPlayer() {
                     resultCastItems.isGone = d.actors.isNullOrEmpty()
                     (resultCastItems.adapter as? ActorAdaptor)?.updateList(d.actors ?: emptyList())
 
+                    if (d.contentRatingText == null) {
+                        // If there is no rating to display, we don't want an empty gap
+                        resultMetaContentRating.width = 0
```
fire-light42, line 706: :raised_eyebrow: what if the contentRatingText gets populated after it was null, does it work then?

Luna712: > 🤨 what if the contentRatingText gets populated after it was null, does it work then?

It shouldn't be populated after it is null, Ive never encountered it happening anyway, but we can add something to ensure that I suppose if needed?

Luna712: If it is null here, then it would be null a few lines earlier when it does `resultMetaContentRating.setText(d.contentRatingText)` so it wouldn't display anything anyways and this should trigger removing the width was my understanding anyway.

Luna712:
Here is a much more advanced fetchContentRating that will display content ratings even if they don't exist in a particular country as fallback, similar to how themoviedb.org does itself, however I did not use it as I could not decide on if it is something to do or not due to its complexity but I am just posting here for the record. I can add it if we want to?
```kotlin
  open suspend fun fetchContentRating(id: Int?, preferredCountry: String, isTvSeries: Boolean): String? {
      id ?: return null
      return if (isTvSeries) {
          val contentRatings = tmdb.tvService().content_ratings(id).awaitResponse().body()
          contentRatings?.results?.firstOrNull { it: ContentRating ->
              it.iso_3166_1 == preferredCountry
          }?.rating ?: contentRatings?.results?.firstOrNull()?.rating
      } else {
          val releaseDates = tmdb.moviesService().releaseDates(id).awaitResponse().body()?.results

          val preferredCountryCertification = releaseDates?.firstOrNull { it: ReleaseDatesResult ->
              it.iso_3166_1 == preferredCountry
          }?.release_dates?.firstOrNull { it: ReleaseDate ->
              !it.certification.isNullOrBlank()
          }?.certification

          if (!preferredCountryCertification.isNullOrBlank()) {
              preferredCountryCertification
          } else {
              val theatricalCertification = releaseDates?.firstOrNull {
                  it.release_dates?.get(0)?.type == ReleaseDate.TYPE_THEATRICAL
              }?.release_dates?.firstOrNull { it: ReleaseDate ->
                  !it.certification.isNullOrBlank()
              }?.certification


              if (!theatricalCertification.isNullOrBlank()) {
                  theatricalCertification
              } else {
                  val firstCertification = releaseDates?.firstOrNull()?.release_dates?.firstOrNull { it: ReleaseDate ->
                      !it.certification.isNullOrBlank()
                  }?.certification

                  firstCertification
              }
          }
      }
  }
  ```

**Comment by KingLucius** [comment]
Thanks for this PR, I need this so much

**Comment by Luna712** [comment]
> Thanks for this PR, I need this so much

You're very welcome. I have PRs ready for extensions to implement support to some like SoraStream as well but we'll see how far it goes as it will definitely need to be implemented in extensions for support, but there is a concern with that since LoadResponse.contentRating doesn't already exist if you add it before CS is updated it will break things. I have no idea how to handle that backwards compatibility so was thinking maybe after this is merged, wait a month or so before doing extensions or something like that. Maybe make a new stable release of CS also so it can be added there as well.

**Comment by KingLucius** [comment]
> > Thanks for this PR, I need this so much
> 
> You're very welcome. I have PRs ready for extensions to implement support to some like SoraStream as well but we'll see how far it goes as it will definitely need to be implemented in extensions for support, but there is a concern with that since LoadResponse.contentRating doesn't already exist if you add it before CS is updated it will break things. I have no idea how to handle that backwards compatibility so was thinking maybe after this is merged, wait a month or so before doing extensions or something like that. Maybe make a new stable release of CS also so it can be added there as well.

Tmdbprovider needs some love indeed 👍, you have to be carefull when modifying the mainapi and always think about back compatability.
Good luck 👍

**Comment by Luna712** [comment]
> > > Thanks for this PR, I need this so much
> > 
> > 
> > You're very welcome. I have PRs ready for extensions to implement support to some like SoraStream as well but we'll see how far it goes as it will definitely need to be implemented in extensions for support, but there is a concern with that since LoadResponse.contentRating doesn't already exist if you add it before CS is updated it will break things. I have no idea how to handle that backwards compatibility so was thinking maybe after this is merged, wait a month or so before doing extensions or something like that. Maybe make a new stable release of CS also so it can be added there as well.
> 
> Tmdbprovider needs some love indeed 👍, you have to be carefull when modifying the mainapi and always think about back compatability. Good luck 👍

Indeed, and thanks. I'm trying to think of anything else that might be good in the future to just add the variables now so at least we don't have to worry about the variables in the future again, but I can't think of anything else right now, can you?

**Comment by KingLucius** [comment]
> > > > Thanks for this PR, I need this so much
> > > 
> > > 
> > > You're very welcome. I have PRs ready for extensions to implement support to some like SoraStream as well but we'll see how far it goes as it will definitely need to be implemented in extensions for support, but there is a concern with that since LoadResponse.contentRating doesn't already exist if you add it before CS is updated it will break things. I have no idea how to handle that backwards compatibility so was thinking maybe after this is merged, wait a month or so before doing extensions or something like that. Maybe make a new stable release of CS also so it can be added there as well.
> > 
> > 
> > Tmdbprovider needs some love indeed 👍, you have to be carefull when modifying the mainapi and always think about back compatability. Good luck 👍
> 
> Indeed, and thanks. I'm trying to think of anything else that might be good in the future to just add the variables now so at least we don't have to worry about the variables in the future again, but I can't think of anything else right now, can you?

I can't think of something particular right now, i will check the tmdb apis what they are providing.

**Comment by Luna712** [comment]
I think maybe original language could be useful as it could 1) display the original language or even potentially allow filtering by it in the future using a language setting but not sure.

**Comment by fire-light42** [comment]
image

What updating mainapi does to a mf

**Comment by fire-light42** [comment]
So this probably wont get merged until a new stable release to not fuck with the current stable

**Comment by CranberrySoup** [comment]
Make another constructor with identical parameters to the old one and it will work :+1:

**Comment by Luna712** [comment]
> Make another constructor with identical parameters to the old one and it will work 👍

What do you mean?

**Comment by Luna712** [comment]
> So this probably wont get merged until a new stable release to not fuck with the current stable

@fire-light42 

When do you think a new stable release will be?

**Comment by CranberrySoup** [comment]
> > Make another constructor with identical parameters to the old one and it will work 👍
> 
> What do you mean?

Example: https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/ExtractorApi.kt#L412-L426

If the old constructor also remains it will be backwards compatible

**Comment by Luna712** [comment]
> image
> 
> What updating mainapi does to a mf

@fire-light42 

It should now maintain backwards compatibility

@CranberrySoup thanks a lot for the suggestion

File: app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt [hunk 4]
```diff
@@ -1193,6 +1193,9 @@ interface LoadResponse {
     var syncData: MutableMap<String, String>
     var posterHeaders: Map<String, String>?
     var backgroundPosterUrl: String?
+    var contentRating: String?
+        get() = throw NotImplementedError()
+        set(value) { contentRating = value }
```
Blatzar, line 1198: Is this WIP?
The setter is recursive and I am unsure how jackson will like exceptions on getters :pensive:

Luna712: > Is this WIP? The setter is recursive and I am unsure how jackson will like exceptions on getters 😔

@Blatzar 

I can just remove the exception and make it return nothing, without the get and set, contentRating is required and the constructors can't avoid it, with this it makes overriding it not required so I did it that why, unless maybe there is another way?

Blatzar: The setter will not work, it calls itself and might even crash the app due to infinite recursion.
If you fix the setter to `set(value) { field = value }` (redundant) it will complain with "Property in an interface cannot have a backing field".

Whatever you are doing is not how its supposed to be done language wise, you'll have to do it any other way :pensive: 

As for the getter: Exceptions in general should really really really be avoided in android as that kills the app immediately. If you ever need to return nothing then nullable is the way to go.

If you really want to crash on incorrect usage of something (not recommended) use debugAssert and it will not cause crashes in production if something is wrong.

Luna712: @Blatzar what about like:
```kt
var contentRating: String?
    get() = null
    set(value) { return }
```

Luna712:
I'm wondering if we should add descriptors for content ratings as well, this might allow some things in the future to use, I have quite a bit of ideas we could do to Tmdbprovider but the way MainApi works, not being very extendable makes it difficult as well...

fire-light42:
> I'm wondering if we should add descriptors for content ratings as well, this might allow some things in the future to use, I have quite a bit of ideas we could do to Tmdbprovider but the way MainApi works, not being very extendable makes it difficult as well...

The problem is that the devs are using the wrong constructor. they should be using newTvSeriesSearchResponse that is built for this exact reason or the several other ones to construct it, but some devs use the built in constructor and this causes this issue

Luna712:
> > I'm wondering if we should add descriptors for content ratings as well, this might allow some things in the future to use, I have quite a bit of ideas we could do to Tmdbprovider but the way MainApi works, not being very extendable makes it difficult as well...
> 
> The problem is that the devs are using the wrong constructor. they should be using newTvSeriesSearchResponse that is built for this exact reason or the several other ones to construct it, but some devs use the built in constructor and this causes this issue

Personally, I think that support to do it that way should be deprecated, and eventually removed. It causes a headache with CS development, and it isn't the way it should be done unless there is a real reason it has to be done the other way?

app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Nevermind I was not thinking straight, I do not believe this is even necessary

Luna712: I removed it @Blatzar and it still works fine, doing provider tests no longer show the error and they appear to work now with backward compatability still.

Blatzar (commented):
LGTM just needs testing to make sure the constructors are working :pray:

Luna712:
@Blatzar thoughts on me adding original language to TmdbProvider also (in a separate PR maybe)? Maybe also content rating descriptors. Mostly so we don't have to change MainAPI again for this, at the very least, even if we don't add it to TmdbProvider itself just yet, maybe just to MainAPI constructors so we don't have to worry about breaking extensions again later if/when we add to TmdbProvider? Also if there is anything else we might want to add to TmdbProvider later on, I was thinking adding now might not be a bad idea since changing MainAPI has so many potential complications...

Just a thought, content rating was always the most important one for me personally to get added anyway, so this is fine if nothing else.

Blatzar:
Only problem with original language I could see is that it would be confusing for extensions with another language dubbed. Then people might confuse that for the dub language :shrug: 
Having it say "Original Language: English" does not conform with how all the other data is displayed.

What would be really cool, but a lot harder to implement would be reviews.

Blatzar:
@fire-light42

Luna712:
@fire-light42 just bumping this... any update?

fire-light42:
> @fire-light42 just bumping this... any update?

Uhhh, we will see when stable drops

Luna712:
> > @fire-light42 just bumping this... any update?
> 
> Uhhh, we will see when stable drops

The point of this PR and backwards compatibility now is to support pre release also, so it can be done in stages, merge this, then stable, then update extensions to support it. But if you prefer to wait I am totally okay with that as well. Thank you! Backwards compatibility should probably be kept for at least one stable iteration anyway, maybe removed in the next stable update after the upcoming one but tnats up to you all also. I appreciate the response.

LagradOst:
Yes, a stable is needed

fire-light42:
Your mitigation for backwards compatibility was nice, however as suspected this also caused forwards compatibility issues where the providers updated but everyone on a lower version of the app cant use them. This is why we dont touch MainAPI unless we know that stable and prerelease users dont get screwed over for even for the most simple change.

Luna712:
> Your mitigation for backwards compatibility was nice, however as suspected this also caused forwards compatibility issues where the providers updated but everyone on a lower version of the app cant use them. This is why we dont touch MainAPI unless we know that stable and prerelease users dont get screwed over for even for the most simple change.

@fire-light42 

I wonder if we can add some sort of 'version' on MainAPI, that can be pulled and checked by extensions to allow them to maintain forward and backward compatibility in the future, using it would break it one last time and require updating, but it would be a way to prevent it again in the future I think. If not that, then I think something should be done to handle this and make it more extendable in the future...

PR #706, opened 2023-10-22
fix: crash when navigating to different fragment whilst in sub-fragment

IndusAryan:
please merge this before the ksp one!
---
the navigation graph still forgets its track but i am not able to reproduce the crash.


fire-light42:
Was about to close due to duplicate of #703 but it looks like you solved it. Code looks fine, however major testing needs to be done to really check if it does not fuck anything up.

fire-light42 (approved):

PR #709, opened 2023-10-24
Use bottom dialogs for synopsis

Luna712:
Only does for phone, not sure if it should for TV also, also only did this because of the TODO, if it is not wanted, this can be closed


fire-light42:
Nice change UI wise, however it does not keep the HTML text styling nor is the function you use used for this. Create a new function in SingleSelectionHelper called smth like showBottomDialogText and make it accept a Spanned as text and title

Luna712:
> Nice change UI wise, however it does not keep the HTML text styling nor is the function you use used for this. Create a new function in SingleSelectionHelper called smth like showBottomDialogText and make it accept a Spanned as text and title

@fire-light42 

Thanks, done (I think)

PR #710, opened 2023-10-25
Use TV results layout on emulator

Luna712:
It already uses TV layout partiually (It pushes the episodes to the left for example), but it does not do the rest that makes it look nice like it ideally should for results.

**Part 4 of emulator layout improvements**


fire-light42 (commented):
🤨 This is good and all, but I would not put subscriptions on true tv layout, only on emulator

Luna712:
> 🤨 This is good and all, but I would not put subscriptions on true tv layout, only on emulator

This does only put them on emulator, just in TV results fragment with an isEmulator check to avoid a lot of duplicated emulator code since it now uses the TV fragment

fire-light42:
> > 🤨 This is good and all, but I would not put subscriptions on true tv layout, only on emulator
> 
> This does only put them on emulator, just in TV results fragment with an isEmulator check to avoid a lot of duplicated emulator code since it now uses the TV fragment

uhhh, ye I saw that shit rn, srry

PR #712, opened 2023-10-25
Move favorites button above bookmarks on TV

Luna712:
I think this looks a bit nicer, but if you don't think it is needed, feel free to close this


Luna712:
I'm not actually 100% decided on this yet

Luna712:
Might reopen later if I decide this is a good thing, for now, closing though

notuser112:
Maybe move the episode list button above favorite and bookmark button as well. Thanks

Luna712:
> Maybe move the episode list button above favorite and bookmark button as well. Thanks

Yeah that's why I closed this, because it didn't look any better with the episode button there anyway. The order seems off, I just don't know the best way it really should be.

PR #713, opened 2023-10-25
Update TmdbProvider.kt

Jubes2011:


fire-light42:
The default behavior should always be non adult, if you really want this then make it follow the app settings

PR #714, opened 2023-10-25
Update build.gradle.kts

Jubes2011:


fire-light42:
Those should not be updated because the prerelease build system wont work. If you want to bump them you also need to fix the runners

PR #715, opened 2023-10-25
Update CMakeLists.txt

Jubes2011:


fire-light42:
cmake is not even used in the project. It was disabled

fire-light42:
https://github.com/recloudstream/cloudstream/blob/master/app/build.gradle.kts#L36

PR #716, opened 2023-10-25
Update MultiAnimeProvider.kt

Jubes2011:


fire-light42:
That is not even used anywhere, just old code that should be removed. no need updating that

PR #717, opened 2023-10-25
Delete old MultiAnimeProvider.kt

fire-light42:


fire-light42:
Remove this as it is not used anywhere. It is not even open so it should not even be used in any external providers

LagradOst (approved):

PR #725, opened 2023-10-26
Reload library when deleting bookmarks

Luna712:
Also removes unused deleteAllBookmarkedData


Luna712:
I forgot to mention what this was really for. It fixes an annoying little bug when removing bookmarks from the home page or changing them to NONE where they stay in library until the app is reatarted or something else triggers a refresh.

PR #726, opened 2023-10-26
Upgrade gradle

Luna712:
I was able to get it to build properly but I can not be certain if prerelease with GitHub actions will correctly work, but building prerelease manually works for me now. Please close if you do not want this done, in which case, I appologize.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -8,6 +8,7 @@
         <option name="testRunner" value="GRADLE" />
         <option name="distributionType" value="DEFAULT_WRAPPED" />
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
+        <option name="gradleJvm" value="jbr-17" />
         <option name="modules">
           <set>
             <option value="$PROJECT_DIR$" />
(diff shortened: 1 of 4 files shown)


Luna712:
> even if workflow fails, i think we need latest log to see whats wrong

@IndusAryan 

Indeed, unless we know what still fails, we can't really fix it, unless we use something like act to test locally, but that isn't always super reliable either. Since it worked for me to build locally unless we get latest log on why it fails, we can't know what is wrong and how to fix it.

fire-light42:
@C10udburst how much will this fuck over prereleases?

Luna712:
> bro/sis @Luna712 don't bump desugar lib plsl i am testing one thing!

Bumping it requires bumping gradle so I thought I would just do it all at once but okay, I will revert that.

Luna712:
> bro/sis @Luna712 don't bump desugar lib plsl i am testing one thing!

I got rid of that part

C10udburst:
> @C10udburst how much will this fuck over prereleases?

It will fuck it over, but I have found a way to fix it I think. Basically removing the dependency `build` from `makeJar` and running `makeJar` after `build` and `assemblePrerelease` (not concurrently). Don't worry, I'll make the changes when the PR gets merged/marked as ready to be merged.

C10udburst (approved):

Luna712:
This should be okay now.

fire-light42 (approved):
@C10udburst Merge anytime if you feel like the build system can handle it

Luna712:
Thanks @C10udburst

PR #733, opened 2023-10-28
Add support for profile locks

### LINKED ISSUE #604: Add support for profile locks in local accounts

I am not sure if this is something can really be done, but I am just postinfg a few my thoughts I have been thinking for a while, but it would be nice if you could set something like a profile pin for local accounts. Since it would likely be stored locally, it would not be something meant to be extremely secure, but just a way to have a little extra control on profiles.

### Other details

_No response_

### Acknowledgements

Luna712:
Resolves #604

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/WhoIsWatchingAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/WhoIsWatchingAdapter.kt
@@ -55,7 +55,6 @@ class WhoIsWatchingAdapter(
             editCallBack = editCallBack,
         )
 
-
     override fun onBindViewHolder(holder: WhoIsWatchingHolder, position: Int) =
         holder.bind(currentList.getOrNull(position))
 
@@ -70,10 +69,15 @@ class WhoIsWatchingAdapter(
         fun bind(card: DataStoreHelper.Account?) {
             when (binding) {
                 is WhoIsWatchingAccountBinding -> binding.apply {
-                    if(card == null) return@apply
+                    if (card == null) return@apply
                     outline.isVisible = card.keyIndex == DataStoreHelper.selectedKeyIndex
                     profileText.text = card.name
                     profileImageBackground.setImage(card.image)
+
+                    // Handle the lock indicator
+                    val isLocked = card.lockPin != null
+                    lockIcon.isVisible = isLocked
+
                     root.setOnClickListener {
                         selectCallBack(card)
                     }
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
@@ -242,28 +242,22 @@ object DataStoreHelper {
         }
 
         binding.applyBtt.setOnClickListener {
-            val currentAccounts = accounts.toMutableList()
-
-            val overrideIndex =
-                currentAccounts.indexOfFirst { it.keyIndex == currentEditAccount.keyIndex }
-
-            // If an account is found that has the same keyIndex, override that one; if not, append it
-            if (overrideIndex != -1) {
-                currentAccounts[overrideIndex] = currentEditAccount
+            if (currentEditAccount.lockPin != null) {
+                // Ask for the current PIN
+                showPinInputDialog(context, currentEditAccount.lockPin, false) { pin ->
+                    if (pin == null || pin != currentEditAccount.lockPin) {
+                        binding.lockProfileCheckbox.isChecked = true
+                    } else {
+                        // PIN is correct, proceed to update the account
+                        performAccountUpdate(currentEditAccount)
+                        dialog.dismissSafe()
+                    }
+                }
             } else {
-                currentAccounts.add(currentEditAccount)
+                // No lock PIN set, proceed to update the account
+                performAccountUpdate(currentEditAccount)
+                dialog.dismissSafe()
             }
-
-            // Save the current homepage for new accounts
-            val currentHomePage = DataStoreHelper.currentHomePage
-
-            // Set the new default account as well as add the key for the new account
-            setAccount(currentEditAccount, false)
-            DataStoreHelper.currentHomePage = currentHomePage
-
-            accounts = currentAccounts.toTypedArray()
-
-            dialog.dismissSafe()
         }
 
         // Handle setting or changing the PIN
@@ -295,6 +289,23 @@ object DataStoreHelper {
         }
 
         canSetPin = true
+    }
+
+    private fun performAccountUpdate(account: Account) {
+        val currentAccounts = accounts.toMutableList()
+
+        val overrideIndex = currentAccounts.indexOfFirst { it.keyIndex == account.keyIndex }
+
+        if (overrideIndex != -1) {
+            currentAccounts[overrideIndex] = account
+        } else {
+            currentAccounts.add(account)
+        }
+
+        val currentHomePage = DataStoreHelper.currentHomePage
+        setAccount(account, false)
+        DataStoreHelper.currentHomePage = currentHomePage
+        accounts = currentAccounts.toTypedArray()
     }
 
     private fun getDefaultAccount(context: Context): Account {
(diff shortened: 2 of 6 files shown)


fire-light42:
While the UI and code is very good it has a major flaw that needs fixing before it gets merged. It can be bypassed all together by holding down the locked profile (to edit it) then pressing apply. Apply should only change the account to that account if the pin was also entered

fire-light42 (changes requested):
Very good code

Luna712:
> While the UI and code is very good it has a major flaw that needs fixing before it gets merged. It can be bypassed all together by holding down the locked profile (to edit it) then pressing apply. Apply should only change the account to that account if the pin was also entered

@fire-light42 

Sorry If I messed up the code in newer one, a bit messier code, but much UX improvement, automatically logs in when the pin is correct and displays good error message (dynamically) when changing the pin, but it made a bit messier, which I apologize for if it is a problem I can fix.

Luna712:
The bypass issue is a good catch and I am working on it.

Luna712:
My new version of this is somewhat inspired by how it works on Netflix, but in my opinion better UI.

fire-light42:
The change still suffers from the same bug along with another one that might have been present in the last one. But if you press lock then dismiss it, it will still have the checked as locked but not be locked

fire-light42:
Change UI all you like, however this will not get merged before all bugs are fixed

Luna712:
@fire-light42 I *think* I fixed those bugs

fire-light42:
Still not fixed the dismiss issue. 
1. Press add a new profile
2. Press lock
3. Dismiss the lock by pressing outside the dialog
4. It will still show locked, but will be be locked

fire-light42:
Also, a lock on default should not be possible as that can also be bypassed by creating an account and then deleting it

Luna712:
> Still not fixed the dismiss issue.
> 
> 1. Press add a new profile
> 2. Press lock
> 3. Dismiss the lock by pressing outside the dialog
> 4. It will still show locked, but will be be locked

I can not reproduce this, clicking outside the dialog dismissing it still keeps ir locked for me.

fire-light42:
image
image
image
image

Luna712:
@fire-light42 this does not appear to happen to me still, tried many combinations and it didn't happen.

Luna712:
oh wait I see

Luna712:
I thought you were talking about if locked to unset lock you could bypass needing the current lock

Luna712:
I'll finish this in about two hours

fire-light42:
Also, from a ux perspective autoclosing if correct with 4 chars is very bad as if I click more than how many items there are it will close it when I try to edit the profile. Then is the bug that I cant remove a pin from a profile

Luna712:
> Also, from a ux perspective autoclosing if correct with 4 chars is very bad as if I click more than how many items there are it will close it when I try to edit the profile. Then is the bug that I cant remove a pin from a profile

It only closes automatically if you enter the correct pin

Luna712:
I'm not sure under what circumstances you couldn't remove a pin, it doesn't auto close when creating a new pin either.

Luna712:
Also it is forcibly limited to 4 characters in layout with `android:maxLength="4"` so there could never be more.

Luna712:
@fire-light42 the bug should be fixed as well as no pins on default account

fire-light42:
I am still unable to remove a pin from an account.

and while your hiding of lock on default account is fine, it becomes impossible to actually edit the default account if it is already locked. This might be an illegal state right now, but I would still like this fixed if we have to change this in the future

fire-light42:
Same with the auto unlocking when entering the correct password. I much rather have a button as stated before it is easy to click one time too many on a phone

fire-light42:
image
image
image
image
image
As you can see, I am unable to remove a lock from an account

fire-light42:
https://github.com/recloudstream/cloudstream/assets/147925818/d7e08a2c-5b94-4485-b411-2012b310725b

And this is the UX issue I am talking about. you click too much and then you get kicked out without changing anything

fire-light42:
Auto doing something is very bad UX wise as it does something you dont want in case of slightly bad input

Luna712:
> Auto doing something is very bad UX wise as it does something you dont want in case of slightly bad input. This can be fine for TV input as keys are real button presses, but on a screen it is easy to double tap or missclick

Take Netflix for example which does this exact behavior, they have whole teams that decide on the best UX, so I did something similar as I felt it would be good behavior if something like they do it, even on mobile. I can make it only do so on TV, but I still don't personally feel it should. It makes it easier to only have to type a PIN and not have to press other keys or something to finalize logging in. It is only 4 characters, if you double tap or misclick you just do it again, it doesn't close if the PIN is wrong, it just clears and you can type again, it is a password field so you can't see the whole thing anyway, and going back to try one digit different can be just as bad and it's easier to type 4 characters again.

fire-light42:
> > Auto doing something is very bad UX wise as it does something you dont want in case of slightly bad input. This can be fine for TV input as keys are real button presses, but on a screen it is easy to double tap or missclick
> 
> Take Netflix for example which does this exact behavior, they have whole teams that decide on the best UX, so I did something similar as I felt it would be good behavior if something like they do it, even on mobile. I can make it only do so on TV, but I still don't personally feel it should. It makes it easier to only have to type a PIN and not have to press other keys or something to finalize logging in. It is only 4 characters, if you double tap or misclick you just do it again, it doesn't close if the PIN is wrong, it just clears and you can type again, it is a password field so you can't see the whole thing anyway, and going back to try one digit different can be just as bad and it's easier to type 4 characters again.

As I said, it is fine on tv, but NOT on phone, and this is when we edit the account, not when we login. And no, this happend to me several times by accident when testing it so it is VERY bad UX. If you want to solve this without a button then make the account dialog not dismiss when clicking outside

Blatzar:
The issue here is that the keyboard gets lowered when done which moves the UI unexpectedly, in addition to the fact that the dialog is dismissable. 

Auto open is nice, that's how android does it, but it comes with both an indication of how many characters are entered and a sizable animation to react that the phone is opened.

Either go all in and add a unlock animation of ~250ms to prevent accidental touches or make the other dialog non cancelable (setCancelable(false)), at least for ~250ms after unlock. I think that is agreeable.

Luna712:
Fixed some issues but one issue still remains when setting pin, clicking the done button. I'll fix it soon.

Blatzar:
Nice! setCanceledOnTouchOutside is even better than setCancelable :pray:

Luna712:
Ok, I *think* everything is fixed now. Thanks again to both @Blatzar and @fire-light42 for the review and suggestions on this as well. I *hope* that's all, if not let me know and I'll change again.

Luna712:
@fire-light42 @Blatzar 

Screenshot_20231029_161844_CloudStream Debug

I've also been working on a new account select page that allows pin protection at startup also, still more work, like auto focusing on last account, etc... and I'm not even sure if it's really something you want, but it is an option if so, it was mostly just a PoC for me to see how it would really work, and potential UX/UI behind it.

Luna712:
It seems to load a lot faster that way also, well loads account selection faster than normal without, then loads main page faster after selecting account.

It might just be my connection though.

Blatzar:
Wonderful! That is a very requested feature :pray: 
I think a lot of TV users would be eternally grateful for this feature

Blatzar:
And your PoC already looks very good :drooling_face:

Luna712:
> Wonderful! That is a very requested feature 🙏 I think a lot of TV users would be eternally grateful for this

@Blatzar 

Glad to hear its something that could be used then! I opened my draft PR in #736

fire-light42:
Will have a look at both tomorrow and hopefully merge both. it looks :fire:

Luna712:
> Will have a look at both tomorrow and hopefully merge both. it looks 🔥

@fire-light42 

Thanks a ton! Though, if you're going to merge both, forget this one, and just merge the other one please, it includes this patch, if it will take longer to get that one merged, this can be merged first and I'll rebase the other, its just more work.

Luna712:
Marking as draft, you can unmark and merge this if you want to do this first, but a major reason I did the account selection PR was to support profile locks from startup so to ensure that had to do it there also, so this isn't needed if the other one is merged. Only if it will take longer/needs a lot more work.

PR #737, opened 2023-10-31
Use nicer grid layout for account select on TV

Luna712:


app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountSelectActivity.kt
```diff
@@ -51,9 +47,13 @@ class AccountSelectActivity : AppCompatActivity() {
         }
         recyclerView.adapter = adapter
 
-        recyclerView.layoutManager = if (isTvSettings()) {
-            LinearLayoutManager(this)
-        } else GridLayoutManager(this, 2)
+        if (isTvSettings()) {
+            val spanSize = if (accounts.count() <= 6) {
+                accounts.count()
+            } else 6
```
Luna712, line 53: I'm debating whether this should be 4 and the accounts UI scaled up a bit or 6

Luna712: I think 6 works fine, I might try something else later but this should be ok to merge for now

Luna712:
Doing in another PR combined with any other improvements

Luna712:
@fire-light42 sorry to bother you but do you think you can merge this one if/when you have time? It doesn't look right on TV as it stands, I forgot to take into account it was larger screens also, not just DPAD to consider, so this improves it for that.

PR #739, opened 2023-10-31
feat(ui): add ok button for pin dialog and use isvisible getter

IndusAryan:


Luna712:
Also I am working on a new PR to improve a few things, one thing it does do is add a done button when creating a new PIN, when you aren't creating a new PIN, everything is dynamic, errors if the PIN is wrong, and auto complete when it is right so this isn't needed for always. Other parts of the PR include account creation at startup and a few other things but I can push that separately if you want me to do the done button faster also?

Luna712:
> imo what's the need of account creation at start-up bcos profile icon is just on home?

It's kinda hard to explain exactly what I am doing, but it basically changes the way accounts creation work to make it a lot easier on TV etc... it is fairly hard to get to account creation on TV so doing from startup makes it a lot easier for UX/UI

Luna712:
Can this be closed please? Later tonight or tomorrow I will hopefully have a new PR that redoes the entire account system UI/UX. If it is decided against, I will do a sperate PR that does this a better way and more convenient for TV and Phone users alike.

Issue #740, opened 2023-10-31
2 Setting options  


ITron-Legends:
### Setting 1 
- Able to turn off the account picker when starting the app (let the user also pick the one it should go to automatically)

### Setting 2 
- Option to stop playing the movie automatically when clicked if it's in the continue watching tab.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

fire-light42:
1. Yes, that would be very nice @Luna712 
2. Why? Continue watching is quite literally made for that. If you want to pull up the info page instead then you can long hold

Luna712:
@fire-light42 
I can work on the first one soo of you'd like? I was thinking something very similar, in fact, I was thinking of some option to turn it off via something clicked on startup (like a "Don't show this again" option, with the option to re-enable in settings later), but maybe not worth it, maybe just a setting would be much better.

imDusht:
I think atleast account picker shouldn't show up if last opened profile doesn't have a lock (& open to this profile)

Luna712:
> I think atleast account picker shouldn't show up if last opened profile doesn't have a lock (& open to this profile)

@imDusht 

My multipurpose PR (#741) adds this as a preference. If you set it to skip account selection at startup, it will use the last used account. If the last used account has a PIN it just requires you enter the PIN at startup, or alternatively (so the app does not get locked out completely if you forget it or something) there is a button in the PIN dialog on startup to "Use Default Account" instead.

PR #741, opened 2023-10-31
Massive changes to account flow

Luna712:
The main purpose of this to to put all account stuff in one place and make it easier for TV users to manage accounts. If you don't want to go about it this way we can do some other way. This has numerous UX and focus improvements as well. Accounts are live updated and this fixes a few preexisting bugs.

I have additional plans for follow up PRs to have even more improvements such as making use of DrawerLayout for some things to make it easier to use. Inputs in dialogs are not always that convenient and I often accidentally dismiss them when I am not done, so may plan for DrawerLayout should improve things, but that will come in a follow-up PR eventually, unless you do not want it to.

Also adds a preference to turn off account selection at startup (refs #740)
If the preference is enabled, and the last used account has a PIN lock, it still asks for PIN at startup, however there is a button in PIN dialog in this case to "Use Default Account" so that of they don't know the PIN they don't get completely locked out.


Luna712:
Screenshot_20231031_173416_CloudStream Debug
This i
Screenshot_20231031_173424_CloudStream Debug
s the current version with more still to come, and a new version of the bottom dialog selector within MainActivity (so you don't always have to exit to this to do anything) It'll just take more work to finilize but I hope it works well enough, if not I'll do a few seperate PRs for the individual important components of this.

Luna712:
> big pencil looking ugly on cards, honestly

@IndusAryan 

Indeed, I am working on design improvement

Luna712:
I might just split this into multiple PRs honestly, it is becoming hard to track things I want to do and that might be easier.

Luna712:
#737 and #742 are the most important parts of this so those can be merged first then I will be rebase this and continue working on perfecting it.

Blatzar:
How about blurring the backgrounds when doing pencil stuff :thinking:

Luna712:
> How about blurring the backgrounds when doing pencil stuff 🤔

@Blatzar 

This is exactly what I was thinking, I did dim the backgrounds, but it didn't seem to be enough, so blurring them seems like a good idea.

Luna712:
@Blatzar @IndusAryan this is pretty much done, other than maybe a few bits of cleanup:
Screenshot_20231103_152228_CloudStream Debug
Screenshot_20231103_152023_CloudStream Debug
Screenshot_20231103_152422_CloudStream Debug

Things to note:
* Long press to manage accounts still works if not on TV (it doesn't work great on TV 95% of the time (before, so it isn't enabled on TV, the way to edit accounts is made more clear with a button available also))
* If you click "Manage Accounts" button in linear account management, it takes you to AccountSelectActivity to manage all accounts, and it runs it already in edit mode. Upon saving an edit to an account it will automatically exit back to MainActivity, resumed in the exact same state you left it, using the edited account. If you click the cancel editing button, and you came from MainActivity management, it will immediately return to MainActivity, rather than making you select an account again.

Those were the biggest things to mention, there was a lot of focus changes for TV with this, as well as some wording improvements, IE actually use something saying "Edit account" instead of "Create account" if actually editing an existing account, this also unifies account logic in one place rather than two adapters (partially identical) it allows us to unify it and manage the account logic a bit easier as well. Overall, in my personal opinion it is much improvement, especially for TV, if you don't want to do it like this, I'll find another way to satisfy any concerns. Thanks!

fire-light42 (changes requested):
1. Optional: I really don't like that you don't do this in viewmodels, it is not a necessary change, but I think you should take a look at it at least. Looking at al this UI code just feels dirty
2. Major: If you delete an account it wont register in the startup menu. 
3. Minor: The big edit symbol is not aligned with the text, make it aligned or in the top right
4. Minor: The edit symbol on all accounts moves the lock to make space, that just looks bad
5. Minor: The manage account button should be a fully horizontally filling BlackButton/WhiteButton, having it in the primary color goes against the design of the rest of the app
6. Minor: The edit button should have a background ripple effect

fire-light42:
By having a viewmodel you could share all the logic between the inapp edit account and the preapp edit account and make it really really easy to modify the UI without causing bugs

Luna712:
@fire-light42 thanks for the suggestions, view models confuse me, but I am working on implementing a view model for live updates at least so you don't have to reopen the bottom dialog if you are editing multiple, and it updates in a live context, but the rest of the UI code, I don't get view models as much. I can try to improve my knowledge of them to improve on that if absolutely needed though.

fire-light42:
> @fire-light42 thanks for the suggestions, view models confuse me, but I am working on implementing a view model for live updates at least so you don't have to reopen the bottom dialog if you are editing multiple, and it updates in a live context, but the rest of the UI code, I don't get view models as much. I can try to improve my knowledge of them to improve on that if absolutely needed though.

View models work by representing the state of the app in the viewmodel without any references to the view or context. Then observing the change in the fragment and updating the views. When a view is clicked it calls the viewmodel to tell that we have change something. By doing this the view is only a representation of the state and not the state itself, so if we wanted to change the UI or remove some part we can just change it without affecting the backed logic at all. 

This makes the app fault resistant to app restarts and makes it hard to get into an invalid state.


For example a simple case would be a counter button

Viewmodel:
```
val count : MutableLiveData<Int> = MutableLiveData(0)

fun add() = count.postValue(count.value + 1)
```
Fragment:
```
observe(viewmodel.count) { count ->
    button.text = "I have been clicked $count times"
}
button.setOnClickListener {
    viewmodel.add()
}
```
no matter what you do here the button will always display the correct data, and if you wanted you could change this to a textview or some listview without any problems

Luna712:
> Minor: The manage account button should be a fully horizontally filling BlackButton/WhiteButton, having it in the primary color goes against the design of the rest of the app

It did not look as nice when I tried that, if focused it looks nice, so I did a material TextButton, so it looks nicer, can change it if you want though, I can probably make it look nice the other way also if needed.

Luna712:
@fire-light42 I think I fixed some of your layout issues, though not 100% sure it is still great or not with that.

Luna712:
@fire-light42 I moved quite a but of logic to ViewModel now, that is shared between the the two places.

Luna712:
> 1. Optional: I really don't like that you don't do this in viewmodels, it is not a necessary change, but I think you should take a look at it at least. Looking at al this UI code just feels dirty
> 2. Major: If you delete an account it wont register in the startup menu.
> 3. Minor: The big edit symbol is not aligned with the text, make it aligned or in the top right
> 4. Minor: The edit symbol on all accounts moves the lock to make space, that just looks bad
> 5. Minor: The manage account button should be a fully horizontally filling BlackButton/WhiteButton, having it in the primary color goes against the design of the rest of the app
> 6. Minor: The edit button should have a background ripple effect

@fire-light42

I think this should all (?) be fixed now.

Luna712:
Linear account select now maintains scroll position, so say you make a new account it finally no longer resets the scroll to the beginning, which was always annoying for me, also now maintains scroll when bottom dialog is closed, so reopening it will scroll to current account automatically, it was always annoying it did not before especially on phone where only a few accounts fit before scroll needed. Even more annoying that it was previously still showing the white focused outline on current account but did not scroll to it, it does now. This also fixes a pre-existing bug where library is not refreshed when a *new* account is created, only when an account is edited or manually switched to, that is fixed now as well.

Luna712:
This also now only completely reloads the home page if the API has changed, making switching accounts a lot faster if they use the same API. Bookmarks, and library are still refreshed regardless.

Luna712:
@fire-light42 I think this should be 100% done now, maybe a few things you might find in a review, but I made some more performance and UX improvements as well. This PR is quite large and I sorta wish I I did it in a few smaller ones, but it works fairly well at least.

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    fun setIsEditing(value: Boolean) {
+        _isEditing.postValue(value)
+    }
+
+    fun toggleIsEditing() {
+        _isEditing.postValue(!(_isEditing.value ?: false))
+    }
+
+    fun updateAccounts(context: Context) {
```
fire-light42, line 36: This should never be used, the viewmodel should instead handle all the logic like
```kotlin
import com.lagradost.cloudstream3.AcraApplication.Companion.context

fun deleteAccount(account : Account) {
	removeKeys(account.keyIndex.toString())
	val currentAccounts = DataStoreHelper.accounts.toMutableList()
	currentAccounts.removeIf { it.keyIndex == account.keyIndex }
	DataStoreHelper.accounts = currentAccounts.toTypedArray()
	
	if (account.keyIndex == DataStoreHelper.selectedKeyIndex) {
		setAccount(getDefaultAccount(context))
	}
        _accounts.postValue(getAccounts(context))
        _selectedKeyIndex.postValue(DataStoreHelper.selectedKeyIndex)
}
fun createAccount(account : Account) = handleAccountUpdate(account, context)
fun editAccount(account : Account) = handleAccountUpdate(account, context)
```
not via updateAccounts. Right now showAccountEditDialog is in charge of deleting the account and that makes no sense. The showAccountEditDialog should instead get the viewmodel and call the functions on it

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt
```diff
@@ -335,6 +335,15 @@ class SettingsGeneral : PreferenceFragmentCompat() {
             return@setOnPreferenceChangeListener true
         }
 
+        settingsManager.edit().putBoolean(
+            getString(R.string.skip_startup_account_select_key),
+            getKey(getString(R.string.skip_startup_account_select_key), false) ?: false
+        ).apply()
+        getPref(R.string.skip_startup_account_select_key)?.setOnPreferenceChangeListener { _, newValue ->
```
fire-light42, line 342: This entire section can be removed as the it should instead be used like 
```kotlin
val skipStartup = settingsManager.getBoolean(ctx.getString(R.string.skip_startup_account_select_key),false) || accounts.count() <= 1
```

fire-light42: jsdelivr_proxy_key is also wrong for the same reason and should be changed too, but you dont need to do that

fire-light42: Also this might want to be moved into account section as it is about accounts

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
```diff
@@ -0,0 +1,359 @@
...
+
+        if (forStartup) {
+            val currentAccount = DataStoreHelper.accounts.firstOrNull {
+                it.keyIndex == DataStoreHelper.selectedKeyIndex
+            }
+
+            builder.setTitle(context.getString(R.string.enter_pin_with_name, currentAccount?.name))
+            builder.setOnDismissListener {
+                if (!isPinValid) {
+                    exitProcess(0)
```
fire-light42, line 223: Why is this here???

app/src/main/res/layout/account_select_linear.xml
```diff
@@ -0,0 +1,44 @@
...
+        android:orientation="horizontal"
+        android:descendantFocusability="afterDescendants"
+        android:layout_marginTop="10dp"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+        tools:itemCount="4"
+        tools:listitem="@layout/account_list_item" />
+
+    <com.google.android.material.button.MaterialButton
```
fire-light42, line 33: The style you are using is only used for a specific kind of dialog, this is not such a dialog and such should instead of the standard blackbutton/whitebutton. Also this for some reason has a layout_marginStart=10dp that causes misaligned + it does not take up the entire bottom making the hitbox a bit fucked impo
```xml
    <com.google.android.material.button.MaterialButton
        android:id="@+id/manage_accounts_button"
        android:layout_gravity="center"
        android:layout_marginBottom="10dp"
        android:text="@string/manage_accounts"
        android:textSize="16sp"
        app:icon="@drawable/ic_baseline_edit_24"
        style="@style/BlackButton" />
```

app/src/main/AndroidManifest.xml
```diff
@@ -161,9 +161,7 @@
 
         <activity
             android:name=".ui.account.AccountSelectActivity"
-            android:configChanges="orientation|screenSize|smallestScreenSize|screenLayout|keyboard|keyboardHidden|navigation"
```
fire-light42, line 164: any reason why you would want to delete all of these? It will recreate the entire activity when you rotate the screen, so if you want to input your pin and rotate the screen it will disappear for example. I see no reason why you would want to make it act like this

fire-light42 (changes requested):
The app works really well, and I did not find any sort of bugs, however I have some questions about some design decisions like not placing accounts skip in the account settings and recreating the entire activity when rotation the screen.

That being said, I think this is my final review before merge as you should be able to fix all these in 10min as it is quite trivial changes. After that we can merge this

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
Luna712: So that if it is wrong, it exits, but only if you close it manually IE cancel button or back button, otherwise it isn't close. It just makes the back button or cancel close instantly rather than showing an empty activity string if trying to enter pin from startup immediately. I can change it to a different method if you want?

app/src/main/AndroidManifest.xml
Luna712: It was unintentional to remove the whole line. I meant to just remove a couple meaningless ones but selected the line apparently. My bad.

app/src/main/res/layout/account_select_linear.xml
Luna712: My concern with this is it looks fine on Phone, but with a background and the button being full length, it doesn't look good on TV at all in my opinion. I'll make the change though.

Luna712:
It should all be fixed now.

app/src/main/res/layout/account_select_linear.xml
fire-light42: @Luna712 If you feel like it does not look good on tv you can make a separate UI for that, however phone is still the target device for this project

fire-light42:
This is ready to merge, however I found a small bug that might need like 1 line to fix. If you go to homescreen->accounts->manage account->edit any account to change eg picture->apply ->home then it will not have updated in the home account switcher?

app/src/main/res/layout/account_select_linear.xml
Luna712: I understand. Thanks!

Luna712:
> This is ready to merge, however I found a small bug that might need like 1 line to fix. If you go to homescreen->accounts->manage account->edit any account to change eg picture->apply ->home then it will not have updated in the home account switcher?

I just remembered, this is the real reason updateAccounts() was created in the view model, then it was used for deletions improperly but not why it existed. I'll see if I can fix it again another way, thanks!

Luna712:
@fire-light42 it should be fixed although maybe not the best way I could not seem to find another way to get it fixed.

Luna712:
Anyway, this can be merged whenever now if you're okay with my method for fixing that last bug, if the commit did not apply, it can be fixed in a follow-up as GitHub seems to be messing up and I can't commit it again...

fire-light42 (approved):

fire-light42:
🔥 pr as always, this is a really nice addition to the ux and functionality of accounts. Glad to finally merge it!🥳 🥳 🥳

Luna712:
Awesome thanks for merging @fire-light42.

The last thing I want to do with accounts is what we discussed in my original PR adding AccountSelectActivity and PINs to make PIN accounts require them again when re-entering the account (unless skip account startup is enabled probably) if that still seems like a good idea to you?

PR #742, opened 2023-11-01
Add done button for when creating new PINs

Luna712:


app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountDialog.kt
```diff
@@ -48,11 +52,32 @@ object AccountDialog {
...
+                if (binding.pinEditTextError.isVisible) {
+                    // If the done button is pressed and there is an error,
+                    // ask again, and mention the error that caused this.
+                    showPinInputDialog(
+                        context = binding.root.context,
+                        currentPin = null,
+                        editAccount = true,
+                        errorText = binding.pinEditTextError.text.toString(),
+                        callback = callback
+                    )
```
Luna712, line 72: Just noting this basically is meant to give the same results as the keyboard done listener down below:
```kt
binding.pinEditText.setOnEditorActionListener { _, actionId, _ ->
            if (actionId == EditorInfo.IME_ACTION_DONE && isPinValid) {
                val enteredPin = binding.pinEditText.text.toString()
                callback.invoke(enteredPin)
                dialog.dismissSafe()
            }
            true
        }
        ```

But, we can't stop this from dismissing the dialog it seems, so we just ask again, until there is no error, or the user cancels it, and we make sure to show the error again so the user knows what went wrong. It just makes it easier so you don't have to toggle the checkbox again, or wonder went went wrong if you accidently press "Done" with no error displaying afterword's. It's not the best practice for this to call itself but it works, and just makes it require less manual work, making it easier on TV for example where this is kinda hard to navigate sometimes.

File: app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountDialog.kt [hunk 2]
```diff
@@ -48,11 +52,32 @@ object AccountDialog {
...
+                if (binding.pinEditTextError.isVisible) {
                     callback.invoke(null)
                 }
             }
-            .create()
+
+        if (isNewPin) {
+            if (errorText != null) binding.pinEditTextError.visibleWithText(errorText)
+            builder.setPositiveButton(R.string.setup_done) { _, _ ->
+                if (binding.pinEditTextError.isVisible) {
```
fire-light42, line 63: While this works, it is really bad to confuse state with the user interface. In a perfect world the UI state should be entirely separate from the business logic of the app, and I do know that this app does not entirely follow this at all times, this is too much. I think I failed to notice this before, but going forward you should use variables or view models to do this

fire-light42 (changes requested):
Try to separate the UI and logic as it makes it a lot easier to manage

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountDialog.kt
fire-light42: Dont do any sort of control flow operations based on the look of UI

Luna712: I think this is fixed

fire-light42 (approved):

PR #748, opened 2023-11-02
Fix two strings

Luna712:
It looks like they were both fixed in my original commits in #727 and #730 but fixing some merge conflict in #727 accidentally removed that fix


C10udburst:
pushed to weblate to avoid conflicts (#681)

Luna712:
> pushed to weblate to avoid conflicts (#681)

Thanks!

PR #750, opened 2023-11-02
Fix some deprecations and other warnings

Luna712:
* Replace deprecated onBackPressed()
* Fixes some AndroidManifest warnings
  * Use android:dataExtractionRules for Android 12+
  * Add tools:ignore for ScopedStorage and QueryAllPackagesPermission
  * Make VideoDownloadRestartReceiver not exported (fixes "
Exported receiver does not require permission") - if this was incorrect I can revert this part
  * Bump tools:targetApi
* Enable android:enableOnBackInvokedCallback for predictive back gestures on Android 13 (if enabled in developer options) and Android 14 (also fixes "OnBackInvokedCallback is not enabled for the application." which is commonly seen in logcat)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -1576,6 +1576,23 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
 //                showToast(this, currentFocus.toString(), Toast.LENGTH_LONG)
 //            }
 //        }
+
+        onBackPressedDispatcher.addCallback(
+            this,
+            object : OnBackPressedCallback(true) {
+                override fun handleOnBackPressed() {
+                    window?.navigationBarColor = colorFromAttribute(R.attr.primaryGrayBackground)
+                    updateLocale()
+
+                    // If we don't disable we end up in a loop with default behavior calling
+                    // this callback as well, so we disable it, run default behavior,
+                    // then re-enable this callback so it can be used for next back press.
+                    isEnabled = false
+                    onBackPressedDispatcher.onBackPressed()
+                    isEnabled = true
+                }
+            }
+        )
     }
 
     private var backPressedCallback: OnBackPressedCallback? = null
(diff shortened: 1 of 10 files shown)


fire-light42:
@Blatzar

Luna712:
This still has one issue left to fix, I'll do shortly.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultTrailerPlayer.kt
```diff
@@ -172,12 +175,22 @@ open class ResultTrailerPlayer : ResultFragmentPhone(), IOnBackPressed {
...
+        if (backPressedCallback == null) {
+            backPressedCallback = object : OnBackPressedCallback(true) {
+                override fun handleOnBackPressed() {
+                    updateFullscreen(false)
+                }
+            }
         }
+
+        backPressedCallback?.isEnabled = true
+        requireActivity().onBackPressedDispatcher.addCallback(requireActivity(), backPressedCallback!!)
```
Blatzar, line 190: Please refrain from ever using requireActivity() or !!

Even if it should logically never be an issue android has a habit of being extremely weird and if it can crash it will crash. Please use activity?. and other such null safe behavior.

Blatzar (changes requested):

Blatzar:
Very nice to see this, maybe we can implement some cool predictive back gestures after this :pray:

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultTrailerPlayer.kt
Luna712: Sorry about that, fixed and thanks!

Blatzar (approved):
Testing looks good :+1:

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -650,34 +650,6 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
         builder.show().setDefaultFocus()
     }
 
-    private fun backPressed() {
-        this.window?.navigationBarColor =
```
fire-light42, line 654: This was removed causing the bug

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
Luna712: Oh thanks! Fixed.

fire-light42 (approved):
🥳  Very good job, nice pr as always. Reply with 👍 or react and I will merge it

PR #760, opened 2023-11-10
Fix faulty automatic subtitle selection

self-similarity:
This fixes subtitles being automatically selected to the first available subtitle (often Arabic) by media3.
This also likely makes it more reliable since it seems like a lot of code is made to account for the unreliable automatic selection.

0 was used instead of a constant since there is no constant for 0.


Luna712:
Thank you so much for fixing this. This has been an annoying little bugger for me for quite some time, especially on TV

PR #763, opened 2023-11-11
fix: upgrade gradle and kgp upgrade

IndusAryan:
let the build workflow fail after merging #726 to get latest log, then merge this one , i think this work as the make jar classes workflow will now run after the prerelease and build push.


C10udburst:
i tested it locally and it didn't work

PR #770, opened 2023-11-15
Fixed flag and name for Levantine Arabic and Najdi Arabic

CranberrySoup:


KingLucius:
I think it's more appropriate to make the official standard arabic translation to use the same flag of "ars" not "ae" as usual of all arabic translations.
All arab countries have their local accents but we used to use the flag of Saudi Arabia as the flag for standard arabic translations.

PR #781, opened 2023-11-20
bump guava, json(tests), kgp, desugaring

IndusAryan:


app/build.gradle.kts
```diff
@@ -178,7 +178,7 @@ dependencies {
 
     // For KSP -> Official Annotation Processors are Not Yet Supported for KSP
     ksp("dev.zacsweers.autoservice:auto-service-ksp:1.1.0")
-    implementation("com.google.guava:guava:32.1.2-android")
+    implementation("com.google.guava:guava:32.1.3-jre")
```
fire-light42, line 181: why switch from android to jre?

IndusAryan: was automatic, i Wii check in few minutes

PR #782, opened 2023-11-21
feat: divide crashlogs in two different forms for tvs and other devices 

IndusAryan:


IndusAryan:
left the old form url for some reasons:

many crashes from different forks come there.

many crashes are already fixed but still recieved because people are using old versions.

rename old folder as legacy crashes.

so pls make 2 new forms and different folder and tell me in here or private and i will push a commit including the new urls.

fire-light42:
@LagradOst

app/src/main/java/com/lagradost/cloudstream3/AcraApplication.kt
```diff
@@ -37,11 +38,17 @@ import kotlin.concurrent.thread
...
+        println("Sending Crash Report")
+
+        val url = if (SettingsFragment.isTrueTvSettings()) {
+            "FORM_URL_FOR_TVs"
+        } else {
+            "FORM_URL_FOR_PHONES"
+        }
+
         val data = mapOf(
             "entry.1993829403" to errorContent.toJSON()
```
C10udburst, line 53: have you checked that this key does not change for different google forms?

IndusAryan: how can i do that ?
probably no if created by same gmail account 🤔

IndusAryan: > have you checked that this key does not differ between forms?

hmm, found out that it won't be an issue because the form and its input data structure is same its just we have seperated their endpoint reciever url

LagradOst:
The crash report system has to be reworked entirely as google forms is a duck tape solution. If you poke at this I would suggest that you instead implement it as a real crash reporting system

IndusAryan:
> The crash report system has to be reworked entirely as google forms is a duck tape solution. If you poke at this I would suggest that you instead implement it as a real crash reporting system

by reworking what??please tell steps of what needs to be ditched and cleaned up etc.

LagradOst:
> > The crash report system has to be reworked entirely as google forms is a duck tape solution. If you poke at this I would suggest that you instead implement it as a real crash reporting system
> 
> how and by reworking what??please tell steps of what needs to be ditched and cleaned up etc.

Instead of reporting to google forms (that just wont load after ~500k crash reports, and is super slow) we should instead use a real crash reporting site. I think saikou uses a more sophisticated crash reporting site, if you want to take a look at that.

IndusAryan:
>  I think saikou uses a more sophisticated crash reporting site
latest saikou fork Dantotsu uses google firebase crashlytics 👽

IndusAryan:
@Blatzar @fire-light42 what do you think of this
https://github.com/bugsnag/bugsnag-android

fire-light42:
Closing as stale.

PR #786, opened 2023-11-28
chome vs chrome (added missing 'r')

rndx539:
the title says it all


IndusAryan:
it's a key , have you searched for its usage

IndusAryan:
it's a preference key in form of string, that can be used as a shared preference instead of the variable, right click the chrome string in .kt file and find if it has any usages,

PR #795, opened 2023-12-04
extractor: added vidplay and fix few extractors

Sofie99:


b4byhuey:
Can you fix Dailymotion? or can you conver from aniyomi [here](https://github.com/aniyomiorg/aniyomi-extensions/raw/master/lib/dailymotion-extractor/src/main/java/eu/kanade/tachiyomi/lib/dailymotionextractor/DailymotionExtractor.kt)

app/src/main/java/com/lagradost/cloudstream3/extractors/Vidplay.kt
```diff
@@ -0,0 +1,92 @@
...
+                this.name,
+                it.file ?: return@map,
+                "$mainUrl/"
+            ).forEach(callback)
+        }
+
+    }
+
+    private suspend fun getKeys(): List<String> {
+        return app.get("https://raw.githubusercontent.com/Claudemirovsky/worstsource-keys/keys/keys.json")
```
fire-light42, line 47: this should be cached in a local variable

fire-light42 (changes requested):
We dont want to ddos github for every call to geturl, save it to a local variable even if app.get may be cached it does not hurt to explicitly cache it

fire-light42 (approved):

fire-light42:
Understandable, I can merge this now if you want

PR #798, opened 2023-12-09
feat (library tabs): save and restore tab selection

Funny-Pen-7005:
Issue:
savedInstanceState is not restoring the tab selection when you open any movie and return to library section. Also, scroll position is not restored (I was not able to add changes to remember scroll position. Someone, pls give it a try).

Steps to reproduce:
1. (Optional if you have liked movies in library) Go to any movie and like it from top bar
2. Go to library, scroll and select favorite from bottom tab selector
3. Click on any movie
4. Go back, it will take you to watching instead of favorite

Affects:
This fix will affect tab functionality of the library section.

Fix:
libraryviewmodel will save the tab position and will be used to restore the tab position.


fire-light42:
Very nice catch, I fixed the other issue with sorting by using this. 🥳 

Good first pr!

PR #805, opened 2023-12-10
feat (resume): based on resume episode, restore details in UI (season, range and dubbed status)

ghost:
fix: changes basically shows the correct season, range and dub status based on the resume details

pls do review.

when resumeMeta will be ExtractorUri?
```
is ExtractorUri -> {
                    DataStoreHelper.setLastWatched(
```
**app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt**

**Movie**
image
**Anime**
image
**TV**
image

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
@@ -1022,6 +1022,7 @@ class GeneratorPlayer : FullScreenPlayer() {
                         resumeMeta.id,
                         resumeMeta.episode,
                         resumeMeta.season,
+                        resumeMeta.dubStatus,
                         isFromDownload = false
                     )
                 }
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragment.kt
@@ -50,6 +50,7 @@ data class ResultEpisode(
     val videoWatchState: VideoWatchState,
     /** Sum of all previous season episode counts + episode */
     val totalEpisodeIndex: Int? = null,
+    val dubStatus: DubStatus? = null
 )
 
 fun ResultEpisode.getRealPosition(): Long {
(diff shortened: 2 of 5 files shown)


Blatzar:
Currently the last used season and dubstatus is already restored, if this should go through that old code needs to be considered too.

@fire-light42

fire-light42:
Season, episode range as well as dubstatus are all stored and are restored to go back to what you selected last time NOT what you watched last time. This pr is not wrong just a different approach, but cs3 already works as intended and I appreciate the PR, but right now there is no issue with cs3.

As for the ExtractUri, it does not contain data because it is a downloaded file that does not contain the data. This extractorUri can come from a random file the user wants to play with a file explorer.

fire-light42:
> What I have done is to show the season, episode range and dubbed status based on the episode you are currently watching. Not based on previous selection.

Yes, I know but the intended experience is to restore the last selected, not watched. your solution is just a different solution.

fire-light42:
you can see the code here where the preferred landing episode, season and dubstatus is selected https://github.com/recloudstream/cloudstream/blob/3c152e04d1a8fa60cfcca8dd1d81f0ded5d1c54b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt#L2624

coxju:
> > I think this is how it should be. As a user, I would rather see what I am actually watching now and what comes next. I think in Netflix and prime video, they show it this way. It has been while I last checked, LOL.
> 
> Netflix does this, however it comes down to user preference. One solution is not correct and other incorrect.

if you feel that way pls close pr. i cannot as i don't have access anymore.

fire-light42:
@Blatzar thoughts on merging this one or keeping the current behavior?

coxju:
If you guys decide to keep it. Please review. As there are changes in many files. Also I am java web dev. I am not very familiar with android and kotlin. So any inputs will be great!

@Blatzar @fire-light42

fire-light42:
> If you guys decide to keep it. Please review. As there are changes in many files. Also I am java web dev. I am not very familiar with android and kotlin. So any inputs will be great!
The code is fine 👍 
However we are just interested in making the UX the best, and we are on the fence on this one

coxju:
no issues. please close the pr.

coxju:
I will try to place a button in the page instead of MR changes, which will update page based on resume details.

fire-light42:
> I will try to place to a button in the page instead of MR changes, which will update page based on resume details.

wdym

coxju:
> I will try to place a button in the page, which will update page based on resume details instead of page load.

like a reset button

PR #806, opened 2023-12-10
refactor (library search): sort and query current page only and delay search when typing

ghost:
Fix: In library, sort and query current page only and delay search for 500ms when user is typing.

Note: I have used snippets from codeium AI to add delay. So the approach maybe incorrect. Pls review.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/library/LibraryFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/library/LibraryFragment.kt
@@ -148,6 +148,11 @@ class LibraryFragment : Fragment() {
         activity?.theme?.resolveAttribute(android.R.attr.textColor, searchExitIconColor, true)
         searchExitIcon?.setColorFilter(searchExitIconColor.data)
 
+        val searchCallback = Runnable {
+            val newText = binding?.mainSearch?.query?.toString() ?: return@Runnable
+            libraryViewModel.sort(ListSorting.Query, newText)
+        }
+
         binding?.mainSearch?.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
             override fun onQueryTextSubmit(query: String?): Boolean {
                 libraryViewModel.sort(ListSorting.Query, query)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/library/LibraryFragment.kt
```diff
@@ -147,9 +147,15 @@ class LibraryFragment : Fragment() {
         activity?.theme?.resolveAttribute(android.R.attr.textColor, searchExitIconColor, true)
         searchExitIcon?.setColorFilter(searchExitIconColor.data)
 
+        // Declare a Handler and a Runnable
+        val inputDelayHandler = Handler(Looper.getMainLooper())
```
Blatzar, line 151: All views are effectively handlers, so you can do stuff like binding?.mainSearch?.postDelayed()

fire-light42:
What is the reason for this pr? Is this because of UX concerns or lagging? I saw that you only update the page you are on, why?

fire-light42:
Then you need to be aware that updating only one page will give it invalid data when swiping back to another page. Clicking via setOnClickListener is not the only way to navigate between tabs.

fire-light42:
image
What happen 😢

coxju:
move repo to my other username. 
tried swipe. it is not working. checking it.

coxju:
> Then you need to be aware that updating only one page will give it invalid data when swiping back to another page. Clicking via setOnClickListener is not the only way to navigate between tabs.

reverted most of it as it was causing issue when swipe. kept postdelayed changes. when swiping current tab position was not updating, so added back changes to remember tab position.

coxju:
@fire-light42 updated upon your comment. pls chk

app/src/main/java/com/lagradost/cloudstream3/ui/library/LibraryFragment.kt
```diff
@@ -163,7 +164,16 @@ class LibraryFragment : Fragment() {
                     return true
                 }
 
-                libraryViewModel.sort(ListSorting.Query, newText)
+                val callback = Runnable {
+                    libraryViewModel.sort(ListSorting.Query, newText)
+                }
+
+                binding?.mainSearch?.removeCallbacks(callback)
```
fire-light42, line 171: While this seams like it should work it does not in fact do because callback is a local variable and not scoped properly it removes nothing making this pr broken. 

If you want to fix this then put the callback outside the scope of setOnQueryTextListener and do something like 

```kotlin
val searchCallback = Runnable {
    val newText = binding?.mainSearch?.query?.toString() ?: return@Runnable
    libraryViewModel.sort(ListSorting.Query, newText)
}
```

fire-light42 (changes requested):
The code looks good however it contains a small bug

app/src/main/java/com/lagradost/cloudstream3/ui/library/LibraryFragment.kt
fire-light42: Right now it queries for every character, as removeCallbacks does not remove the callback

fire-light42:
Very nice 🥳

PR #808, opened 2023-12-10
add prevent option in CloudflareKiller

Sofie99:


b4byhuey:
Please look [here](https://github.com/aniyomiorg/aniyomi-extensions/blob/master/lib/chillx-extractor/src/main/java/eu/kanade/tachiyomi/lib/chillxextractor/ChillxExtractor.kt), aniyomi get Chilx key from the body

app/src/main/java/com/lagradost/cloudstream3/network/CloudflareKiller.kt
```diff
@@ -48,19 +50,26 @@ class CloudflareKiller : Interceptor {
 
     override fun intercept(chain: Interceptor.Chain): Response = runBlocking {
         val request = chain.request()
+        val response = chain.proceed(request)
```
fire-light42, line 53: Suppose cookies are not null then this line always tries to do a request without cookies and only after that it shoots another request with cookies at line 67. This wastes a request.

fire-light42 (changes requested):

fire-light42 (approved):

fire-light42:
Screenshot_20231216_182028_CloudStream_Beta
This might need to be reverted

PR #810, opened 2023-12-11
Fixing Source List Displaying Empty Items

yueehaoo:


app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerGeneratorViewModel.kt
```diff
@@ -184,16 +184,20 @@ class PlayerGeneratorViewModel : ViewModel() {
             _loadingLinks.postValue(Resource.Loading())
             val loadingState = safeApiCall {
                 generator?.generateLinks(type = type, clearCache = clearCache, callback = {
-                    currentLinks.add(it)
-                    // Clone to prevent ConcurrentModificationException
-                    normalSafeApiCall {
-                        // Extra normalSafeApiCall since .toSet() iterates.
-                        _currentLinks.postValue(currentLinks.toSet())
+                    if (it.first?.url?.isNotEmpty() == true) {
```
Blatzar, line 187: This does not account for ExtractorUri, in that case it.first would be null

fire-light42:
This should instead be in RepoLinkGenerator

fire-light42:
🙏 Nice, this should fix all the empty links

PR #812, opened 2023-12-12
refactor (home) : show account icon when no provider selected

coxju:
Is quick search applicable? should it also show when no provider selected?

Not added for TV. Is it required?

image

image


IndusAryan:
quick search visibility can be overrided but won't function with none provider, why it is needed if none

coxju:
> quick search visibility can be overrided but won't function with none provider, why it is needed if none

ok sounds good

fire-light42:
🥳 Very good pr!

PR #813, opened 2023-12-13
 feat (player): optional rotate button in player and setting to enable auto rotate based on video orientation

coxju:
- icon added
- Strings might be incorrect
- code needs review

setting page:
image

with rotate enabled:
image


fire-light42:
This is a really good addition, and the code looks nice. However it fucks up a bit with the current code located at https://github.com/recloudstream/cloudstream/blob/3c152e04d1a8fa60cfcca8dd1d81f0ded5d1c54b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt#L289 that keeps the player locked in LANDSCAPE when clicking the lock symbol. 

So if you want this merged you can keep this behavior however you need to fix the updateOrientation function to not cause any bugs. By pressing lock you lock the current orientation completely otherwise you do sensor landscape/portrait.

coxju:
@fire-light42 can you share some desc on how updateorientarion and lockorientation is working? when are they triggered?

coxju:
where to get the icons?

fire-light42:
> @fire-light42 can you share some desc on how updateorientarion and lockorientation is working? when are they triggered?

Right now when the user enters the player they will be turned into sensor landscape and when pressing the lock button it will lock to the exact orientation that the player is currently in. updateOrientation is called when entering the player along with when you toggle the lock button. 

So in your case you need to add a bool like IsVerticalOrientation that is set when first loading a video by overriding playerDimensionsLoaded, then calling updateOrientation. updateOrientation should lock the player in the exact orientation the user is currenty in if isLocked otherwise it should use the sensor version of IsVerticalOrientation. Then the button just changes `IsVerticalOrientation = !IsVerticalOrientation; updateOrientation();`

If you want icons then look at https://fonts.google.com/icons

image
https://developer.android.com/guide/topics/manifest/activity-element.html#screen

coxju:
@fire-light42 thank you for explaining in detail. 

I am trying to use playerDimensionsLoaded but is not being triggered when I override in fullscreenplayer.kt

coxju:
@fire-light42  changes are done. pls review.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -286,31 +289,78 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
-            //    if (rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_270) ActivityInfo.SCREEN_ORIENTATION_PORTRAIT else ActivityInfo.SCREEN_ORIENTATION_REVERSE_PORTRAIT
+            Configuration.ORIENTATION_LANDSCAPE ->
+                orientation =
+                    if (rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_90)
+                        ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE
+                    else
+                        ActivityInfo.SCREEN_ORIENTATION_REVERSE_LANDSCAPE
+
+            Configuration.ORIENTATION_SQUARE, Configuration.ORIENTATION_UNDEFINED ->
+                dynamicOrientation(activity)
```
fire-light42, line 339: This will cause a bug as orientation will still be zero and 0 = SCREEN_ORIENTATION_LANDSCAPE, so this function will be called then it will call `activity.requestedOrientation = SCREEN_ORIENTATION_LANDSCAPE` right under this

fire-light42: you have the same problem in the other functons

fire-light42 (commented):
Looks good, however I have not tested this yet so you might get some more feedback soon

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
coxju: This is how it was -> `Configuration.ORIENTATION_SQUARE, Configuration.ORIENTATION_UNDEFINED, Configuration.ORIENTATION_PORTRAIT -> ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE`
Is this doing anything? there is no assignment.

Now in function i have added -> `activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE`

So what should be the change? should I keep it how it was? how to test this Configuration.ORIENTATION_SQUARE, Configuration.ORIENTATION_UNDEFINED?

fire-light42: What I mean is that if you read the code 
image
Then it will set `activity.requestedOrientation = orientation` and `orientation = 0` when `currentOrientation = Configuration.ORIENTATION_SQUARE, Configuration.ORIENTATION_UNDEFINED` so dynamicOrientation wont do anything as it will be overridden 2 lines later

fire-light42: In this case I think dynamicOrientation should return what orientation it should be and then set `orientation = dynamicOrientation(activity)` inside the when statments

coxju: ohk I missed that

coxju: @fire-light42 thanks. Pls chk now.

coxju:
@fire-light42 fixed the issue youve shared. pls review and let me know there is any other issue.

Mostafazarzora:
What


في الاثنين، ١٨ ديسمبر ٢٠٢٣, ٨:٥٥ ص coxju ***@***.***> كتب:

Issue #530, opened 2023-08-02
audio keeps playing after exiting the video, and that of the previous episode on clicking next ep


arxhiebro:
### Steps to reproduce

(on tv) 
1. play an episode
2. Exit or click on next episode
3. Wait for the error to show up followed by the audio of the episode which was playing

### Expected behavior

The previous video as well as auido should stop playing 

### Actual behavior

Audio keeps playing

### Cloudstream version and commit hash

4.1.1

### Android version

Android 10

### Logcat

```java
Unable to get logcat to phone through tv
```

### Other details

https://discord.com/channels/737724143126052974/1136191575152668692

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [X] If related to a provider, I have checked the site and it works, but not the app.
- [X] I will fill out all of the requested information in this form.

Luna712:
@fire-light42 @LagradOst @Blatzar sorry for direct ping here, but any idea what causes this? This is one of the most annoying issues on TV that I have ever seen, about 90% of the time this happens, unless instead of clicking the next EP button, you fast-forward to the very end of the video, that seems to be the only way this is fixed. If you exit the video, and gave autoplay next EP on it continually plays the previous episode audio with no way of fixing without force stopping the app. I have tried to look into this but can not find anything on it. It happens on FireTV, as well as Google TV, and very often. I can not get Logcat from TV either.

Again sorry for the direct ping, but this is the most annoying issue that now instead of playing on TV directly I mostly just cast from my phone to TV which isn't as good. I just hope we can find some way to fix it...

P.S. A couple times I actually had to unplug my Fire TV to fix this as it was just qucker than going back into the app, but sometimes it still plays when you exit the app to go back to home screen, and even pressing the power button to the TV didn't actually stop it, which is more of an OS issue but just goes to show how deadlocked you can get when audio doesn't stop playing. It can get frustrating. A side issue is on Google TV sometimes it gets stuck in some PiP window without being able to re-enter the app, which can be just as annoying. PiP on Google TV doesn't seem the most efficient, and probably doesn't have to much use if it doesn't work much anyway. But that doesn't happen as often and isn't as annoying as the audio issue.

arxhiebro:
just adding some more insights
[...]

PR #822, opened 2023-12-17
CS3IPlayer: stop player when releasing

Luna712:
This *might* help #530, but not 100% sure and couldn't test at the moment. I am unsure if it will have additional side effects.


fire-light42 (approved):
Not tested, but I cant see how this will fuck it up so I will merge

PR #826, opened 2023-12-19
feat(security): add biometric fingerprint sensor / face unlock authentication 

IndusAryan:
tested and working well here, test on your side, unlike old fingerprint manager we dont need to play with any credentials or storing crytpograhic data with keys, with new jetpack api its seamless and simple. 
i have kept the auth async with main activity otherwise it will slow app open.


fire-light42:
What is the intended problem this pr solves? Sure it is a cool feature, but very niche

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,141 @@
...
+
+                override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
+                    super.onAuthenticationError(errorCode, errString)
+                    Toast.makeText(context?.applicationContext, "Authentication error: $errString", Toast.LENGTH_SHORT).show()
+                    activity.finish()
+                }
+
+                override fun onAuthenticationSucceeded(result: BiometricPrompt.AuthenticationResult) {
+                    super.onAuthenticationSucceeded(result)
+                    Log.d("Cs3Auth", "Biometric succeeded.")
```
fire-light42, line 39: move "Cs3Auth" to a variable called TAG

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    fun initializeBiometrics(activity: MainActivity) {
+        val executor = ContextCompat.getMainExecutor(activity)
+        biometricManager = BiometricManager.from(activity)
+
+        biometricPrompt = BiometricPrompt(activity, executor, object : BiometricPrompt.AuthenticationCallback() {
+
+                override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
+                    super.onAuthenticationError(errorCode, errString)
+                    Toast.makeText(context?.applicationContext, "Authentication error: $errString", Toast.LENGTH_SHORT).show()
```
fire-light42, line 33: use stringres for messages, and use the showToast function instead

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,141 @@
...
+
+                override fun onAuthenticationFailed() {
+                    super.onAuthenticationFailed()
+                    Toast.makeText(context?.applicationContext, "Authentication failed", Toast.LENGTH_SHORT).show()
+                    activity.finish()
+                }
+            })
+
+        promptInfo = BiometricPrompt.PromptInfo.Builder()
+            .setTitle("CloudStream")
```
fire-light42, line 50: stringres

fire-light42 (changes requested):
You need to use the apps showToast along with string resources for translation and UI consistency.

I have not tested this pr yet, so more specific UX feedback will come when I do

fire-light42:
after testing this, I can comfirm that it works however I still would like some cover of the screen when doing the auth as rn you can still see all the bookmarks and recently watched. also makes no sense ro not put this on acc screen instead

IndusAryan:
> after testing this, I can comfirm that it works however I still would like some cover of the screen when doing the auth as rn you can still see all the bookmarks and recently watched. also makes no sense ro not put this on acc screen instead

if account select activity is chosen as first then they already have the Pin feature for seperate accounts, and if not then the tapping user profile makes way to main activity, so it ask the bio prompt there.
feature is good because app lock is not available on all devcies as well as it's small to be a bloat

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: done this one

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: ye

IndusAryan:
so, migrated the biom auth to account select  and it parses activity as app compat, and main activity code will still be present to auth when no account are there, because that code when shifted to ASA , it fails to finish() main because it needs a this@param everywhere

IndusAryan:
i am not able to hide auth preference in non mobiles, as well as hide background while prompting. any hint or way awaiting

fire-light42:
> i am not able to hide auth preference in non mobiles, as well as hide background while prompting. any hint or way awaiting

wdym

fire-light42 (approved):
Code looks good, but testing + hide content when auth is needed

IndusAryan:
> > i am not able to hide auth preference in non mobiles, as well as hide background while prompting. any hint or way awaiting
> 
> wdym

anyone can turn on this preference in tv, may cause crashes, apps have this feature, like the preference can be greyed out so that it cant be enabled but somehoow its not working.

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1152,6 +1156,16 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
...
+        /** Biometric stuff for users without accounts **/
+        val authEnabled = settingsManager.getBoolean(getString(R.string.biometric_enabled_key), false)
+        val noAccounts = settingsManager.getBoolean(getString(R.string.skip_startup_account_select_key), false) || accounts.count() <= 1
+
+        if (isTruePhone() && authEnabled && noAccounts ) {
+            BiometricAuthenticator.initializeBiometrics(this@MainActivity)
+            BiometricAuthenticator.checkBiometricAvailability()
+            BiometricAuthenticator.biometricPrompt.authenticate(promptInfo)
+        }
+
```
Luna712, line 1168: Nice patch, but this can still be kept to be done in AccountSelectActivity, before we navigate to the main activity if we are skipping account selection. This would just prevent duplicate code. IE see how PIN is asked even if we are skipping account selection and a PIN exists on the last used account. A little difference but some type of premise for the most part if I am understanding this patch correctly.

IndusAryan: yes, it xan prompted on account selects navigation to main and works but when auth fails or dismissed there, it doesn't finishes the app, totally against authentication, as it requires oncreate of main activity to finish() it.

IndusAryan:
@Blatzar @fire-light42 i have re written this , working good on my tests,  do review. now it supports face unlocking too.

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,177 @@
...
+                    showToast("$errString", LENGTH_SHORT)
+                    Log.i(TAG, "$errorCode")
+                    activity.finish()
+                    failedAttempts++
+                    if (failedAttempts >= MAX_FAILED_ATTEMPTS) {
+                        failedAttempts = 0
+                        activity.finish()
+                    } else {
+                        failedAttempts = 0
+                        MainActivity().finish()
```
Blatzar, line 60: What is this?
You are creating a mainactivity object and finishing it?

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,177 @@
...
+
+                override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
+                    super.onAuthenticationError(errorCode, errString)
+                    showToast("$errString", LENGTH_SHORT)
+                    Log.i(TAG, "$errorCode")
+                    activity.finish()
+                    failedAttempts++
+                    if (failedAttempts >= MAX_FAILED_ATTEMPTS) {
+                        failedAttempts = 0
+                        activity.finish()
```
Blatzar, line 57: You called activity.finish() above though?

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                }
+            })
+    }
+
+    private fun initBiometricPrompt(
+        activity: Activity,
+        title: Int,
+        setDeviceCred: Boolean,
```
Blatzar, line 84: Please add a comment or something detailing what this is, since it is not obvious based on the name nor usage

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+
+        return result
+    }
+
+    // only needed for Android 9 and below
+    fun deviceHasPasswordPinLock(context: Context?): Boolean {
+        val keyMgr =
+            context?.getSystemService(AppCompatActivity.KEYGUARD_SERVICE) as KeyguardManager
```
Blatzar, line 150: This cast throws an exception if context is null, either handle that with as? or dont allow nullable argument

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,177 @@
...
+            promptInfo = BiometricPrompt.PromptInfo.Builder()
+                .setTitle(activity.getString(title))
+                .setDescription(description)
+                .setDeviceCredentialAllowed(true)
+                .build()
+        }
+    }
+
+    private fun isBiometricHardWareAvailable(): Boolean {
+        var result = false
```
Blatzar, line 117: I'd personally much prefer inlining this instead of using vars, but that is just a nitpick.

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,177 @@
...
+import com.lagradost.cloudstream3.R
+
+object BiometricAuthenticator {
+
+    private const val MAX_FAILED_ATTEMPTS = 3
+    private var failedAttempts = 0
+    const val TAG = "cs3Auth"
+
+    private lateinit var biometricManager: BiometricManager
+    lateinit var biometricPrompt: BiometricPrompt
```
Blatzar, line 27: Please no lateinit vars, especially if they are public, its far too easy to shoot yourself in the foot using them.

Blatzar (commented):
Testing wise everything seems to work well :pray: 

Only thing I'm thinking about is whether biometrics should be a toggle (alongside pins) for each account instead of an app wide login. Do you think that makes sense? I don't share devices so I don't know if that is preferable.

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: fixed

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: ok, its the main builder of the prompt

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: current implementation responses robust truthy and falsy values(may look big but is good)

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: changed to nullable

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: done

app/src/main/java/com/lagradost/cloudstream3/utils/BiometricAuthenticator.kt
IndusAryan: fixed, wrote explicitly so that it finish main activity forcibly on error, but not needed as it automatically finishes in case of unknown error

IndusAryan:
@Blatzar @fire-light42 i have pushed final changes , i think this is ready for your final review now.

PR #834, opened 2023-12-20
refactor: move buildconfig, bump ksp & better trailer scraping

IndusAryan:


app/build.gradle.kts
```diff
@@ -139,6 +139,10 @@ android {
         abortOnError = false
         checkReleaseBuilds = false
     }
+    
+    buildFeatures {
```
fire-light42, line 143: why?

IndusAryan:
does ping works in review comment? @fire-light42 , is this newpipe extractor change ok ?, it has nothing to do with extractor bump, but some trailers werent working for example if we watch oppenheimer trailer on sodastream, it plays a 28 minute promo interview of the team, but now after this it is playing trailer (3 min)

fire-light42 (commented):
Will merge after testing

PR #850, opened 2024-01-03
feat : eye icon will show the current season, dub status and episode range for given show

coxju:
image

changes from - https://github.com/recloudstream/cloudstream/pull/805


fire-light42:
Why? I dont see the use case for this, just feels like an extra click for nothing if I understand it correct.

coxju:
this will keep the current implementation. https://github.com/recloudstream/cloudstream/pull/805 here you mentioned you would like to keep how it was.
icon click will show current one without touching the drop-downs.
if you where checking other season and want to go back to the current one. This will be quick. Give it a go.

coxju:
closing this for now.
I have addressed my concern differently here - https://github.com/recloudstream/cloudstream/pull/851

PR #851, opened 2024-01-03
feat: show the episodic range popup with current selection checked

coxju:
before
image

now
image


fire-light42 (commented):
Looks good, will merge after testing

PR #858, opened 2024-01-07
feat : show favorite button in bottom dialog

coxju:
Screenshot:
after adding to favorite
image

after removing from favorite
image


fire-light42:
Damn, looks sexy. The code looks fine, and will merge this after testing (max 5 days).

LagradOst (changes requested):
After testing this I found that this causes an UI bug in the library if you remove/add a favorite while in library it does not refresh. In toggleFavoriteStatus add the line MainActivity.reloadLibraryEvent(true)

LagradOst:
I will merge this and fix it myself along with another bottom dialog issue

PR #865, opened 2024-01-10
feat : invalidate link cache after 20 mins

coxju:
- additionaly clear cache if there is player errors or no links found


LagradOst:
Why was this closed?

PR #875, opened 2024-01-13
feat : invalidate link cache after 20 mins

coxju:
- additionaly clear cache if there is player errors or no links found


fire-light42:
No this changes the MainAPI fucking over all extentions, all the logic should be in the repolinkgenerator cache

fire-light42 (approved):

PR #876, opened 2024-01-13
feat (play with cloudstream) : open links with matching provider

coxju:


app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadedPlayerActivity.kt
```diff
@@ -36,6 +39,22 @@ class DownloadedPlayerActivity : AppCompatActivity() {
     }
 
     private fun playLink(url: String) {
+        // check if url matches any provider
+        synchronized(APIHolder.apis) {
+            for (api in APIHolder.apis) {
+                if (url.startsWith(api.mainUrl) ||
+                    // this is to match mirror domains - like example.com, example.net
```
C10udburst, line 46: I don't quite understand why you used a fuzzy match here. Most providers with mirror domains already handle mirror domains. This approach might cause a URL not to load that would have done so otherwise, for example if you have two different providers that have similar domains.

fire-light42:
I can merge this, but why add this?

fire-light42:
> > I can merge this, but why add this?
> 
> this will useful just like how it is for extractor links. If you find something to watch directly on the site, then using play with cloudstream it will directly open without needing search and find in app.

But this function is for playing with cloudstream, not open with cloudstream. the function is just to PLAY a videofile, and giving the same url to a mainpage will never work

fire-light42:
> I have tried with sflix it was working. it will not play but it will open the result page. Can we move this to a separate intent handler "open in cloudstream"?

This already exist, however I think it got removed from the android manifest when the extension system was implemented https://github.com/recloudstream/cloudstream/blob/399b28c75b00ea9522a89a4ccd4077acf23256f6/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt#L433

fire-light42:
> @fire-light42 can we add it back? what can be done?
It was removed because it explicitly named what sites that could be opened in cs3, but you can add it back matching all urls

PR #877, opened 2024-01-13
feat (loadExtractor) : match mirror domains of extractor link

coxju:


app/src/main/java/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -592,7 +593,12 @@ suspend fun loadExtractor(
     val currentUrl = unshortenLinkSafe(url)
     val compareUrl = currentUrl.lowercase().replace(schemaStripRegex, "")
     for (extractor in extractorApis) {
-        if (compareUrl.startsWith(extractor.mainUrl.replace(schemaStripRegex, ""))) {
+        if (compareUrl.startsWith(extractor.mainUrl.replace(schemaStripRegex, "")) ||
+            // this is to match mirror domains - like example.com, example.net
+            FuzzySearch.partialRatio(
```
Blatzar, line 598: I think this is better done in a second loop, lest a partial domain gets prioritized over an exact match

Blatzar:
Good thinking!

PR #883, opened 2024-01-15
feat: add function to check if video is live stream (playback only)

IndusAryan:


fire-light42:
Why?

IndusAryan:
lucious will use this to control his iptv extensions
second i will use this to modify the time counter if i want 
by this we can set some custom things also like disabling playback speed in live etc

fire-light42:
the code right now is nonsense, it is a void function that takes context for no reason. add it in the real pr where it is used

PR #888, opened 2024-01-16
feat : run custom js in webviewresolver

coxju:
this will work similar to rhino
https://stackoverflow.com/questions/19788294/how-does-evaluatejavascript-work


fire-light42 (changes requested):
This changes a public facing API, you need to add the old constructor to not fuck over all extentions on stable

fire-light42 (approved):
Lets hope this does not break stable or pre

PR #892, opened 2024-01-17
Screw it, Self host keys

KillerDogeEmpire:
I am tired of the keys breaking, as soon as I notice them break, I can now fix immediately without making another new PR. this will be the last key change PR. Sorry for so many, I just couldn't take them not working so I will host them


KillerDogeEmpire:
I noticed that the key I swapped to before self hosting has a 5 - 1h delay before getting new key, I fixed that so it happens exactly every hour 10 seconds before it turns to the new hour. It seems to work. But if it breaks I will immediately fix when I have time if I'm not in the middle of school

PR #901, opened 2024-01-20
feat : match extractor based on regex

ghost:


C10udburst:
> @fire-light42 @C10udburst @LagradOst please let me know if there is another way for below issue. I have a extractor where iframe domain is different every time. Fuzzy search doesn't match for all subdomains.
> 
> sub1.example.com sub2.example.com subsub1.example.com subsub2.example.com
> 
> Also I think we can reduce number of extractors, if we use pattern instead of using startswith.

Perhaps adding a function such as this to `ExtractorApi` would be a solution:
```kt
fun matches(str: String): Boolean {
	return str.startsWith(mainUrl)
}
```
This way extractors with unconventional domains could override this function with custom code.

fire-light42:
> This way extractors with unconventional domains could override this function with custom code.
This has already been discussed, and we wanted extractors to use regex but never implemented it. a function like matches or a open string regex should be implemented

PR #902, opened 2024-01-20
feat: added xtreamestream extractor

coxju:


app/src/main/java/com/lagradost/cloudstream3/extractors/Xtremestream.kt
```diff
@@ -0,0 +1,39 @@
+package com.lagradost.cloudstream3.extractors
+
+import com.lagradost.cloudstream3.app
+import com.lagradost.cloudstream3.utils.ExtractorApi
+import com.lagradost.cloudstream3.utils.ExtractorLink
+import com.lagradost.cloudstream3.utils.M3u8Helper
+
+open class Xtremestream : ExtractorApi() {
+    override var name = "Xtremestream"
+    override var mainUrl = "xtremestream.co"
```
fire-light42, line 10: https://

fire-light42: does not matter, the mainurl should specify if it is secure or not

PR #915, opened 2024-01-22
feat(ui): add ignore battery optimisation dialog for uniterrupted downloads and notifications

IndusAryan:
so android nowadays is very powerful in dozing apps and killing them in background even when app is using foreground services, google apps are whitelisted already by default, users can manually make cs3 unrestrict but most of the time they dont know they can do this, so now get a dialog to set it to unrestricted with proper info of better downloading and subscription notification.

idk if message can be more better or informatic. when they press ok, the app info of cs3 opens where they can go to app battery usage setting.


app/src/main/AndroidManifest.xml
```diff
@@ -14,7 +14,7 @@
     <uses-permission android:name="com.android.providers.tv.permission.WRITE_EPG_DATA" /> <!-- Used for Android TV watch next -->
     <uses-permission android:name="android.permission.UPDATE_PACKAGES_WITHOUT_USER_ACTION" /> <!-- Used for updates without prompt -->
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" /> <!-- Used for update service -->
-
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
```
fire-light42, line 17: You request REQUEST_IGNORE_BATTERY_OPTIMIZATIONS without using  ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1174,8 +1178,12 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
             null
         }
 
-        changeStatusBarState(isEmulatorSettings())
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !isAppOptimised && !isTrueTvSettings()
```
fire-light42, line 1181: Another annoying popup when installing the app, this is not what we want. This should only show up when actually trying to download stuff.

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1747,4 +1755,20 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
             false
         }
     }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    private fun showBatteryOptimizationDialog() {
```
fire-light42, line 1760: This dialog also pop up every time until you press ok, making it VERY anoying

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1174,8 +1178,12 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
             null
         }
 
-        changeStatusBarState(isEmulatorSettings())
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !isAppOptimised && !isTrueTvSettings()
+            && !isTvSettings() && !isEmulatorSettings()) {
+            showBatteryOptimizationDialog()
```
fire-light42, line 1183: This also has no settings related option to access this after doing it for the first time

app/src/main/AndroidManifest.xml
fire-light42: Look at https://stackoverflow.com/questions/44862176/request-ignore-battery-optimizations-how-to-do-it-right

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -321,6 +323,8 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
          */
         val reloadLibraryEvent = Event<Boolean>()
 
+        /** For battery optimisation **/
+        private var isAppOptimised : Boolean by UserPreferenceDelegate("isAppOptimised", false)
```
fire-light42, line 327: UserPreferenceDelegate is just a wrapper for set/get key, so restoring a backup skips this even if it is not optmized

fire-light42 (changes requested):
This is very good download wise, but VERY bad UX wise. You want as few fullscreen popups as possible. 

1. You use REQUEST_IGNORE_BATTERY_OPTIMIZATIONS wrong
2. No option for retrying
3. No option for ignore
4. UserPreferenceDelegate is only local and cant actually keep track if the app is optimized or not, get the current status before requesting or rerequesting 
5. Only request this ***when*** downloading as it is unnecessary otherwise. 
6. Why only after M?

IndusAryan:
>     4. UserPreferenceDelegate is only local and cant actually keep track if the app is optimized or not, get the current status before requesting or rerequesting

 i used power manager to get status of optimization before prompting, the problem i faced that after setting unrestricted , it takes hours for the status to get updated, that too different, maybe instantly in some devices turning feature into a jackass bug.

>     6. Why only after M?

in below marshmallow, battery savers dont doze apps that are in foreground, the fu*kery started in marshmallow api 6

KingLucius:
Check this if you want some ideas
https://github.com/WaseemSabir/BatteryPermissionHelper

IndusAryan:
> Check this if you want some ideas https://github.com/WaseemSabir/BatteryPermissionHelper

this lib look like a beta and tested for less sample size of vendors

KingLucius:
I think if we should do this, i suggest it as a step of the intial setup.

IndusAryan:
that's already denied by firelight in 5th point of review, the setup itself looks very annoying

fire-light42:
> we have 2 things now, should there be a preference also to set this, if yes then in which settings?
> 
> should the cancel button permanently dismiss the dialog forever?
> 
> @fire-light42 @KingLucius

Sorry for the late reply, but if you want to get this merged it should also be added in settings and if you dismiss it then dont ask again (ever). Then you should also remove requireContext() as that WILL crash android. Also isTruePhone = !isTvSettings() that is why we dont have that function.

IndusAryan:
@fire-light42 i have pushed some final changes and tested on my side good, there is preference as well as it dismisses forever when canceled, but i couldnt understand why requireContext() needs to be removed, because app crashes and sometimes things dont work without it, lacking context in logcat, it preference also it works good and opens info directly.

fire-light42:
> but i couldnt understand why requireContext() needs to be removed

Because sometimes android just decides that there is no context for whatever reason, and that causes the app to crash. So you always want to treat context as a nullable variable that cant be trusted. Both the android docs and examples use requireContext, but in cs3 we want to avoid that as much as possible, as we KNOW that it causes a significant amount of crashes because android just sucks.

IndusAryan:
> Because sometimes android just decides that there is no context for whatever reason, and that causes the app to crash. So you always want to treat context as a nullable variable that cant be trusted. Both the android docs and examples use requireContext, but in cs3 we want to avoid that as much as possible, as we KNOW that it causes a significant amount of crashes because android just sucks.

i have first tried defining ctx as nullable context, if not available it tries obtaining one as mentioned that android causes it, the dialog invocation is under the run catching that will pevent app crash if both ctx are absent, it this approach OK?

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt
```diff
@@ -204,6 +210,21 @@ class SettingsGeneral : PreferenceFragmentCompat() {
             return@setOnPreferenceClickListener true
         }
 
+        // hide preference on tvs and emulators
+        if (!isLayout(PHONE)) {
+            getPref(R.string.battery_optimisation_key)?.isEnabled = false
```
fire-light42, line 215: getPref(R.string.battery_optimisation_key)?.isEnabled = isLayout(PHONE)

app/src/main/java/com/lagradost/cloudstream3/utils/PowerManagerAPI.kt
```diff
@@ -0,0 +1,88 @@
...
+                    .show()
+            }
+        } catch (e: Exception) {
+            Log.e(TAG, "Error showing battery optimization dialog", e)
+        }
+    }
+
+    private fun shouldShowBatteryOptimizationDialog(context: Context): Boolean {
+        val isRestricted = isAppRestricted(context)
+        val isOptimizedShown = PreferenceManager.getDefaultSharedPreferences(context)
```
fire-light42, line 66: isOptimized***Not***Shown

app/src/main/java/com/lagradost/cloudstream3/utils/PowerManagerAPI.kt
```diff
@@ -0,0 +1,88 @@
...
+                        intentOpenAppInfo(it)
+                    }
+                    .setNegativeButton(R.string.cancel) { _, _ ->
+                        settingsManager.edit()
+                            .putBoolean(context.getString(R.string.battery_optimisation_key), false)
+                            .apply()
+                    }
+                    .show()
+            }
+        } catch (e: Exception) {
```
fire-light42, line 59: t : Throwable as Exception does not protect you from errors

app/src/main/java/com/lagradost/cloudstream3/utils/PowerManagerAPI.kt
```diff
@@ -0,0 +1,88 @@
...
+    fun showBatteryOptimizationDialog(context: Context) {
+        val settingsManager = PreferenceManager.getDefaultSharedPreferences(context)
+
+        try {
+            context.let {
+                AlertDialog.Builder(it)
+                    .setTitle(R.string.battery_dialog_title)
+                    .setIcon(R.drawable.ic_battery)
+                    .setMessage(R.string.battery_dialog_message)
+                    .setPositiveButton(R.string.ok) { _, _ ->
```
fire-light42, line 49: not open?

fire-light42 (commented):
Minor changes before merge

fire-light42 (approved):

PR #923, opened 2024-02-03
Add SubScene

CranberrySoup:
I had to change the subtitle API to make it work with Zip files but it is backwards compatible. 

Worked with @aymanbest to implement the API.


CranberrySoup:
deleteOnExit does not seem to work properly as of now.

CranberrySoup:
*"Note that on Android, the application lifecycle does not include VM termination, so calling this method will not ensure that files are deleted. "*
All other instances of deleteOnExit should be reviewed too.

fire-light42:
👍

SittingDuck999:
Splendid job. Where to select the subs from subscene?

Imon24:
superb work.❤️‍🔥
I'm facing a little problem. when trying to select sub of a series that has all subs in one zip it always selecting the 1st sub (ep1). can't select other subs like e2, e3, e4.

PR #925, opened 2024-02-07
chore: add tooltips on results toolbar and rename speed mode

IndusAryan:
i removed the skyuk related rant, as it looks shady and gives away fake intentions for cs3 app, as it mentions dmca and can dig up past.


KillerDogeEmpire:
How dare you rename the speed

PR #919, opened 2024-01-29
refactor: migrate toast to view bindings, fix gravity bug, upgrade gradle & ksp

IndusAryan:

PR #929, opened 2024-02-14
feat: make cloudstream builds & compilation fast, using congifuration cache

IndusAryan:
🦍

also reverted the gradle plugin & kotlin symbol processing upgrade bump from #919 into this PR.


app/build.gradle.kts
```diff
@@ -233,6 +235,21 @@ dependencies {
...
 }
 
+tasks.register<Exec>("retrieveCommitHash") {
+    // task needed because configurationCache does not support parallel git commands
+    group = "build"
+    description = "Retrieves the commit hash."
+
+    doFirst {
+        commandLine("git", "rev-parse", "--short", "HEAD")
+        val commitOutput = layout.projectDirectory.file("commit-hash.txt")
```
Blatzar, line 245: Would it be possible to move this to the build dir?
Having it as a part of the app code will be confusing, especially with future pull requests where this file might give merge conflicts.
Otherwise this looks very good :pray: :pray: :pray:

IndusAryan: ok , ill check, currently i am finishing off biometric authenticator now.

IndusAryan:
sorry i have to close this, i am just unable to complete this pr, now it has even stopped writing the hash to file, it remains 0kb dump file, plus buildDir is very fucky and cannot be used for tasks like this, i have tried using res/raw but still results are same. maybe someone else better than me will definitely finish this.

PR #934, opened 2024-02-15
Long press Extensions Repo to copy URL

KingLucius:
On Mobile only, you can long press extensions repo to copy URL.
I don't know if this gonna be an addition to TV, you can review and tell me if you want to on TV also


IndusAryan:
thanks man, i wanted this since long and was about to add in some time, needed this for shaering to people instead of going to ext repos everytime, not needed in TV imo

KingLucius:
> thanks man, i wanted this since long and was about to add in some time, needed this for shaering to people instead of going to ext repos everytime, not needed in TV imo

Yes ! , thank you for Title copy PR actually which made me remember this hh

Blatzar:
Good PR :+1:

PR #941, opened 2024-02-18
WebViewResolver: added custom timeout

Sofie99:


fire-light42:
Should work, lets hope this does not break stable 👍

fire-light42:
(it will break stable if you use this api before next update)

PR #944, opened 2024-02-20
Fix Resume Button on TV

KingLucius:
Fixes the behavior of buttons visibility on TV in resume case.


fire-light42:
I am unsure if this conflicts with #950

KingLucius:
> I am unsure if this conflicts with #950

No need for it any more if we are going to merge #950

Issue #755, opened 2023-11-09
Play button doesn't currently show the next unwatched episode to play


LinkofHyrule89:
### Steps to reproduce

The play button always seems to just say Episode 1 and does not change to the current unwatched episode when one becomes available.

[Screenshot_20231109-043718](https://github.com/recloudstream/cloudstream/assets/3983279/9eae4acd-8c5e-430f-a5e1-93ad3f1b132e)

### Expected behavior

When an unwatched episode is available it should say that episode instead of episode 1. If you mark if so's watched or watch them manually it should offer the next unwatched episode on the play button.

!

### Actual behavior

Currently if there are multiple watched episodes The play button does not offer the next unwatched episode.

### Cloudstream version and commit hash

4.2.1 11136fe

### Android version

Android 10 (Sony Bravia Google TV)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [X] If related to a provider, I have checked the site and it works, but not the app.
- [X] I will fill out all of the requested information in this form.

Blatzar:
Looks like it is caused by removeLastWatched being called when the last episode is seen. 

Possible to replicate using:

```kt
        val animeResponse = (loadResponse as AnimeLoadResponse)
        animeResponse.episodes = animeResponse.episodes.apply {
            this[DubStatus.Subbed] = this[DubStatus.Subbed]!!.subList(0,2)
        }
```
In postSuccessful in case anyone wants to replicate it.

LinkofHyrule89:
This bug still exist in the current latest nightly build are there any updates on this?

KingLucius:
Can you please check #983 if this fixed incase of actuall watching (still working on mark as watched button)

LinkofHyrule89:
It seems like of you wash and episode and it goes to the next episode and the episode is in the "Continue watching" it ahead of correctly. But if you mark an episode as watched or remove it from the continue watching it shows it as Episode 1 still. There is an edge case where if it's multiple sets of episodes it'll just start with the first episode of the season selected. So it seems like it's still pretty much broken in most cases. When there a be episode that's unwatched that gets added and the previous episode is watched it still shows episode 1 instead of that newest unwatched episode. I am on build 4.3.1 638cc4f.

[...]

PR #950, opened 2024-02-25
New TV UI

KingLucius:


app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1185,6 +1185,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener {
 
                 if(isTrueTvSettings()) {
                     newLocalBinding.root.viewTreeObserver.addOnGlobalFocusChangeListener { _, newFocus ->
+                        if (newFocus?.tag == "tv_no_focus_tag") return@addOnGlobalFocusChangeListener
```
Blatzar, line 1188: Please use 
```kt
context.getString(R.string.tv_no_focus_tag)
```

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -251,8 +251,11 @@ class ResultFragmentTv : Fragment() {
             //episodesShadow.rotationX = 180.0f//if(episodesShadow.isRtl()) 180.0f else 0.0f
             
             val leftListener: View.OnFocusChangeListener =
-                View.OnFocusChangeListener { _, hasFocus ->
+                View.OnFocusChangeListener { view, hasFocus ->
                     if (!hasFocus) return@OnFocusChangeListener
+                    if (view?.tag == "tv_no_focus_tag"){
```
Blatzar, line 256: Same here

app/src/main/res/layout/fragment_home_head_tv.xml
```diff
@@ -157,6 +157,7 @@
                     android:id="@+id/home_preview_hidden_prev_focus"
                     android:layout_width="1dp"
                     android:layout_height="1dp"
+                    android:tag="tv_no_focus_tag"
```
Blatzar, line 160:
```xml
android:tag="@string/tv_no_focus_tag"
```

app/src/main/res/layout/fragment_home_head_tv.xml
```diff
@@ -189,6 +190,7 @@
                     android:id="@+id/home_preview_hidden_next_focus"
                     android:layout_width="1dp"
                     android:layout_height="1dp"
+                    android:tag="tv_no_focus_tag"
```
Blatzar, line 193: Same here

Blatzar (commented):
Code looks good, I haven't tested the UI though.

fire-light42:
Contains 2 bugs atm
1. Episodes disappear (After playing a trailer)
2. Remaining is fucked 
image

KingLucius:
> Contains 2 bugs atm
> 
> 1. Episodes disappear (After playing a trailer)
> 2. Remaining is fucked
>    image

First issue happens after back from Episode playing also

fire-light42 (commented):
This is a fast code review, I need to test this in an emulator so another more in depth review might come later

app/src/main/res/values/strings.xml
```diff
@@ -306,7 +306,14 @@
     <string name="go_forward_30">+30</string>
     <string name="delete_message" formatted="true">This will permanently delete %s\nAre you sure?</string>
     <string name="resume_time_left" formatted="true">%dm\nremaining</string>
+    <string name="resume_remaining" formatted="true">%s\nremaining</string>
     <string name="status_ongoing">Ongoing</string>
+    <string name="days_short">d</string>
```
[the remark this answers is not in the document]
fire-light42: eg https://github.com/recloudstream/cloudstream/pull/878/files#diff-e230dcbac950214bc4ed6fbc02ac00a73bcccb15ccc1d34d57e87eb9ff057e56R1547

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -129,9 +132,17 @@ class ResultFragmentTv : Fragment() {
...
      */
     private fun focusPlayButton() {
-        binding?.resultPlayMovie?.requestFocus()
-        binding?.resultPlaySeries?.requestFocus()
-        binding?.resultResumeSeries?.requestFocus()
+        binding?.resultPlayMovieButton?.requestFocus()
+        binding?.resultPlaySeriesButton?.requestFocus()
+        binding?.resultResumeSeriesButton?.requestFocus()
+    }
+    private fun setFocusRotation(id: Int)
```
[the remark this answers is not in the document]
KingLucius: for the setFocusRotation , it gets the last visible button and sets the nextfocusright to be the visible play/resume button in movie case, so the focus will be rotational and avoid going to some weired unvisible item.

LinkofHyrule89:
Is this issue fixed in this updated UI? https://github.com/recloudstream/cloudstream/issues/755

KingLucius:
> Is this issue fixed in this updated UI? #755

[here](https://github.com/recloudstream/cloudstream/issues/755#issuecomment-1806157292) lag mentioned it's intentional behavior so it's not a bug

LinkofHyrule89:
> > Is this issue fixed in this updated UI? #755
> 
> [here](https://github.com/recloudstream/cloudstream/issues/755#issuecomment-1806157292) lag mentioned it's intentional behavior so it's not a bug

That doesn't make sense because it never moves to the next unwatched episode so it's clearly a bug. This needs to be reevaluated. It used to work correctly previously before the last big update. I think most people would agree it should work similar to Netflix or Crunchyroll or any other service and properly show the next unwatched episode.

KingLucius:
> > > Is this issue fixed in this updated UI? #755
> > 
> > [here](https://github.com/recloudstream/cloudstream/issues/755#issuecomment-1806157292) lag mentioned it's intentional behavior so it's not a bug
> 
> That doesn't make sense because it never moves to the next unwatched episode so it's clearly a bug. This needs to be reevaluated. It used to work correctly previously before the last big update. I think most people would agree it should work similar to Netflix or Crunchyroll or any other service and properly show the next unwatched episode.

Ok let me study the details first, and working on a solution if there is one.

fire-light42:
> That doesn't make sense because it never moves to the next unwatched episode so it's clearly a bug. This needs to be reevaluated. It used to work correctly previously before the last big update. I think most people would agree it should work similar to Netflix or Crunchyroll or any other service and properly show the next unwatched episode.

It is the intended behavior as we dont know the metadata of the next episode before it exists

fire-light42:
I tested this pr now, and can say that it works fine. However there is a few small issues impo, the button wraps the text really bad. I am not really sure how to fix this when we have multiple languages

image

So I have some suggestions for discussion (I prefer 1 and 4)

1. layout_width="wrap_content"
image

2. You can do ellipsize="marquee", however it might fuck up longer text if the end is required
image

3. android:singleLine="true" + requiresFadingEdge="horizontal", will fuck up longer text, but looks nice as it is centered
image

4. singleLine="true" + scrollHorizontally="true" + marqueeRepeatLimit="marquee_forever" + ellipsize="marquee" +resultBookmarkText.isSelected = result_bookmark_Button.hasfocus
this will make the text scroll when you have the focus over the set button
studio64_a02up0qICT

But for all these I think you should use a specific style for the text

The second nitpick is that the None icon is filled, when it should be an outlined bookmark/+ like favorite. (I know this is how it also looks on mobile atm, but that is because it uses colorprimary instead)
image
image


Other than that, I think this pr is ready to be merged. @Blatzar thoughts?

KingLucius:
Suggestion #4 is great in my opinion, i will test all these suggestions and feedback
For the None icon, great detail, i will work on it

LinkofHyrule89:
> > > > Is this issue fixed in this updated UI? #755
> > > 
> > > 
> > > [here](https://github.com/recloudstream/cloudstream/issues/755#issuecomment-1806157292) lag mentioned it's intentional behavior so it's not a bug
> > 
> > 
> > That doesn't make sense because it never moves to the next unwatched episode so it's clearly a bug. This needs to be reevaluated. It used to work correctly previously before the last big update. I think most people would agree it should work similar to Netflix or Crunchyroll or any other service and properly show the next unwatched episode.
> 
> Ok let me study the details first, and working on a solution if there is one.

Ok I appreciate it! I don't expect it to guess the next episode if there is no meta data but, if there are watched episodes it should ignore them and skip to the next unwatched episode or just show a rewatch button for the last episode that exists for the show. Maybe run a check for unwatched episodes when you load into the listing itself?

Blatzar:
> Maybe run a check for unwatched episodes when you load into the listing itself?

No, we want to limit the spam, and it would introduce a whole lot of other design issues. Such as making the load time longer, deciding on how many to load. Not loading all would be inconsistent and loading all would take a lot of time and cause a lot of unnecessary spam. 

You rewatch idea has merit though. The problem with that is that it may 'trick' users into rewatching unless done right. The ideal solution would obviously be to only remove continue watching if the series wasn't ongoing, but as far as I know we lack that metadata.

The current design stems from these considerations. Netflix and Crunchyroll can design as they do because they have more metadata and they can design their system to get the new episodes for everything without spamming network requests.

LinkofHyrule89:
I guess I'm trying to understand how come it worked previously but after the last redesign was rolled out it doesn't work it properly anymore. Some sort of logic must have been removed or changed so it ignores what episodes are watched and just suggest the first episode instead. Obviously it's more of a pet peeve than anything else since I just have to open the episode list and scroll to the bottom but when a show has a ton of episodes it's a little harder. 

Well I appreciate your hard work on this other revision hopefully you can figure out the one particular issue I'm having In the long run.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -327,6 +327,78 @@ class ResultFragmentTv : Fragment() {
                 }
             }
 
+            resultPlayMovieButton.setOnFocusChangeListener { _, hasFocus ->
```
fire-light42, line 330: bro, wth am I seeing??
make this a variable and do smth like let focuschange = { v,f -> v.isSelected = f }

for v in listof(resultPlayMovieText,resultPlaySeriesButton...) { v.setOnFocusChangeListener(focuschange) }

fire-light42: because rn it is wet code

app/src/main/res/layout/fragment_result_tv.xml
```diff
@@ -289,7 +289,10 @@ https://developer.android.com/design/ui/tv/samples/jet-fit
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:gravity="top|center_horizontal"
-                        android:minLines="2"
+                        android:singleLine="true"
```
fire-light42, line 292: Also make this a style

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
KingLucius: totally right, resolved

app/src/main/res/layout/fragment_result_tv.xml
KingLucius: Done

Luna712:
:+1: :rocket:  I really love this new UI. Good job! :tada:

app/src/main/res/values/strings.xml
```diff
@@ -745,4 +746,8 @@
     <string name="rotate_video_desc">Display a toggle button for screen orientation</string>
     <string name="auto_rotate_video_desc">Enable automatic switching of screen orientation based on video orientation</string>
     <string name="auto_rotate_video">Auto rotate</string>
+    <string name="favorite">Favorite</string>
+    <string name="movie">Movie</string>
```
fire-light42, line 750: movies_singular?

app/src/main/res/values/strings.xml
```diff
@@ -745,4 +746,8 @@
...
+    <string name="unfavorite">Unfavorite</string>
+    <string name="result_search">Search</string>
```
fire-light42, line 752: search_hint or title_search?

fire-light42 (approved):

KingLucius:
> :+1: :rocket:  I really love this new UI. Good job! :tada:

Thank you so much 🎉

Luna712 (approved):

KingLucius:
for the 
image
I prefer it to be just a text like rest of meta data.
what do you think?

Luna712:
> for the image I prefer it to be just a text like rest of meta data. what do you think?

I personally like it better as-is, it is easier to read and stands out more and is more inline with how a lot of other applications/sites display content ratings.

KingLucius:
I have added the "Coming soon" support & using "Episode" string in case that Season is null as the E1 is small and there is a space already

fire-light42 (approved):

Luna712 (approved):

fire-light42:
Do you want this merged?

KingLucius:
> Do you want this merged?

I good with this so far, you can merge it
Thank you

PR #965, opened 2024-03-06
feat(ui): make cloudstream superfast boi via baseline & startup profile generation

IndusAryan:
[Overview](https://developer.android.com/topic/performance/baselineprofiles/overview)

slowly we can write certain navigation pattern and scroll tests that will increase performance further by observing pattern and reducing jitter, as of now the startup profile needs to some 8 hrs to generate startup profile in phones after app update to take some effect.


fire-light42:
Java moment, why did they choose a compiled but not compiled language to make an OS around💀

app/build.gradle.kts
```diff
@@ -231,6 +232,11 @@ dependencies {
     implementation("androidx.work:work-runtime:2.9.0")
     implementation("androidx.work:work-runtime-ktx:2.9.0")
     implementation("com.github.Blatzar:NiceHttp:0.4.11") // HTTP Lib
+
+    // Baseline Profile and Automation
```
fire-light42, line 236: testImplementation?

fire-light42:
Just asking, but does this collect any use data or change anything on the userside? I do not really understand what it does.

app/build.gradle.kts
IndusAryan: no they are not test implementation, the profiler generates the profile and and the module is stored with apk, in assets/dexopt. if we rename to test then probably it will fail even if succeds the baseline profile wont be included in app

IndusAryan:
> Just asking, but does this collect any use data or change anything on the userside? I do not really understand what it does.

it doesnt collect any sensitive data, currently our startup profile will increase app open speed by precompiling classes ahead of time .
but baseline profile is capable of increasing scroll speed , reducing jitter and lag when navigating if we write test for it and include those test result in profile data with every stable release in coming months.

as of now in nutshell it is for fast startup and app open only.

PR #973, opened 2024-03-08
Simkl: show ReWatching in library

Luna712:
Previously rewatching didn't even function at all, setting "rewatching" just forced it to watching, and rewatching itself did absolutely nothing.

Though, I'm unsure if this would cause issues or why it wasn't in there tl begin with. If it'll cause issues, feel free to close...

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SimklApi.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SimklApi.kt
@@ -226,7 +226,7 @@ class SimklApi(index: Int) : AccountManager(index), SyncAPI {
             Paused(2, R.string.type_on_hold, "hold"),
             Dropped(3, R.string.type_dropped, "dropped"),
             Planning(4, R.string.type_plan_to_watch, "plantowatch"),
-            ReWatching(5, R.string.type_re_watching, "watching"),
+            ReWatching(5, R.string.type_re_watching, "rewatching"),
             None(-1, R.string.none, null);
 
             companion object {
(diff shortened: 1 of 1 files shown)


Blatzar:
This suggests that there is no "rewatching" https://simkl.docs.apiary.io/#reference/sync/last-activities/get-last-activity#Last%20Activities

Is this tested?

Luna712:
> This suggests that there is no "rewatching" https://simkl.docs.apiary.io/#reference/sync/last-activities/get-last-activity#Last%20Activities
> 
> Is this tested?

It should probably be removed then I guess? i meant to make it clear in the PR description that this was untested, and thought I did. No it is not tested, I didn't have time to at the moment. Apologies for that.

Blatzar:
> It should probably be removed then I guess? i meant to make it clear in the PR description that this was untested, and thought I did. No it is not tested, I didn't have time to at the moment. Apologies for that.

I just wanted to be fully sure, lest there be any unnecessary confusion, no need to apologize, its good that you look at the code and come with new ideas :pray:

I think it exists currently in the code to get parity with cloudstreams global `SyncWatchType`, but at the very least I can agree that the code requires a comment since it is not intuitive why it is like that.

Luna712:
Hmmm, I can't seem to test Simkl changes, as I can't login to Simkl from CS Debug: `{"error":"client_failed","message":"Your client_id is wrong. Try another one."}`

Sid954 (approved):

Blatzar:
The client id and such are secrets not included in the source code, to prevent abuse
You need to register your own app at simkl and then put your id and secret in your system environment or your `local.properties` file like this:

```
simkl.id=...
simkl.secret=...
```

See https://github.com/recloudstream/cloudstream/blob/master/app/build.gradle.kts#L77-L86

Luna712:
Oh, thanks!

PR #983, opened 2024-03-11
New TV UI bug fixes

KingLucius:
- Episode down to description closes Episodes Holder.
- Rework main buttons visibility, should fix the visibility bugs.
- Fix Episode space bug (Episode1)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
@@ -247,7 +247,6 @@ class ResultFragmentTv : Fragment() {
             )
         // ===== ===== =====
         var comingSoon = false
-        var resumeMode = false
 
         binding?.apply {
             //episodesShadow.rotationX = 180.0f//if(episodesShadow.isRtl()) 180.0f else 0.0f
@@ -456,10 +455,11 @@ class ResultFragmentTv : Fragment() {
             binding?.apply {
 
                 if (resume == null) {
-                    resumeMode = false
                     return@observeNullable
                 }
-                resumeMode = true
+                resultResumeSeries.isVisible = true
+                resultPlayMovie.isVisible = false
+                resultPlaySeries.isVisible = false
 
                 // show progress no matter if series or movie
                 resume.progress?.let { progress ->
@@ -469,14 +469,8 @@ class ResultFragmentTv : Fragment() {
                         this.max = progress.maxProgress
                         this.progress = progress.progress
                     }
-
                     resultResumeProgressHolder.isVisible = true
-                    resultResumeSeries.isVisible = true
-                    resultPlayMovie.isVisible = false
-                    resultPlaySeries.isVisible = false
                 } ?: run {
-                    //resumeMode = false
-                    resultResumeProgressHolder.isVisible = false
                     resultResumeProgressHolder.isVisible = false
                 }
 
@@ -646,7 +640,7 @@ class ResultFragmentTv : Fragment() {
         }
 
         observeNullable(viewModel.movie) { data ->
-            if (data == null || resumeMode) {
+            if (data == null ) {
                 return@observeNullable
             }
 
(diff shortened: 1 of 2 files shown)


Luna712:
This does fix the button visibility issues for me but I'm not sure how to test the others as I personally couldn't seem to reproduce.

Luna712 (approved):

KillerDogeEmpire:
@KingLucius The fixes are great, they all work good. this has fixed the issues.

KillerDogeEmpire (approved):

KingLucius:
Thanks guys for testing 🙏

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -646,15 +646,14 @@ class ResultFragmentTv : Fragment() {
         }
 
         observeNullable(viewModel.movie) { data ->
-            if (data == null) return@observeNullable
+            if (data == null || resumeMode) {
```
fire-light42, line 649: observe does not have a determined order, so I dont like this. It just sounds like you are making the bug less likely

KingLucius: That's why i am checking the resume this way, i didn't find another way to do this.
I am open for suggestions though

fire-light42: I would move all the visibility logic into resumeWatching and make this function act as a setOnClickListener, then do the nextFocusRightId in the loadpage section. So this makes it so:

Page fills all the metadata about the thing
Movie adds the listener
Resume watching handles the visibility

But if this is too much the viewmodel should be simplified, as all the parts should be independent and support out of order eval

fire-light42: What you can do is making a dummy header view, that way you can have an "X AND Y" visibility, this is very common in cs3 for these things

KingLucius: > I would move all the visibility logic into resumeWatching and make this function act as a setOnClickListener, then do the nextFocusRightId in the loadpage section. So this makes it so:
> 
> Page fills all the metadata about the thing
> Movie adds the listener
> Resume watching handles the visibility
> 
> But if this is too much the viewmodel should be simplified, as all the parts should be independent and support out of order eval
> 
> 

I got that, will rework based on that

KingLucius: > What you can do is making a dummy header view, that way you can have an "X AND Y" visibility, this is very common in cs3 for these things

I don't understand that bro 😂 please elaborate or guide me to a code in CS i can read to get the idea as an example

fire-light42: > I don't understand that bro 😂 please elaborate or guide me to a code in CS i can read to get the idea as an example


Well the idea is to have a view that holds many items, see for example apply_btt_holder https://github.com/recloudstream/cloudstream/blob/81df68e1370e5c08169f07e1f69979691dad0ba6/app/src/main/res/layout/bottom_selection_dialog.xml#L36

Then you can change the visibility of the holder https://github.com/recloudstream/cloudstream/blob/81df68e1370e5c08169f07e1f69979691dad0ba6/app/src/main/java/com/lagradost/cloudstream3/utils/SingleSelectionHelper.kt#L116

But the neat thing is that it has a separate visibility from the view itself. So in this case if I wanted to toggle the visibility of apply by two bools, lets say realShowApply AND showApply then I can set the headers visibility to realShowApply and the apply buttons visibility to showApply. That way both the buttons (cancel and apply) are turned on if realShowApply is true, but the apply buttons is only visible if both realShowApply is true and showApply is true.

In your case I would group all buttons visible in one mode in one header view and turn that view on or off, but this will still keep the visibility state of each individual button.

KingLucius: I handeled most of the visibility in ResumeWatching, and in Movie I check for Resume button visibility to toggle Play button visibility

fire-light42:
I was unable to replicate the bug, @Luna712 can you check?

Luna712:
Sure, I can test this in a couple hours

Luna712 (approved):
This fixes all of the issues I could reproduce

PR #996, opened 2024-03-19
Set play button to first unwatched Episode on TV

### LINKED ISSUE #755: Play button doesn't currently show the next unwatched episode to play

KingLucius:
Fixes #755

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
@@ -781,25 +781,28 @@ class ResultFragmentTv : Fragment() {
 
                 //    resultEpisodeLoading.isVisible = episodes is Resource.Loading
                 if (episodes is Resource.Success) {
-                    val first = episodes.value.firstOrNull()
-                    if (first != null) {
+
+                    val lastWatchedIndex = episodes.value.indexOfLast { ep -> ep.videoWatchState == VideoWatchState.Watched }
+                    val firstUnwatched = episodes.value.getOrElse(lastWatchedIndex + 1) { episodes.value.firstOrNull() }
+
+                    if (firstUnwatched != null) {
                         resultPlaySeriesText.text =
                             when {
-                                first.season != null ->
-                                    "${getString(R.string.season_short)}${first.season}:${getString(R.string.episode_short)}${first.episode}"
-                                else -> "${getString(R.string.episode)} ${first.episode}"
+                                firstUnwatched.season != null ->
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -781,25 +781,28 @@ class ResultFragmentTv : Fragment() {
 
                 //    resultEpisodeLoading.isVisible = episodes is Resource.Loading
                 if (episodes is Resource.Success) {
-                    val first = episodes.value.firstOrNull()
-                    if (first != null) {
+
+                    val lastWatchedIndex = episodes.value.indexOfLast { ep -> ep.videoWatchState == VideoWatchState.Watched }
```
fire-light42, line 785: no? this will get you a wrong result as watchstate is seperate from watched. use getWatchProgress, getDisplayPosition or getRealPosition

fire-light42: watch state if smth that you manually toggle

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
fire-light42: or did you mean watch state? because you need to specify that if that is the case

KingLucius: I want to get the last index of the "marked as watched" episode, then i return rhe next one so this will be the first unwatched

KingLucius: ep.videoWatchState become VideoWatchState.Watched after actual watching or use "Mark as watched"
both works

fire-light42 (approved):

PR #1003, opened 2024-03-21
First steps for multiplatform API

CranberrySoup:
This can act as a framework wherein we can port the entire API surface of extensions to a multi-platform project.
This will make extensions portable and might make it possible to sandbox them in the future.

Unsolved problems so far:
- ~BuildConfig is hardcoded. It could probably be fixed with reflection~
- ~cloudstream.jar is not tested~
- ~Maven publishing of the jvm jar~


CranberrySoup:
This should now be ready.

PR #1005, opened 2024-03-21
fix: extension activity interruption

int3debug:
Screenshot_2024-03-20-13-50-16-453_com lagradost cloudstream3 prerelease debug_1

Show this activity if only local extension loaded


fire-light42:
I dont understand, what does this fix?

PR #1019, opened 2024-03-30
refactor: improve add new repository dialog & update gradle to 8.7

IndusAryan:
no one uses all extensions, plus only few of them work, but users always download all the extensions even disabled one, with current dialog, now we tell them that repo has been added and they can download extensions what they want along with a staturory warning the cs3 doesnt take responsibility of third party extensions as people have been using lots of 3rd party extensions.
tell me if message can be improved, i tried my best 🐯


mizu-048:
"CloudStream3 does not host any of the content streamed through it and thus the developers and contributing members cannot be held responsible for the violation of laws in ones locality.

The extensions can execute arbitrary code and read all of cloudstream3's data, therefore proceed with caution. No entity can be held responsible for the temporary or permanent damage caused to your device.

The extensions are not to be affiliated with cloudstream3 and thus no support will be provided regarding their utilization."

How's that? bit to long but ye

mizu-048:
atleast fix the "does not takes"

IndusAryan:
disclaimer is already present in app settings as well as repo readme, not needed everywhere

mizu-048:
i meant the warning

PR #1025, opened 2024-04-03
feat(ui): new dialog on adding repository and auto redirection

IndusAryan:


IndusAryan:
> Explain why this should be added and what it does. I can read the code, but I dont understand the intent behind it.

Out of all extensions for every repository only few work, but people press download button and add all of them , even disabled ones, that contrubutes to huge amount of bloat as well as waste of time, for example an Indian presses download button when adding hexated repo, instead of only 2 hindi extesnions he downloads whole bunch of indonesian, phillipino and sex extensions that are of no use to him, not only uninstalling them is a waste of time, but produces bloat in dialog box, in new way we how a reposiotry doenload succesfully dialog and redirect that particular repo to download her desired extensions, and ask them that we dont take any responsibility of 3rd party extensions, because nowadays people are using some 3rd party sites that we are not affiliated with, and thus they will download only those with their flags and languages.

this is a win win fr, btw big fan bhai big fan

Luna712:
For the record I quite like this PR as I accidentally press download all sometimes myself then spend time removing extensions so I do think this is a good idea personally.

fire-light42:
I dont see any obvious bugs from reading the code, so if I dont find anything in testing this will be merged.

app/src/main/res/values/strings.xml
```diff
@@ -606,7 +609,8 @@
     <string name="view_public_repositories_button">View community repositories</string>
     <string name="view_public_repositories_button_short">Public list</string>
     <string name="uppercase_all_subtitles">Uppercase all subtitles</string>
-    <string name="download_all_plugins_from_repo">Download all plugins from this repository?</string>
+    <string name="download_all_plugins_from_repo">%1$s has been added. You can install your desired extensions from 𝗦𝗲𝘁𝘁𝗶𝗻𝗴𝘀 > 𝗘𝘅𝘁𝗲𝗻𝘀𝗶𝗼𝗻𝘀 > %1$s.
```
fire-light42, line 612: The key for this NEEDS to be changed, as all languages other than english will just be confusing because they have not translated this yet. Also this is a bit long and no user will read this entire shit and will just press OK. Make it easier to read and remove `CloudStream 3 does not takes any responsibility for using third-party extensions neither provides support for them`

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -429,25 +430,37 @@ object AppUtils {
...
-            val context = this
-            val builder: AlertDialog.Builder = AlertDialog.Builder(this)
-            builder.setTitle(
-                repositoryName
-            )
-            builder.setMessage(
-                R.string.download_all_plugins_from_repo
-            )
+            val builder : AlertDialog.Builder = AlertDialog.Builder(this)
             builder.apply {
```
fire-light42, line 455: why not have two buttons, one for open and one for dismiss????

fire-light42:
This will be merged after further testing.

app/src/main/java/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -429,25 +430,39 @@ object AppUtils {
...
+            }
+        }
 
-                setNegativeButton(R.string.no) { _, _ -> }
+        runOnUiThread {
+            AlertDialog.Builder(this).apply {
+                setTitle(repositoryName)
+                setMessage(R.string.download_all_plugins_from_repo)
+                // won't show "open button" when adding from settings
+                if (!isExtensionsFragment) {
```
fire-light42, line 458: Why??????

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/utils/AppUtils.kt
fire-light42: Nah, removing that option just makes it a waste of a popup then

IndusAryan: what to do then, remove popup here or add button ? without popup we can show a toast of added successfuly, non intrusive also

fire-light42: Users will just be confused it you "download" a repo, but the extentions does not get downloaded nor do they get redirected or even get a dialog about it. The dialog *should* have the option to just open it, because it is not obvious that more needs downloading, nor that it is clickable.

fire-light42 (approved):
Nice 👍

PR #1026, opened 2024-04-06
Trakt meta provider for extensions

KingLucius:


app/src/main/java/com/lagradost/cloudstream3/metaproviders/TraktProvider.kt
```diff
@@ -0,0 +1,585 @@
...
+        val req = app.get(
+            url = imageApiUrl,
+            headers = mapOf(
+                "accept" to "application/json",
+            )
+        ).toString()
+        return parseJson<Images>(req)
+    }
+
+    private suspend fun getImages(tmdbId: String, movie: Boolean = true) : Images {
```
fire-light42, line 320: 👎 This is really really bad, we never want to make one get request for every search response as that more or less creates 100 requests for 1 search instead of 1. Is there any way to get the image of a searchresponse without making an additional request?

fire-light42 (changes requested):
1. Prevent excessive get requests
2. I would like to have ALL fields that are related to JSON to have a JsonProperty

Expect additional commends after real testing.

KingLucius:
Is there any issue about this PR?

PR #1032, opened 2024-04-09
Mark Episode as Watched if progress over 90%

KingLucius:
This solves the issue of filtering the default episode of play button, setting the episode as watched if progress is over 90%


fire-light42:
This is not the correct fix. Watched is what you **manually** change. Fix `default episode of play button` by using the watch % instead.

PR #1035, opened 2024-04-11
Defaults Play button to first unwatched Episode

KingLucius:


app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentTv.kt
```diff
@@ -782,7 +783,7 @@ class ResultFragmentTv : Fragment() {
                 //    resultEpisodeLoading.isVisible = episodes is Resource.Loading
                 if (episodes is Resource.Success) {
 
-                    val lastWatchedIndex = episodes.value.indexOfLast { ep -> ep.videoWatchState == VideoWatchState.Watched }
+                    val lastWatchedIndex = episodes.value.indexOfLast { ep -> ep.getWatchProgress() >= (NEXT_WATCH_EPISODE_PERCENTAGE / 100).toFloat() }
```
fire-light42, line 786: (90 / 100).toFloat() ? Wont that just be zero due to int rounding?

fire-light42: I think you mean to do NEXT_WATCH_EPISODE_PERCENTAGE.toFloat() / 100.0f

KingLucius: > (90 / 100).toFloat() ? Wont that just be zero due to int rounding?

That would be 0.90f

KingLucius: > I think you mean to do NEXT_WATCH_EPISODE_PERCENTAGE.toFloat() / 100.0f

Same i think , i will double check again

PR #1047, opened 2024-04-17
bugfix: fixing regex special chars break it

int3debug:
Special chars breaks regex
Regex is differently between jvm and android
Credit for bugfix @Lag bug found @int3debug


app/src/main/java/com/lagradost/cloudstream3/utils/JsUnpacker.kt
```diff
@@ -1,6 +1,6 @@
 package com.lagradost.cloudstream3.utils
 
-import com.lagradost.cloudstream3.mvvm.logError
+import com.lagradost.safefile.logError
```
fire-light42, line 3: Wrong import

fire-light42 (changes requested):

fire-light42 (approved):

PR #1049, opened 2024-04-20
Filtering first unwatched episode respects watched state

KingLucius:


Luna712:
@fire-light42 seems build is failing but doesn't seem related to this but rather NewPipeExtractor

PR #1056, opened 2024-04-25
Ported more files for multiplatform

CranberrySoup:
Some stuff in MainActivity was moved to the companion object to prevent class collision between the library and app.
I have yet to figure out a test to catch this automatically.


LinkofHyrule89:
Can you make Cloudstream into a PWA using WASM?

fire-light42:
> Can you make Cloudstream into a PWA using WASM?

No that is not how any of this works. Also you are in the wrong place to ask that, create an issue instead for these questions.

fire-light42:
Lets hope this does not break main

PR #1058, opened 2024-04-27
feat(ui): Episode Air date & Upcoming countdown

KingLucius:
1. Shows EpisodeCardViewHolderSmall if both poster & description is null.
2. Adds Episode Air date.
3. Upcoming count down if the episode is not released yet.

image


fire-light42 (approved):
Looks good, will merge after testing

PR #1060, opened 2024-04-27
Don't check year when checking duplicates if year is empty

Luna712:
Some sources don't use year which makes this not match when it really should match


fire-light42:
Nice catch 👍

PR #1063, opened 2024-04-28
hotfix: Remove resume information

int3debug:
Remove resume information on failed download
Fixes this bug and you can try download again

66g3zs1vohxlelwq


app/src/main/java/com/lagradost/cloudstream3/ui/download/button/PieFetchButton.kt
```diff
@@ -167,6 +170,9 @@ open class PieFetchButton(context: Context, attributeSet: AttributeSet) :
         this.setPersistentId(card.id)
         view.setOnClickListener {
             if (isZeroBytes) {
+                if (getKey(KEY_RESUME_PACKAGES, card.id) != null) {
```
fire-light42, line 173: No need to check, just call removeKey

fire-light42 (commented):
I am a bit unsure if this is the correct way to do it

PR #1065, opened 2024-04-29
Updates and Chillx Extractor Updated

phisher98:
Updates and Chillx Extractor Updated

Kudos to @rushi-chavan for updated Keys

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/extractors/Chillx.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/extractors/Chillx.kt
@@ -2,9 +2,7 @@ package com.lagradost.cloudstream3.extractors
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.lagradost.cloudstream3.*
-import com.lagradost.cloudstream3.extractors.helper.*
 import com.lagradost.cloudstream3.extractors.helper.AesHelper.cryptoAESHandler
-import com.lagradost.cloudstream3.utils.AppUtils
 import com.lagradost.cloudstream3.utils.ExtractorApi
 import com.lagradost.cloudstream3.utils.ExtractorLink
 import com.lagradost.cloudstream3.utils.M3u8Helper
@@ -29,6 +27,7 @@ class AnimesagaStream : Chillx() {
     override val name = "AnimesagaStream"
     override val mainUrl = "https://stream.anplay.in"
 }
+@Suppress("NAME_SHADOWING")
 open class Chillx : ExtractorApi() {
     override val name = "Chillx"
     override val mainUrl = "https://chillx.top"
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/lagradost/cloudstream3/extractors/Chillx.kt
```diff
@@ -24,34 +22,30 @@ class Watchx : Chillx() {
...
-                headers = mapOf(
-                    "Accept" to "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
-                    "Accept-Language" to "en-US,en;q=0.5",
-                )
+                referer = url,
             ).text
         )?.groupValues?.get(1)
-        val decrypt = cryptoAESHandler(master ?: return, getKey().toByteArray(), false)?.replace("\\", "") ?: throw ErrorLoadingException("failed to decrypt")
-
+        val key = app.get("https://raw.githubusercontent.com/rushi-chavan/multi-keys/keys/keys.json").parsedSafe<Keys>()?.key?.get(0) ?: throw ErrorLoadingException("Unable to get key")
```
fire-light42, line 44: Make this cached, we dont want to ddos github for every url.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/extractors/Chillx.kt
phisher98: I have assigned that as key can be changed and previously as well its fetching from a Github json hosted on github so do you want me to assign key static in Extractor ?

fire-light42: @phisher98 What I mean is that you should do something like this, as then you dont have to do an extra get request for every time you use this extractor
```kotlin
companion object {
  private val key : String? = null
  suspend fun getKey() {
    key ?: app.get("...").parsedSafe<Keys>()?.key?.get(0)?.also { key = it }
  }
}

```

b4byhuey:
private var key: String? = null

val key = fetchKey() ?: throw ErrorLoadingException("Unable to get key")
		
val decrypt = cryptoAESHandler(master ?: return, key.toByteArray(), false)
            ?.replace("\\", "")
            ?: throw ErrorLoadingException("failed to decrypt")

private suspend fun fetchKey(): String? {
        return app.get("https://raw.githubusercontent.com/rushi-chavan/multi-keys/keys/keys.json")
            .parsedSafe<Keys>()?.key?.get(0)?.also { key = it }
    }


@phisher98 try this

phisher98:
@b4byhuey Yeah That what i was saying but Thanks for the help

fire-light42:
The key is still not cached

fire-light42 (approved):

PR #1071, opened 2024-05-03
Show Season number for next airing episode

KingLucius:
- Season number support to NextAiring of MainApi (not sure about back compatibility).
- Fix next air text in TV Layout.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
@@ -1452,6 +1452,7 @@ fun TvType?.isEpisodeBased(): Boolean {
 data class NextAiring(
     val episode: Int,
     val unixTime: Long,
+    val season: Int? = null,
 )
 
 /**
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1452,6 +1452,7 @@ fun TvType?.isEpisodeBased(): Boolean {
 data class NextAiring(
     val episode: Int,
     val unixTime: Long,
+    val season: Int? = null,
```
fire-light42, line 1455: This needs a backwards compatible constructor that holds episode and unixtime

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: having the last argument as null might look like that, but is in fact not from the JVM, so it will give a fatal crash for all extentions using it

KingLucius: Oh, didn't know that
Now I got why you are so careful when touching MainApi
I will do it today

fire-light42 (approved):

PR #1074, opened 2024-05-03
feat(ui): Hide Downloads & Settings Back button on TV

KingLucius:


fire-light42:
What about emulator? I think emulators should have an back button as not all mouse have a back button.

KingLucius:
> What about emulator? I think emulators should have an back button as not all mouse have a back button.

You're right, updated

PR #1084, opened 2024-05-09
Add basic fcast support

CranberrySoup:
This supports starting videos using fcast using the long press menu.

Things to improve in the future:

- Proper cast UI with play, pause and seek
- Subtitle support
- CloudStream acting as a proper fcast receiver

These should not be too hard to implement given the now existing framework. Subtitle support is probably the easiest to implement in case anyone else wants to contribute.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
@@ -83,6 +83,10 @@ import com.lagradost.cloudstream3.utils.DataStoreHelper.setVideoWatchState
 import com.lagradost.cloudstream3.utils.DataStoreHelper.updateSubscribedData
 import com.lagradost.cloudstream3.utils.UIHelper.clipboardHelper
 import com.lagradost.cloudstream3.utils.UIHelper.navigate
+import com.lagradost.cloudstream3.utils.fcast.FcastManager
+import com.lagradost.cloudstream3.utils.fcast.FcastSession
+import com.lagradost.cloudstream3.utils.fcast.Opcode
+import com.lagradost.cloudstream3.utils.fcast.PlayMessage
 import kotlinx.coroutines.*
 import java.io.File
 import java.util.concurrent.TimeUnit
@@ -1509,6 +1513,13 @@ class ResultViewModel2 : ViewModel() {
                         )
                     )
                 }
+
+                if (FcastManager.currentDevices.isNotEmpty()) {
+                    options.add(
+                        txt(R.string.player_settings_play_in_fcast) to ACTION_FCAST
+                    )
+                }
+
                 options.add(txt(R.string.episode_action_play_in_app) to ACTION_PLAY_EPISODE_IN_PLAYER)
 
                 for (app in apps) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/lagradost/cloudstream3/utils/fcast/FcastSession.kt
```diff
@@ -0,0 +1,57 @@
...
+        ioSafe {
+            val socket = acquireSocket()
+            val outputStream = DataOutputStream(socket.getOutputStream())
+
+            val json = message?.toJson()
+            val content = json?.toByteArray() ?: ByteArray(0)
+            val size = content.size + 1
+
+            val sizeArray = ByteArray(4) { num ->
+                (size shr (8 * num) and 0xff).toByte()
```
fire-light42, line 48: I assume this is because of Little Endian, if so then add a comment linking to https://gitlab.com/futo-org/fcast/-/wikis/Protocol-version-1

fire-light42: I think you can just use reverseBytes here as java is Big Endian

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1684,6 +1695,39 @@ class ResultViewModel2 : ViewModel() {
...
+                val devices = FcastManager.currentDevices.toList()
+                postPopup(
+                    txt(R.string.player_settings_select_cast_device),
+                    devices.map { txt(it.name) }) { index ->
+                    if (index == null) return@postPopup
+                    val device = devices.getOrNull(index)
+
+                    acquireSingleLink(
+                        click.data,
+                        LoadType.Chromecast,
```
fire-light42, line 1708: This should be its own param as it will use the noreferer param in the extraction due to `isCasting = LoadType.Chromecast == type`

fire-light42 (commented):
Code looks good 👍

app/src/main/java/com/lagradost/cloudstream3/utils/fcast/FcastSession.kt
CranberrySoup: Unable to use this nicely due to ByteArray

oathzed:
> This supports starting videos using fcast using the long press menu.
> 
> Things to improve in the future:
> 
> * Proper cast UI with play, pause and seek
> * Subtitle support
> * CloudStream acting as a proper fcast receiver
> 
> These should not be too hard to implement given the now existing framework. Subtitle support is probably the easiest to implement in case anyone else wants to contribute.

Well done on this.

It would be nicer UX though if the receiver showed up under your cast button like all the other DNLA receivers. So it's not under a long press. Also if possible if a source doesn't work, then it will try another one till it works, just like how the internal player works.

fire-light42:
> It would be nicer UX though if the receiver showed up under your cast button like all the other DNLA receivers. So it's not under a long press. Also if possible if a source doesn't work, then it will try another one till it works, just like how the internal player works.

Add <ins>***basic***</ins> fcast support

oathzed:
> > It would be nicer UX though if the receiver showed up under your cast button like all the other DNLA receivers. So it's not under a long press. Also if possible if a source doesn't work, then it will try another one till it works, just like how the internal player works.
> 
> Add _**basic**_ fcast support

I understand, I'm just explaining how it would be better eventually when it gets improved Incase this wasn't how y'all were gonna do it

PR #1085, opened 2024-05-10
feat(TV UI): Fix online subtitles dialog focus

KingLucius:
- Fixes Subtitle language filter focus on TV.
- Fixes moving right/left between Search, Year & Language filter.


fire-light42:
I did not test this, but the code looks ok

PR #1091, opened 2024-05-14
chore: update material,kotlin compiler,newpipe extractor,rhino-js,guava,corektx

IndusAryan:


Luna712:
@IndusAryan @fire-light42 this messes some layout up (or I at least assume it was this)

For example Simkl rating
| Screenshot_20240528_012045_CloudStream | 
|:--:| 
| *Before* |

| Screenshot_20240528_012034_CloudStream Beta |
|:--:| 
| *After* |

IndusAryan:
messed up how ? design changes with time, if original style is still present, send me the xml of this , will see .

fire-light42:
> messed up how ? design changes with time, if original style is still present, send me the xml of this , will see .

https://m3.material.io/components/sliders/overview it is simply how material 3 is

Luna712:
> > messed up how ? design changes with time, if original style is still present, send me the xml of this , will see .
> 
> https://m3.material.io/components/sliders/overview it is simply how material 3 is

@fire-light42 

I'm aware of that but IMO it looks ugly how it is now and it should at least be tuned up a bit even if it can't be exactly how it was before.

Issue #237, opened 2022-12-06
More playback speed options


imDusht:
### Describe your suggested feature

App has playback speed upto 2x but sometimes depending on the playing content it feels slow, so i would like more options maybe upto 3x ur more

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

imDusht:
> insane
> 
> how can 2x be slow

As i said depending on the playing content sometimes it feels slow, like i was watching a movie and felt like that so suggested this

imDusht:
Thanks for your concern 😊

PR #1092, opened 2024-05-14
More Video Speed Options

homealone944:
Add additional options for video speed:  2.5x and 3x

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -516,10 +516,12 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
                 "1.4x",
                 "1.5x",
                 "1.75x",
-                "2x"
+                "2x",
+                "2.5x",
+                "3x"
             )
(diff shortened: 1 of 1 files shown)


KillerDogeEmpire:
@homealone944 add up to 5x. 

(This is a joke. Who needs even 3x. 2x is enough)

homealone944:
> @homealone944 add up to 5x.
> 
> (This is a joke. Who needs even 3x. 2x is enough)

I watch my youtube at 2.5x, sometimes 3x. Been disliking having to go down to 2x when i use cloudstream.  I would be alright just having it in my own fork, but when I built it, there was an error connecting my Simkl account, so I figured I'd try to get it in the main branch.

> error was something like client_id was set to null when trying to connect Simkl Account

fire-light42:
While we appreciate your pr, this was intentionally not added, see https://github.com/recloudstream/cloudstream/issues/237

The reason why Simkl stopped working is because you need to add your own simkl api key as a secret key, see https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SimklApi.kt#L144 and https://github.com/recloudstream/cloudstream/blob/master/app/build.gradle.kts#L82

PR #1093, opened 2024-05-15
Use Season index to search Subs

KingLucius:
If Extenstion is using the SeasonData and doesn't provide both (season & displaySeason) as displaySeason is nullable.

https://github.com/recloudstream/cloudstream/blob/ee4d1dedc5adb1be656a05d1ee0f41b11f9d0a84/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt#L2319
https://github.com/recloudstream/cloudstream/blob/ee4d1dedc5adb1be656a05d1ee0f41b11f9d0a84/app/src/main/java/com/lagradost/cloudstream3/metaproviders/TraktProvider.kt#L172
using seasonIndex assures subtitles search uses the right season number in all cases.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
@@ -265,9 +265,10 @@ class GeneratorPlayer : FullScreenPlayer() {
 
         when (val newMeta = currentMeta) {
             is ResultEpisode -> {
+                Log.d("King", "TempMeta => $newMeta")
                 if (!newMeta.tvType.isMovieType()) {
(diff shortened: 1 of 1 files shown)


fire-light42:
No, this is wrong seasonIndex is a unique value, where as seasonData.displaySeason is the "real" season number. This means for example that if we have duplicate seasons the seasonIndex wont be correct. seasonIndex is only used internally as it is unique

KingLucius:
> No, this is wrong as seasonIndex is a unique value, where as seasonData.displaySeason is the "real" season number. This means for example that if we have duplicate seasons the seasonIndex wont be correct. seasonIndex is only used internally as it is unique

OK I will solved another way
Thanks

PR #1101, opened 2024-05-21
SubDL Account login support

KingLucius:
- Account login support, which solves the rate limit issue.
- Hearing impaired indicator.


fire-light42 (commented):
I dont see anything wrong with the code, will be merged after testing

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/Subdl.kt
```diff
@@ -1,21 +1,80 @@
...
-class SubDL : AbstractSubProvider {
-    //API Documentation: https://subdl.com/api-doc
-    val mainUrl = "https://subdl.com/"
-    val name = "SubDL"
+class SubDlApi(index: Int) : InAppAuthAPIManager(index), AbstractSubApi {
     override val idPrefix = "subdl"
+    override val name = "SubDL"
+    override val icon = R.drawable.subdl_logo_big
+    override val requiresPassword = true
+    override val requiresUsername = true
```
fire-light42, line 24: This is wrong, it should be requiresEmail as you must input the mail to login

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/Subdl.kt
```diff
@@ -181,7 +181,7 @@ class SubDlApi(index: Int) : InAppAuthAPIManager(index), AbstractSubApi {
     }
 
     data class SubtitleOAuthEntity(
```
fire-light42, line 183: I did not see this before, but if you are storing this as a the SUBDL_SUBTITLES_USER_KEY then it *needs* JsonProperty

fire-light42 (changes requested):
After this, then it should be ready to merge

fire-light42:
I also see that you are storing the password, that might want to be avoided via https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/BackupUtils.kt#L50

KingLucius:
> I also see that you are storing the password, that might want to be avoided via https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/utils/BackupUtils.kt#L50

Nice catch, updated

fire-light42 (approved):

PR #1117, opened 2024-06-04
feature(ui): hide NSFW plugins

int3debug:
Hide NSFW plugins if Settings / Providers NSFW is disabled

Before: 
s2j04sokq87zxm2j

After:
ym5de4x6k35xe96w


fire-light42:
Did not test it but, looks good. 👍

PR #1123, opened 2024-06-08
feat(TV UI): Accounts PIN login support

KingLucius:
- Default method on TV, If sync api supports authentication via PIN.
- You can local authenticating also.
- QR scanning for URL opening.

**Note**: CS logo can be added to QR code but the code was overkill as QR Lib recieve the logo as ByteArray so I removed it.


IndusAryan:
> * Default method on TV, If sync api supports authentication via PIN.
> 
>     * You can local authenticating also.
> 
>     * QR scanning for URL opening.
> 
> 
> **Note**: CS logo can be added to QR code but the code was overkill as QR Lib recieve the logo as ByteArray so I removed it.

can we embed an image overflow on the qr image itself in centre

KingLucius:
> > * Default method on TV, If sync api supports authentication via PIN.
> >   
> >   * You can local authenticating also.
> >   * QR scanning for URL opening.
> > 
> > **Note**: CS logo can be added to QR code but the code was overkill as QR Lib recieve the logo as ByteArray so I removed it.
> 
> can we embed an image overflow on the qr image itself in centre

it's already supported, only need the logo to be passed as byteArray, I have succesfully did it but then removed the code, I can do it again if you guys sees it's better

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsAccount.kt
```diff
@@ -133,7 +142,85 @@ class SettingsAccount : PreferenceFragmentCompat(), BiometricAuthenticator.Biome
...
+                            }
+                            val dialog = builder.create()
+
+                            ioSafe {
+                                try {
+                                    val pinCodeData = api.getDevicePin()
+                                    if (pinCodeData == null) {
+                                        activity.runOnUiThread {
+                                            showToast(
+                                                activity.getString(R.string.device_pin_error_message)
```
fire-light42, line 171: 1. toast is thread safe
2. toast can take a stringres

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsAccount.kt
```diff
@@ -133,7 +142,85 @@ class SettingsAccount : PreferenceFragmentCompat(), BiometricAuthenticator.Biome
...
+                                            activity.getString(R.string.device_pin_expired_message)
+                                        )
+                                        dialog.dismissSafe(activity)
+                                    }
+                                } catch (e: Exception) {
+                                    logError(e)
+                                }
+                            }
+                        } else {
+                            api.authenticate(activity)
```
fire-light42, line 241: this is really weird to read, reverse this if statment so this is on top and else is the rest of the code

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsAccount.kt
```diff
@@ -133,7 +142,85 @@ class SettingsAccount : PreferenceFragmentCompat(), BiometricAuthenticator.Biome
...
+                                                        activity.getString(R.string.authenticated_user)
+                                                            .format(
+                                                                api.name
+                                                            )
+                                                    )
+                                                dialog.dismissSafe(activity)
+                                            }
+                                            return@ioSafe
+                                        }
+                                        delay(1000)
```
fire-light42, line 227: this entire setup is weird and prone to errors. Just use something like this https://developer.android.com/reference/android/os/CountDownTimer instead

fire-light42 (changes requested):
This is my initial review, more might come after I have tested it

app/src/main/res/values/strings.xml
```diff
@@ -775,4 +777,8 @@
     <string name="custom_media_singluar">Media</string>
     <string name="reset_btn">Reset</string>
     <string name="cs3wiki">CloudStream Wiki</string>
+    <string name="device_pin_url_message">Visit <b>%s</b> on your smartphone or computer and enter the above code</string>
+    <string name="device_pin_error_message">"Can't get the device PIN code, try local authentication"</string>
```
fire-light42, line 781: Why prefix and suffix with " ????

KingLucius: Mistake

fire-light42 (approved):
Works, and looks very nice 👍

KingLucius:
> Works, and looks very nice 👍 

Thanks, merged in the right time
I will add this to trakt sync api also 👍

PR #1124, opened 2024-06-08
feat(Extensions): Trakt season names remove

KingLucius:
if DisplaySeason is null, it conflicts with subs provider so I have removed it from trakt as most seasons doesn't have names, only the special season has a name
as the option now is to add DisplaySeason in trakt and that leads to display seasons to be duplicated as trakt uses the below as seasons names:
Specials
Season 1
Season 2


fire-light42:
Ok?

KingLucius:
> Ok?

Yes

PR #1137, opened 2024-06-16
feat(UI): Reorganize Settings

KingLucius:
- Accounts Section & Remove "account" from title.
- Security Section for Biometric that is hidden on TV.
- Move "send logs" to "Action" section.


fire-light42:
nice 👍

PR #1139, opened 2024-06-17
feat(TV UI): Player's Top controls redesign

KingLucius:
- Fixes Title alignment in case of RTL languages.
- Adds "Restart" & "Next Episode" in top controls bar.
- Removes the "Next Episode" button under progress seekbar.
- Changes "Tracks" icon.
- Margin for top bar is same on right & left.
- Removes the back button from player loading UI.

image


fire-light42:
This might not be ready for review yet, but when pressing the reset button it might be nice to add a Y/N popup/popdown to make it impossible to make a missinput. I would also suggest that playerGoForward is always visible when hasNextEpisode because right now it hides itself when under 50%, this is due to mobile!=tv.

KingLucius:
> This might not be ready for review yet, but when pressing the reset button it might be nice to add a Y/N popup/popdown to make it impossible to make a missinput. I would also suggest that playerGoForward is always visible when hasNextEpisode because right now it hides itself when under 50%, this is due to mobile!=tv.

- Do you mean Subtitles sync reset button ?
- playerGoForward is visible from the beginning not after 50% as depends on hasNextEpisode not PreloadNextEpisode

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1095,7 +1095,8 @@ class GeneratorPlayer : FullScreenPlayer() {
         }
 
         playerBinding?.playerSkipOp?.isVisible = isOpVisible
-        playerBinding?.playerSkipEpisode?.isVisible =
+
+        playerBinding?.playerGoForward?.isVisible =
```
fire-light42, line 1099: `playerGoForward is visible from the beginning not after 50% as it depends on hasNextEpisode not PreloadNextEpisode` 
No? `isOpVisible = percentage < SKIP_OP_VIDEO_PERCENTAGE` => `playerGoForward.isVisible = percentage > SKIP_OP_VIDEO_PERCENTAGE` meaning that in the event on an anime, the top button is not visible on TV. This was the behavior before on phone as we did not want to clutter the bottom bar, but on TV this should not be the behavior.

fire-light42:
Any update on this? The UI looks nice from your screenshot.

KingLucius:
> Any update on this? The UI looks nice from your screenshot.

Just got busy, will work on it next weekend

KingLucius:
I was working on redesign for the player with logo, description & episode view but let's delay this as I am super busy these days
later I can resume work on the redesign

DestroyerBDT:
So the back button will still be there

DestroyerBDT:
Also, we don't need a restart button

Luna712:
> Also, we don't need a restart button

@DestroyerBDT 

I disagree, on TV, reminding to the beginning when you want can be quite annoying.

fire-light42 (approved):
Code looks good, however I would like to have text under the icons when hovering like Netflix. 👍

KingLucius:
> Code looks good, however I would like to have text under the icons when hovering like Netflix. 👍 

Ok will work on it

KingLucius:
> Code looks good, however I would like to have text under the icons when hovering like Netflix. 👍

Done

Feels like this PR needs a lot of testings, don't have time for deep test these days so I need your help testing this

fire-light42:
Why was this closed?

KingLucius:
> Why was this closed?

The rebase ruined it for some reason, created a new one with additional fixes and will push today for review

PR #1140, opened 2024-06-18
feat(Extensions): Consider time zone in Trakt durations

KingLucius:
- Fixes wrong time zone for the upcoming durations.
- Optimized Imports.
- Auto formatted the code.


fire-light42:
Not tested, however it looks good

PR #1145, opened 2024-06-19
Downloads: performance improvements and merge adapters

Luna712:
### Fixes
* The download icon bug is fixed
* The bug where when scrolling the bytes sometimes use from the wrong item is fixed
* Lag is improved — when you get to 20-30 items it still lags but it is still about 2x better then before this change with the same amount of downloads
* Fixes extra unnecessary reload items that cause it to visually refresh items every time you open the downloads tab, or go back from a child entry
* Fixes the issue of when a download is deleted it causes the byte size to disappear from the next item directly underneath the one that was deleted until you went out of downloads and back in
* The go to child arrow icon actually displays now

### Cleanup/Refactoring
* Major cleanup and refractoring to DownloadFragment and DownloadChildFragment
* Merge DownloadHeaderAdapter and DownloadChildAdapter into a single more modernized DownloadAdapter to create a more unified code base and making it easier in the future for common code handling for things planned in the future such as multi-delete handling
* Use DiffUtil to more seamlessly add or remove items from the adapter by only adding, removing, or reloading the specific items we need to rather than reloading the entire adapter dataset

# Performance Profiling
* **Before**: took approximately 18 seconds to load the downloads tab for the first time
* **After**: took approximately 9 seconds to load the downloads tab for the first time

**NOTE**: Both times use the same exact downloads. However, this result could just be a coincidence.

## Future (Potential)
*Just tracking my future plans here so I can remember what my next PRs should be when improving the downloads interface and performance.*

### Performance
*While this PR does improve performance with large items it still has the potential to lag, these changes may significantly improve that.*
* Improve caching for the file sizes for each download — e.g. cache the total file size and use cached size for full pie button if 100% downloaded
* Don't display current size / total size if it is 100% downloaded, just show one of them. It doesn't make sense to e.g. show 2GB / 2GB to me, and it makes it more performant to use cached value for total size only
* Improve Glide caching for thumbnails to handle more performative

### Features
* Add synopsis/description from metadata to download display
* Add support for multi-delete downloads
* Add support for deleting entire series by deleting a series item which would delete all children in it as well
* Add thumbnail display to children (maybe, depends on if it ends up negatively impacting performance)


Luna712:
@fire-light42

I tested this now and it seems to work with pretty significant performance improvements. For one the updateList in view model was always refreshing data causing it to sometimes refresh multiple times on a single request which this also appears to correct (at least visually stops the items from refreshing unless they actually have updates to refresh) additionally the merginf of the adapters was just done as a step towards the multi download select because it allows more common code rather than a bunch of duplicate stuff to work around the difference in adapters. So it just makes future expansion a bit easiser.

If you don't want it done like this I can do it another way if you have better suggestions. Overall this should be ready for review though now.

Luna712:
Okay so further testing, this actually seems to fix a lot more problems then I even intended.
* The download icon bug is fixed
* The bug where when scrolling the bytes sometimes use from the wrong item is fixed
* Lag is improved — when you get to 20-30 items it still lags but it is still about 2x better then before this change with the same amount of downloads
* It fixes extra unnecessary reload items that cause it to visually refresh items every time you open the downloads tab, or go back from a child entry
* Fixes the issue of when a download is deleted it causes the byte size to disappear from the next item directly underneath the one that was deleted until you went out of downloads and back in — I am not to confident this is 100% fixed but it does appear to be significantly improved as it didn't reproduce like it used to 100% of the time

Luna712:
Okay my new version should be done. I decided to go further with this PR and make things even better and more organized so it's easier to work on my next project: multi-delete downloads. Making the code cleaner and easier to manage helps make that a bit easier as well...

fire-light42 (commented):
Just looking at the code, I dont see any obvious bug or mistake. It will be merged after testing if I dont find anything

Luna712:
@fire-light42 thanks so much for merging. This does seem to have done the trick with performance. The lag is almost completely solved now. There is still a minor bug with the file size but I'll try and patch it later. This seemed tk have fixed all of the most annoying things for me though. So glad we got this part done.

PR #1146, opened 2024-06-20
feat(ui): show currently syncing api logo on navigation bar and rail

IndusAryan:


app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -570,18 +568,37 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
...
+                navRailView.apply {
+                    menu.findItem(R.id.navigation_library)?.setIcon(drawable)
+                }
+                navView.apply {
+                    menu.findItem(R.id.navigation_library)?.setIcon(drawable)
+                }
+            }
+
+            fun changeLibraryIcon() {
+                when (lastAPI) {
```
fire-light42, line 590: Why? You quite literally observe currentApiName so why not use that?

IndusAryan: because if we do that, then it requires the user to go to library everytime, then it starts observing it, thus everytime the icon will be resetted, and until LVM is not observed

IndusAryan:
> Is it intended that the icon only updates when you click on another icon? Excuse me Sir I just saw it

sort of intentional, as we want to change the icon when user finalize his api and exit library, otherwise if he is only changing api inside to see his catalogue then every time icon will change, thats why its called lastAPI not currentAPI

fire-light42 (changes requested):
You should not place the update code in nav change, place in in the create and observe correctly.

IndusAryan:
> You should not place the update code in nav change, place in in the create and observe correctly.

if i dont put the code within update navbar then icon doesnt change, the multiple functions like updatenavbar, onconfiguration changed are preventing seamless icon change

fire-light42:
> if i dont put the code within update navbar then icon doesnt change, the multiple functions like updatenavbar, onconfiguration changed are preventing seamless icon change

Nuh uh, this is the code I ran without any issue

```kotlin
// In MainActivity.kt::onCreate
        val libraryViewModel = ViewModelProvider(this@MainActivity)[LibraryViewModel::class.java]
        libraryViewModel.currentApiName.observe(this@MainActivity) {
            val icon = libraryViewModel.currentSyncApi?.icon?: R.drawable.library_icon
            binding?.navRailView?.menu?.findItem(R.id.navigation_library)?.setIcon(icon)
            binding?.navView?.menu?.findItem(R.id.navigation_library)?.setIcon(icon)
        }
```

IndusAryan:
ooh, i was setting it in onCreateView() to not slow down app start but if i am right to think that, but in this the library folder icon doesnt appear, when set to Local we get the server like icon

fire-light42:
> ooh, i was setting it in onCreateView() to not slow down app start but if i am right to think that, but in this the library folder icon doesnt appear, when set to Local we get the server like icon, if it looks good to you then OK, otherwise it needs to be dealt in this way like if apiname == local set r.drawable.libicon

Yep, just add a if statement

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1114,7 +1114,13 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
         // Change library icon with logo of current api in sync
         libraryViewModel = ViewModelProvider(this)[LibraryViewModel::class.java]
         libraryViewModel?.currentApiName?.observe(this) {
-            val icon = libraryViewModel?.currentSyncApi?.icon ?: R.drawable.library_icon
+            val syncAPI =  libraryViewModel?.currentSyncApi
+            Log.i("SYNC_API", syncAPI?.name.toString())
+            val icon = if (syncAPI?.name == "Local") {
```
fire-light42, line 1119: use idPrefix instead and use == localApi.idPrefix

fire-light42 (approved):

rockhero1234:
why everytime I have to choose for api, please make it stay after app restart

IndusAryan:
> why everytime I have to choose for api, please make it stay after app restart

i think your sync api resets to local when restarting, i have this bug in my debug version also, it resets to local

PR #1158, opened 2024-06-27
feat(ui): ability to play any local video from files using file chooser

IndusAryan:
i have not hidden the button in tv as of now, not sure if same intent will work seamless in tv as i dont have android tv, @KingLucius on weekend please test


app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
```diff
@@ -189,6 +200,24 @@ class DownloadFragment : Fragment() {
...
+            .addCategory(Intent.CATEGORY_OPENABLE)
+            .addFlags(FLAG_GRANT_READ_URI_PERMISSION) // Request temporary access
+        try {
+            videoResultLauncher.launch(
+                Intent.createChooser(
+                    intent,
+                    getString(R.string.open_local_video)
+                )
+            )
+        } catch (t: ActivityNotFoundException) {
```
fire-light42, line 216: is ActivityNotFoundException really the only exception?

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
```diff
@@ -189,6 +200,24 @@ class DownloadFragment : Fragment() {
...
+            t.printStackTrace()
```
fire-light42, line 217: logError ?

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
```diff
@@ -247,4 +276,33 @@ class DownloadFragment : Fragment() {
             binding?.downloadStreamButton?.extend()
         }
     }
+
+    // Open local video from files using content provider x safeFile
+    private val videoResultLauncher = registerForActivityResult(
```
fire-light42, line 281: This is very nested, try doing stuff like `if (result.resultCode != Activity.RESULT_OK) return` to remove unnecessary spacing

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
```diff
@@ -247,4 +276,33 @@ class DownloadFragment : Fragment() {
...
+    ) { result ->
+        if (result.resultCode == Activity.RESULT_OK) {
+            val intentData = result?.data?.data
+            intentData.let { selectedVideoUri ->
+                val name = SafeFile.fromUri(
+                    context ?: return@let ,
+                    selectedVideoUri ?: return@registerForActivityResult
+                )?.name()
+
+                activity?.navigate(
```
fire-light42, line 292: Use this instead
https://github.com/recloudstream/cloudstream/blob/55a0eb66cb5fa3833fd3a84f6cd231ece7e7edeb/app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadedPlayerActivity.kt#L50

fire-light42 (commented):
Good addition, however I am a bit unsure if fab is the correct UI for this or if the icon is correct. The guidelines also say that the secondary should be smaller https://m3.material.io/components/floating-action-button/guidelines

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
IndusAryan: in most cases the only exception will be the Files app not found in some shitty roms, so now it is in normal safe call to catch and log any error

IndusAryan:
i have used small fab now
[]()

if you are still uncertain abt this look, then the next player does it like this 


the 2 functions are in the helper object now, i am not sure if offlineplaybackhelper is good name or not, so as it's location in player dir unlike utils or appUtil.kt
i have tested the local video playback and its working fine

app/src/main/java/com/lagradost/cloudstream3/ui/player/OfflinePlaybackHelper.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.content.Context
+import android.net.Uri
+import androidx.core.content.ContextCompat.getString
+import com.lagradost.cloudstream3.CommonActivity.activity
+import com.lagradost.cloudstream3.R
+import com.lagradost.cloudstream3.utils.ExtractorUri
+import com.lagradost.cloudstream3.utils.UIHelper.navigate
+import com.lagradost.safefile.SafeFile
+
+object OfflinePlaybackHelper {
```
fire-light42, line 13: Both these use an implicit acitvity, make them pass in activity instead of context

fire-light42:
I want to merge this, however after testing I found that the file chooser button is still visible on TV layout. The problem is not the functionality but the navigation to that item. Navigating with a dpad to a floating button does not really make sense. Change the UI for TV or make it gone on tv.

IndusAryan:
> Change the UI for TV or make it gone on tv.
for now i have hidden it on the tv layout, work on buttons later, after i get some time

Moviellect:
A _local video_ selector UI that looks like **Next Player** is a good idea, it would be a great user experience.

KingLucius:
it's hidden on TV

PR #1163, opened 2024-07-01
Port large parts of the API to crossplatform

CranberrySoup:
This ports MainAPI at the cost of some changes:

- SyncAPI file renamed to SyncApi to prevent collisions
- Webview lost some constructors, not sure if it breaks some extensions
- ~Base64 encoding method now uses the kotlin encoding and it does not use ISO_8859_1~ The kotlin decoder uses ISO_8859_1 under the hood and the output is identical.
- AppUtils was split into AppUtils in the library and AppContextUtils in the app
- An init method to get context working properly in library

This is not thoroughly tested yet. 
Some extensions might break if relying on context methods in AppContextUtils (such as openWebView).
When merged it should be possible to run cloudstream extensions natively in the JVM.


CranberrySoup:
I could not find any issues with extensions when testing. This is ready for review.

fire-light42 (approved):
Looks good, however more testing is needed to ensure no breaking changes

Luna712:
@fire-light42 @CranberrySoup I can't seem to build or clone this repo locally anymore...

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt'
  'library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainApi.kt'
```

So only one file is cloning causing ton of issues missing the other.

PR #1164, opened 2024-07-01
Major performance and bug fixes to downloads

Luna712:
I tracked down underlying issues to setPersistentId. So I instead improved this using IO context, and this makes it use the view model more when in the downloads view, because otherwise when running savedData in an IO context it causes it to wait before updating progress which causes visible UI delays. This prevents that with downloads with the IsDone status.

This also has a UI change: If a download is complete, it no longer shows bytes / bytes, it only shows bytes once. This is both a UI and performance improvement to have to do less. Finally, this does some cleanup to DownloadAdapter and DownloadViewModel as well and fixes the display of the no downloads messages which previously seemed to have intended implementation but did nothing.

This also fixes the underlying cause of many bugs like the byte mismatch, while my previous fix for that was just a patch-job, this should fix the underlying cause of them as well.

Overall when testing this makes downloads load with almost no lag at all now.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
@@ -15,8 +15,8 @@ import com.lagradost.cloudstream3.databinding.DownloadHeaderEpisodeBinding
 import com.lagradost.cloudstream3.mvvm.logError
 import com.lagradost.cloudstream3.ui.download.button.DownloadStatusTell
 import com.lagradost.cloudstream3.utils.AppUtils.getNameFull
-import com.lagradost.cloudstream3.utils.DataStoreHelper
 import com.lagradost.cloudstream3.utils.DataStoreHelper.fixVisual
+import com.lagradost.cloudstream3.utils.DataStoreHelper.getViewPos
 import com.lagradost.cloudstream3.utils.UIHelper.setImage
 import com.lagradost.cloudstream3.utils.VideoDownloadHelper
 
@@ -116,7 +116,7 @@ class DownloadAdapter(
                 }
 
                 downloadHeaderTitle.text = d.name
-                val mbString = formatShortFileSize(itemView.context, card.totalBytes)
+                val formattedSizeString = formatShortFileSize(itemView.context, card.totalBytes)
 
                 if (card.child != null) {
                     downloadHeaderGotoChild.isVisible = false
@@ -131,7 +131,7 @@ class DownloadAdapter(
                         downloadButton.applyMetaData(card.child.id, card.currentBytes, card.totalBytes)
                         // We will let the view model handle this
                         downloadButton.doSetProgress = false
-                        downloadHeaderInfo.text = formatShortFileSize(downloadHeaderInfo.context, card.totalBytes)
+                        downloadHeaderInfo.text = formattedSizeString
                     } else downloadButton.doSetProgress = true
 
                     downloadButton.setDefaultClickListener(card.child, downloadHeaderInfo, mediaClickCallback)
@@ -152,7 +152,7 @@ class DownloadAdapter(
                                     R.plurals.episodes,
                                     card.totalDownloads
                                 ),
-                                mbString
+                                formattedSizeString
                             )
                     } catch (e: Exception) {
                         // You probably formatted incorrectly
@@ -173,7 +173,7 @@ class DownloadAdapter(
             val d = card.data
 
             binding.apply {
-                val posDur = DataStoreHelper.getViewPos(d.id)
+                val posDur = getViewPos(d.id)
                 downloadChildEpisodeProgress.apply {
                     isVisible = posDur != null
                     posDur?.let {
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
@@ -130,9 +130,9 @@ class DownloadAdapter(
                         downloadButton.setProgress(card.currentBytes, card.totalBytes)
                         downloadButton.applyMetaData(card.child.id, card.currentBytes, card.totalBytes)
                         // We will let the view model handle this
-                        downloadButton.setProgressText = false
+                        downloadButton.doSetProgress = false
                         downloadHeaderInfo.text = formatShortFileSize(downloadHeaderInfo.context, card.totalBytes)
-                    } else downloadButton.setProgressText = true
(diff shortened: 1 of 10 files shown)


Luna712:
I just realized this has some big issues with it... my bad...

Luna712:
There has got to be a way to make it work... but this way breaks in progress downloads indicator and download progress in EpisodeAdapter, etc... so is not the way to do it. But there must be a way to make this performant...

fire-light42:
> There has got to be a way to make it work... but this way breaks in progress downloads indicator and download progress in EpisodeAdapter, etc... so is not the way to do it. But there must be a way to make this performant...

Try doing it in a IO context then switch back to main thread as getDownloadFileInfoAndUpdateSettings checks the file on disk and doing that on the main thread lags it.

Luna712:
> > There has got to be a way to make it work... but this way breaks in progress downloads indicator and download progress in EpisodeAdapter, etc... so is not the way to do it. But there must be a way to make this performant...
> 
> Try doing it in a IO context then switch back to main thread as getDownloadFileInfoAndUpdateSettings checks the file on disk and doing that on the main thread lags it.

@fire-light42 

Done but a bit different as when just doing that it caused a delay where all icons showed the default download icon, for a noticeable time when opening downloads so I made it use the view model where possible. If you want me to do this another way I can certainly give it a try.

app/src/main/res/values/strings.xml
```diff
@@ -339,6 +339,10 @@
     <string name="livestreams">Livestreams</string>
     <string name="nsfw">NSFW</string>
     <string name="others">Others</string>
+    <plurals name="episodes">
```
fire-light42, line 342: <string name="episode">Episode</string>
    <string name="episodes">Episodes</string>

already exists, this will just make extra translation

Luna712: Ah I was aware it did but using getQuantityString made for cleaner code in the adapter. I can change it back though if you want?

fire-light42: Even if it is cleaner, it would be much better to do all plurals in a seperate PR and on all languages. Just doing it on english on a single item will be a real PITA to maintain and fix as you now have 2 translation keys on the same thing. 

Also before you do that PR, check that the translation site can handle plurals.

fire-light42 (changes requested):
Nope, this PR does not work in a recyclerview as views do not invalidate themself, presumably because of `if (progressSet) return`. 

I wont merge this PR before *ALL* recyclerview problems are solved *BOTH* in downloads and resultpage.

Luna712:
> Nope, this PR does not work in a recyclerview as views do not invalidate themself, presumably because of `if (progressSet) return`.
> 
> I wont merge this PR before _ALL_ recyclerview problems are solved _BOTH_ in downloads and resultpage.

This does work for me when testing though... in both places...

fire-light42:
> > Nope, this PR does not work in a recyclerview as views do not invalidate themself, presumably because of `if (progressSet) return`.
> > I wont merge this PR before _ALL_ recyclerview problems are solved _BOTH_ in downloads and resultpage.
> 
> This does work for me when testing though... in both places...

No. Try downloading an episode in a long running series and then flip between the seasons. I was able to reliably make it show it is currently downloading an episode (or several) on other seasons from 1 download. You MUST use something like `if (id == persistentId) return` instead.

fire-light42:
Properly using recycleviews without invalid data or memory leaks can be a real PITA to find, debug and resolve. Therefore it is extra important to verify that all UI is re-rendered on bind if it is a different item.

Luna712:
I can't seem to reproduce that but I do see how the issue could potentially happen. I will push a fix in a little bit.

app/src/main/res/values/strings.xml
Luna712: @fire-light42 what if I do:
```xml
<plurals name="episodes" translatable="false">
    <item quantity="one">@string/episode</item>
    <item quantity="other">@string/episodes</item>
</plurals>
 ```
 
 ? If not I can just revert that part no problem but if you are fine with this method I'd personally prefer it.

Comment (Luna712) [reply app/src/main/res/values/strings.xml]: I did this but will revert if you don't like it this way.

**Comment by Luna712** [comment]
@fire-light42 I *think* the issue you had *may* be fixed, but I still could not reproduce to confirm

**Review by fire-light42 — CHANGES_REQUESTED** [review]
The code works and looks good, however an issue I did not find in my last review snuck into b06d9f224dc82ae8b914237dbc6540492922eb5b (#1145) and I want to you to fix that in this pr. 

The issue is as follows: 
1. Go into a resultpage with multiple seasons
2. Download any episode
3. Force quit
4. Reenter the app and all download will automatically resume
5. Go into the same series
6. Pause that download
7. Switch to another season
8. Switch back to the season

If you do this correctly then you will see that the progress on the paused download does not match with the progress downloaded and it will need to resume to update to the correct progress. 

https://github.com/recloudstream/cloudstream/assets/147925818/a6db76d8-9b68-4490-88f9-0ee8daba8cea

**Review by fire-light42 — APPROVED** [review]
Ye, the issue I found is not related to this or your last pr but might need an real https://github.com/recloudstream/cloudstream/issues issue.

File: app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt [hunk 2]
```diff
@@ -93,110 +94,128 @@ class DownloadAdapter(
...
+                        logError(e)
                     }
 
-                    downloadButton.setDefaultClickListener(card.data, downloadChildEpisodeTextExtra, mediaClickCallback)
-
-                    downloadChildEpisodeText.apply {
-                        text = context.getNameFull(d.name, d.episode, d.season)
-                        isSelected = true // Needed for text repeating
+                    episodeHolder.setOnClickListener {
+                        clickCallback.invoke(DownloadHeaderClickEvent(0, d))
```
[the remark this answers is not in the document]
Luna712: Done

Luna712:
> Ye, the issue I found is not related to this or your last pr but might need an real https://github.com/recloudstream/cloudstream/issues issue.

@fire-light42 

That's a good catch. I will try and fix that in a follow-up as well. But I'd rather not further complicate things in this PR and I'm not sure how to fix it at the moment.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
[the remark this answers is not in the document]
Luna712: @fire-light42 I actually remembered why these didn't use constants, because they didn't do the same thing so it made no sense. So I instead added new constants that are descriptive for what DownloadHeaderClickEvent actually does.

Issue #1167, opened 2024-07-02
Syncing a show status with AniList crashes app


navkaran0305:
### Steps to reproduce

1. Sign into the AniList account in the Cloudstream app.
2. Go to any Anime which is recognized by AniList.
3. Swipe from the left to open the show status window.
4. Change the status of the show (For example: Watching -> Plan to Watch)
5. Click the 'Sync' button at the bottom which caues the app to crash and the sync to not complete.

![31335a58-a292-41fc-8ce9-70e409c6ad8b](https://github.com/recloudstream/cloudstream/assets/132475876/435fb5c6-a207-44fb-9b5f-80d739dce97c)

### Expected behavior

The app should not crash and the changed show status should sync successfully

### Actual behavior

The app crashes and goes into safe mode. The syncing process is unsuccessful.

![025cfaa6-e552-4f27-a693-fffa695bf2f4](https://github.com/recloudstream/cloudstream/assets/132475876/82c90404-e56d-4a53-80ed-b8a1c22c4803)

### Cloudstream version and commit hash

Version - 4.3.2-PRE  Commit Hash - a5582a7

### Android version

Android 14

### Logcat

```java
--------- beginning of crash
07-01 22:51:20.460  4965  5086 E AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
07-01 22:51:20.460  4965  5086 E AndroidRuntime: Process: com.lagradost.cloudstream3.prerelease, PID: 4965
07-01 22:51:20.460  4965  5086 E AndroidRuntime: com.lagradost.cloudstream3.mvvm.DebugException: THIS IS A DEBUG EXCEPTION!
07-01 22:51:20.460  4965  5086 E AndroidRuntime: No Simkl token!
07-01 22:51:20.460  4965  5086 E AndroidRuntime: 	at com.lagradost.cloudstream3.syncproviders.providers.SimklApi.getToken(SimklApi.kt:1142)
07-01 22:51:20.460  4965  5086 E AndroidRuntime: 	at com.lagradost.cloudstream3.syncproviders.providers.SimklApi.access$getToken(SimklApi.kt:45)
07-01 22:51:20.460  4965  5086 E AndroidRuntime: 	at com.lagradost.cloudstream3.syncproviders.providers.SimklApi$HeaderInterceptor.intercept(SimklApi.kt:776)
[...]

PR #1168, opened 2024-07-03
Fix debug exceptions in releases

CranberrySoup:
Fixes https://github.com/recloudstream/cloudstream/issues/1167

Tested and works, although I would have liked a better build flavor detection.
`arg.contains("debug", true)` could be swapped with more precise checks, but that would be more likely to break if the flavor name changes.


fire-light42:
Holy hell, gradle moment 💀

PR #1169, opened 2024-07-03
chore(locales): fix locale issues

keyiflerolsun:


C10udburst:
Please consider applying your changes on [our weblate](https://hosted.weblate.org/engage/cloudstream/) instead of here as that allows us to merge many small changes like these into a single commit as well as prevent build issues (like one that occured here). Merging translation PRs from Github also sometimes causes merge conflicts when the big weblate PR gets merged.

PR #1171, opened 2024-07-03
Fix an IllegalStateException crash

Luna712:
When you play something, then go home on the device, and come back to something that should run popCurrentPage(), IE `noLinksFound()`, it often causes the app to crash into safe mode with IllegalStateException.

This may not be the best solution here, but testing seems to indicate it does work to fix this.

I thought about using `supportFragmentManager.popBackStack()`, but that seems to cause a bigger delay in doing anything, even if we don't have any issues here. Also `onBackPressedDispatcher.onBackPressed()` can be more in line with how Android handles back press actions, respecting the back stack and fragment lifecycle more conveniently so went with this approach.


fire-light42:
I have no idea if this fixes the issue as it is like 1 frame timing, however I will trust you on this.

PR #1174, opened 2024-07-04
Extractor: added Pichive & Sobreatsesuyp

keyiflerolsun:
image
image


Luna712:
@keyiflerolsun Just a heads up, this has a merge conflict due to extractor paths changing to https://github.com/recloudstream/cloudstream/tree/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors

PR #1175, opened 2024-07-04
Fix download button display bug in adapter

Luna712:
I had a patch-fix for this before so it does not really happen anymore but this should fix the actual issue behind it also, making it more reliable than relying on `setItemViewCacheSize`


fire-light42:
Nice catch

PR #1182, opened 2024-07-05
Move setting downloadStorageAppbar visibility to downloadBytes

Luna712:
I think this was always intended. If not it should just be removed completely. If you have the app itself installed you are using storage on your device (which is what usedBytes is), I think the intended purpose was used bytes within the app (which is what downloadBytes is)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
@@ -101,10 +101,10 @@ class DownloadFragment : Fragment() {
         }
         observe(downloadsViewModel.usedBytes) {
             updateStorageInfo(view.context, it, R.string.used_storage, binding?.downloadUsedTxt, binding?.downloadUsed)
-            binding?.downloadStorageAppbar?.isVisible = it > 0
         }
         observe(downloadsViewModel.downloadBytes) {
             updateStorageInfo(view.context, it, R.string.app_storage, binding?.downloadAppTxt, binding?.downloadApp)
+            binding?.downloadStorageAppbar?.isVisible = it > 0
         }
 
         val adapter = DownloadAdapter(


fire-light42:
No, I think it was like this before but was changed to this to not make the UI disappear.

Luna712:
> No, I think it was like this before but was changed to this to not make the UI disappear.

@fire-light42 

I'm fine with that but I think the check should just be removed then as it really can never be true as if the app itself is installed it is using storage so will never be 0 or less, or could it?

fire-light42:
> > No, I think it was like this before but was changed to this to not make the UI disappear.
> 
> @fire-light42
> 
> I'm fine with that but I think the check should just be removed then as it really can never be true as if the app itself is installed it is using storage so will never be 0 or less, or could it?

I think the check is not for less then zero, but instead for fucked storage output. aka if some storage permission is not allowed.

Luna712:
> I think the check is not for less then zero, but instead for fucked storage output. aka if some storage permission is not allowed.

@fire-light42 

ohhh I understand makes sense but maybe would be good to make more clear, also not exactly sure the benefits of that... what does it actually hide in that case? Anyway thanks for the clarification!

fire-light42:
> ohhh I understand makes sense but maybe would be good to make more clear, also not exactly sure the benefits of that... what does it actually hide in that case? Anyway thanks for the clarification!

We dont want to display nonsensical data. This data is accessible even with no storage perms, but you cant trust storage in any way on android. So in case of a critical error in storage it should be invisible.

Luna712:
> > ohhh I understand makes sense but maybe would be good to make more clear, also not exactly sure the benefits of that... what does it actually hide in that case? Anyway thanks for the clarification!
> 
> We dont want to display nonsensical data. This data is accessible even with no storage perms, but you cant trust storage in any way on android. So in case of a critical error in storage it should be invisible.

@fire-light42 

That makes total sense... funnily enough I actually realized that exact thing a little while ago as some debug testing on my other multi-delete PR I messed something up causing the exact circumstance you mention here so it makes total sense. Thanks!

PR #1183, opened 2024-07-05
Faster remove downloads from adapter after deletion

Luna712:
I'm not sure if you'll want this or not. But I was going to push this in my last performance PR but decided to do it separately so it could be reviewed separately. If you don't want it this can be closed no worries.


Luna712:
Sorry I messed up fixing a merge conflict since I did this a few days ago. I wilm fix it.

Luna712:
I would rather not to do this right now I may do something else later

PR #1197, opened 2024-07-12
Update Vidplay.kt

AvocadoWasSeized:
Due to KillerDogeEmpire wanting to not have anything to do with CS3 anymore, he will delete the vidplay keys repo so I have updated it with the MultiKeys repo


RowdyRushya:
Sorry @AvocadoWasSeized, but can you take down this PR as i created another one where I fixed multiple issues and it also has static method for keys fetch.
And I think this PR will not work well as the new html page is providing multiple keys and you have not written data class to extract the Vidplay key from it.

AvocadoWasSeized:
@RowdyRushya I see, 

PR closed in favor of #1198

PR #1202, opened 2024-07-17
updating vidplay encryption method

RowdyRushya:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidplay.kt
```diff
@@ -1,13 +1,16 @@
 package com.lagradost.cloudstream3.extractors
 
+import android.util.Base64
```
fire-light42, line 3: This will fail, use the built in base64 helper as that works without android.

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidplay.kt
RowdyRushya: inbuilt Base64Encode is working well but Base64Decode is having some issue so I switched to Kotlin.

fire-light42 (approved):

PR #1204, opened 2024-07-20
Exclude download keys from backups

Luna712:
I am not certain if this is correct to do or not so if not please just close. Thanks!


fire-light42:
No. We dont want this, because you might want to migrate from stable to pre, or simply reinstall. There is no reason to exclude this from backups.

Luna712:
> No. We dont want this, because you might want to migrate from stable to pre, or simply reinstall. There is no reason to exclude this from backups.

@fire-light42 

Fair enough, but the reason I did this is because we do exclude DOWNLOAD_EPISODE_CACHE but not DOWNLOAD_HEADER_CACHE, doesn't that mean restoring can become inconsistent? Should episodes be removed from excluding then also? Also wouldn't restoring still not properly work if you use something other than the default download path (since permissions may be wrong, we exclude that key also), and actually could rather mess it up a bit?

fire-light42:
Git blame blames you for that, and it slipped through review. https://github.com/recloudstream/cloudstream/pull/979

Luna712:
> Git blame blames you for that, and it slipped through review. #979

Oh that was me because it was applied in another PR (the cloud backup one) and I just applied the fix from there. Whoops

PR #1207, opened 2024-07-22
feat(UI): Show Episode Runtime

KingLucius:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1710,7 +1710,25 @@ data class Episode(
     var rating: Int? = null,
     var description: String? = null,
     var date: Long? = null,
-)
+    var runTime: Int? = null,
```
fire-light42, line 1713: Add a comment about it being in seconds or suffix with Sec

app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
```diff
@@ -292,6 +297,12 @@ class EpisodeAdapter(
                     episodeDate.isVisible = false
                 }
 
+                episodeRuntime.setText(
+                    txt(
+                        card.runTime?.times(60L)?.let { secondsToReadable(it.toInt(), "0 mins") }
```
fire-light42, line 302: 0 mins? It should say 1 min or nothing at all.

KingLucius: Any chance that this one will every be triggered?

fire-light42 (approved):
This can be merged anytime, but MainAPI changes might need to be postponed until stable release.

KingLucius:
> This can be merged anytime, but MainAPI changes might need to be postponed until stable release.

There is a backwards compatibility constructor for that

Luna712:
> > This can be merged anytime, but MainAPI changes might need to be postponed until stable release.
> 
> There is a backwards compatibility constructor for that

@KingLucius 

AFAIK that only works for some extensions when there is a stable release as has happened with my last content rating PR which also had a backwards compatible constructor.

Luna712:
Otherwise extensions become incompatible on stable

KingLucius:
> > > This can be merged anytime, but MainAPI changes might need to be postponed until stable release.
> > 
> > There is a backwards compatibility constructor for that
> 
> @KingLucius 
> 
> AFAIK that only works for some extensions when there is a stable release as has happened with my last content rating PR which also had a backwards compatible constructor.

That needs more testing for check what is the issue
Anyway we are not in hurry, we can wait no issue

PR #1208, opened 2024-07-22
Add system dark theme

### LINKED ISSUE #341: automatic dark/light mode switch on android 11 and lower

ik theres monet support but theres no system auto dark and light support for older android versions

### Other details

_No response_

### Acknowledgements

epireyn:
This adds an option to set the theme accordingly to the system's. Closes #341.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
@@ -88,7 +88,7 @@ class SettingsUI : PreferenceFragmentCompat() {
         getPref(R.string.app_theme_key)?.setOnPreferenceClickListener {
             val prefNames = resources.getStringArray(R.array.themes_names).toMutableList()
             val prefValues = resources.getStringArray(R.array.themes_names_values).toMutableList()
-            val removeFunction = { text: String ->
+            val removeIncompatible = { text: String ->
                 val toRemove = prefValues
                     .mapIndexed { idx, s -> if (s.startsWith(text)) idx else null }
                     .filterNotNull()
@@ -100,10 +100,10 @@ class SettingsUI : PreferenceFragmentCompat() {
                 }
             }
             if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) { // remove monet on android 11 and less
-                removeFunction("Monet")
+                removeIncompatible("Monet")
             }
             if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) { // Remove system on android 9 and less
-                removeFunction("System")
+                removeIncompatible("System")
             }
 
             val currentLayout =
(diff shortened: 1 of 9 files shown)


app/src/main/res/values-pl/array.xml
```diff
@@ -252,13 +252,15 @@
 
 
     <string-array name="themes_names">
+        <item>System</item>
         <item>Ciemny</item>
         <item>Szary</item>
         <item>Amoled</item>
         <item>Flashbang</item>
-        <item>Material You</item>
+        <item>Material You (Android 12+)</item>
```
fire-light42, line 260: Why "(Android 12+)" that is simply redundant.

fire-light42 (changes requested):

epireyn:
Indeed, I noticed afterwards but forgot to revert it.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -88,10 +88,9 @@ class SettingsUI : PreferenceFragmentCompat() {
         getPref(R.string.app_theme_key)?.setOnPreferenceClickListener {
             val prefNames = resources.getStringArray(R.array.themes_names).toMutableList()
             val prefValues = resources.getStringArray(R.array.themes_names_values).toMutableList()
-
-            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) { // remove monet on android 11 and less
+            val removeFunction = { text: String ->
```
C10udburst, line 91: maybe removeTheme would be more descriptive?

fire-light42 (changes requested):
This makes system the default because of `settingsManager.getString(getString(R.string.primary_color_key), prefValues.first())`

"Dark" should be the default, so I need you to change it.

Luna712:
> This makes system the default because of `settingsManager.getString(getString(R.string.primary_color_key), prefValues.first())`
> 
> "Dark" should be the default, so I need you to change it.

@fire-light42

If the intent is to always default to dark, wouldn't it make sense to instead force that to Dark as a default rather than `.first()` it seems that would be less prone to potential changes messing it up. Anyway it is up to you, that is just my personal opinion but there may also be a reason why not to do that. But doing it also just seems like it would better indicate the intended default as well.

epireyn:
> > This makes system the default because of `settingsManager.getString(getString(R.string.primary_color_key), prefValues.first())`
> > "Dark" should be the default, so I need you to change it.
> 
> @fire-light42
> 
> If the intent is to always default to dark, wouldn't it make sense to instead force that to Dark as a default rather than `.first()` it seems that would be less prone to potential changes messing it up. Anyway it is up to you, that is just my personal opinion but there may also be a reason why not to do that. But doing it also just seems like it would better indicate the intended default as well.

It would be the best, however it requires hard-coding the value. If it was to change, the themes wouldn't work at all. I reorder the list, it's no big issue.

app/src/main/res/values-tr/array.xml
```diff
@@ -277,13 +277,15 @@
...
     <string-array name="themes_names">
+        <item>Sistem</item>
         <item>Koyu</item>
         <item>Gri</item>
         <item>Amoled</item>
         <item>Flaş Bombası</item>
         <item>Material You</item>
     </string-array>
     <string-array name="themes_names_values">
+        <item>System</item>
```
fire-light42, line 288: You forgot to reorder all arrays

fire-light42 (changes requested):

fire-light42 (approved):

fire-light42:
Now more people can be flashbanged by the white theme 🤩

PR #1210, opened 2024-07-23
Add the option to hide video controls

### LINKED ISSUE #1190: Video Controls With Their Name

The controls appearing at the bottom of the video player are scrolling due to being more controls. This is affecting both the UI and UX negatively.

Screenshot_2024-07-08-08-10-45-93

I suggest adding a toggle button in [Settings > Player > Player features] labeled "**Show control's name**" to address this issue.

### Other details

_No response_

### Acknowledgements

- [X] My suggestion is **NOT** about adding a new provider

epireyn:
Closes #1190


app/src/main/res/values/strings.xml
```diff
@@ -797,4 +797,6 @@
     <string name="device_pin_error_message">Can\'t get the device PIN code, try local authentication</string>
     <string name="device_pin_expired_message">PIN code is now expired !</string>
     <string name="device_pin_counter_text">Code expires in %1$dm %2$ds</string>
+    <string name="hide_player_control_names_key">hide_player_control_names_key</string>
```
C10udburst, line 800:
```suggestion
    <string name="hide_player_control_names_key" translatable="false">hide_player_control_names_key</string>
```

Same for other languages.

KingLucius:
1. This one is not needed on TV as there is a lot of space under the progress bar, so I think we should hide this on TV
2. Why not you use a list of buttons ID and map thru them, to avoid bugs in the future if we added another element with different type

epireyn:
To answer the second point, it only removes Material Buttons and iterates on linear layouts. Any other element is ignored. It makes it easier to add other buttons without the need to maintain a comprehensive list.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1591,6 +1600,19 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         }
     }
 
+    private fun PlayerCustomLayoutBinding.hideControlsNames() {
+        fun iterate(layout: LinearLayout) {
+            layout.children.forEach {
+                if (it is MaterialButton) {
+                    it.textSize = 0f
```
fire-light42, line 1607: This leaves the button non centered and with weird padding. Do at least this.
```kotlin
it.textSize = 0f
it.iconPadding = 0
it.iconGravity = MaterialButton.ICON_GRAVITY_TEXT_START
it.setPadding(0,0,0,0)
```

fire-light42 (changes requested):

LagradOst (approved):

PR #1216, opened 2024-07-26
metaproviders: use new* methods

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -978,7 +978,7 @@ data class TvSeriesSearchResponse(
     override var type: TvType? = null,
 
     override var posterUrl: String? = null,
-    val year: Int? = null,
+    var year: Int? = null,
     val episodes: Int? = null,
```
fire-light42, line 982: If you change year, then also change episodes to a var.

fire-light42 (commented):
I will leave this to be merged before stable as it changes MainAPI. You can also change some other vals to vars like LiveSearchResponse::lang to reduce the number of times of breaking MainAPI changes.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: I can if you want but I am kinda confused why?
Year needs to be so it can be changed when using newTVSeriesSearchResponse but with that you can pass episodes.

Luna712: Nevermind I just realized I was looking at newTVSeriesLoadResponse... I will change this. Sorry

Luna712:
> I will leave this to be merged before stable as it changes MainAPI. You can also change some other vals to vars like LiveSearchResponse::lang to reduce the number of times of breaking MainAPI changes.

Would changing a val to a var really break MainAPI at all though? I thought that was kinda interchangeable when using val to var just not var to val?

fire-light42:
> > I will leave this to be merged before stable as it changes MainAPI. You can also change some other vals to vars like LiveSearchResponse::lang to reduce the number of times of breaking MainAPI changes.
> 
> Would changing a val to a var really break MainAPI at all though? I thought that was kinda interchangeable when using val to var just not var to val?

It will break extentions on stable that are using prerelease jar as target. To explain why, look at how var generates a set and get function, while val only generates a getter. Any prerelease code running on stable will call the setter and generate an NoSuchMethodError. Look at https://godbolt.org/z/8r5oW98n6

Luna712:
> > > I will leave this to be merged before stable as it changes MainAPI. You can also change some other vals to vars like LiveSearchResponse::lang to reduce the number of times of breaking MainAPI changes.
> > 
> > 
> > Would changing a val to a var really break MainAPI at all though? I thought that was kinda interchangeable when using val to var just not var to val?
> 
> It will break extentions on stable that are using prerelease jar as target. To explain why, look at how var generates a set and get function, while val only generates a getter. Any prerelease code running on stable will call the setter and generate an NoSuchMethodError. Look at https://godbolt.org/z/8r5oW98n6

Thank you that is very good to know.

LagradOst:
I have waited for this pr for a long time now.

LagradOst:
🙏

PR #1218, opened 2024-07-28
Introduce version catalog for dependency management

CodeWithTamim:
Adds a libs.versions.toml file to centralize dependency management. This improves dependency consistency, reduces maintenance overhead, and simplifies version updates across modules.


fire-light42:
Nice pr, however I have some problems building it. I will merge this after I have figured out if it is an android studio issue or something else.

CodeWithTamim:
😭

PR #1219, opened 2024-07-28
fixed: Test Search & VidMoxy, RapidVid extractors

keyiflerolsun:
we can manipulate the title like this ;

image


keyiflerolsun:
this is good for me;

image

PR #1220, opened 2024-07-28
Fix for navigation UI bug

### LINKED ISSUE #26: Bottom Nav Bar UI buggy in Settings
- Settings submenus don't update the nav bar to highlight settings, it stays in the previous state. Buggy when user switches from another screen like Home to a Settings submenu.
- Also when you are on a settings sub screen and tap settings in bottom bar, its expected to take you to settings page but it doesn't.

Proposed easy fix: Hide bottom nav bar when entering a settings submenu

cc: @LagradOst , @Blatzar

Luna712:
Fixes #26


fire-light42:
Nice one 👍

PR #1221, opened 2024-07-28
Fix cast recycler scrolling

### LINKED ISSUE #563: Caste recycler works only once
### Steps to reproduce

open app and swipe casts on movie or a series
it will swipe
now go to other movie or series it will open recommendations fragment.

### Expected behavior

caste should work everytime 

### Actual behavior

it doesn't work always,probably users needs to belong to a upper caste :) :harold:

Latest PreRelease

### Android version

A13

### Logcat

_No response_

### Other details

_No response_

### Acknowledgements

Luna712:
Fixes #563


fire-light42:
This does not fix it completely, as it was still stuck after going to a recommendation, then pressing `<-`.

Luna712:
> This does not fix it completely, as it was still stuck after going to a recommendation, then pressing `<-`.

Ah right will fix soon thanks!

Luna712:
@fire-light42 I fixed it as best I could. I think it works mostly but sometimes if you open recommendations and close it it takes a few seconds for cast recycler to work again or if you scroll up or down sometimes. Not 100% reliable but a lot better then it was (I think anyway)

fire-light42:
> @fire-light42 I fixed it as best I could. I think it works mostly but sometimes if you open recommendations and close it it takes a few seconds for cast recycler to work again or if you scroll up or down sometimes. Not 100% reliable but a lot better then it was (I think anyway)

If it is not 100% reliable, then you need to comment that in the code.

Luna712:
> > @fire-light42 I fixed it as best I could. I think it works mostly but sometimes if you open recommendations and close it it takes a few seconds for cast recycler to work again or if you scroll up or down sometimes. Not 100% reliable but a lot better then it was (I think anyway)
> 
> If it is not 100% reliable, then you need to comment that in the code.

@fire-light42 

Thanks, I added a comment for that now.

fire-light42:
It is indeed very jank, but I guess it is better than what we have at the moment. Discord storing ids instead of views 😔

PR #1226, opened 2024-07-30
Fix Trailers on API<33

KingLucius:
Recent NewPipeExtractor updates pushed minimum sdk to 33 which needs desugar_jdk_libs_nio flavor to support below 33


fire-light42:
Nice!

PR #1228, opened 2024-07-30
Add support for Next Episode in downloads

### LINKED ISSUE #1023: ADD NEXT EPISODE BUTTON FOR DOWNLOADED FILE

There Is No Way To Play next ep in cs3 when playing download file, we need to go back to download list and play from there also the player itself didn't play next ep when current ep is being completed

add a button for in the video player for download file that let us play next ep or previous ep from that player

### Other details

_No response_

### Acknowledgements

Luna712:
closes #1023

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadFileGenerator.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadFileGenerator.kt
@@ -1,5 +1,6 @@
 package com.lagradost.cloudstream3.ui.player
 
+import android.net.Uri
 import com.lagradost.cloudstream3.AcraApplication.Companion.context
 import com.lagradost.cloudstream3.CommonActivity.activity
 import com.lagradost.cloudstream3.R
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadFileGenerator.kt
```diff
@@ -59,7 +62,19 @@ class DownloadFileGenerator(
         offset: Int
     ): Boolean {
         val meta = episodes[currentIndex + offset]
-        callback(null to meta)
+
```
fire-light42, line 65: Add `If (meta.uri == Uri.EMPTY)`, as we can skip this step if uri is supplied.

fire-light42 (commented):
I have not tested this yet, it is looks good.

app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadFileGenerator.kt
Luna712: Done

PR #1237, opened 2024-07-31
Helper Added: CryptoJS

keyiflerolsun:


fire-light42 (changes requested):
This pr looks good, however I would like 
1. Do not use ***android***.util.Base64 as that will fail the lib jar, as that is cross platform and not only for android. 
1. Add the test from https://gist.github.com/thackerronak/554c985c3001b16810af5fc0eb5c358f
2. EvpKDF->evpkdf

fire-light42 (approved):

PR #1238, opened 2024-07-31
Add support for progress on header downloads

Luna712:


fire-light42 (approved):
Looks good, will merge after testing 👍

PR #1239, opened 2024-07-31
Fix TV focus issues for add repo input

Luna712:


app/src/main/res/layout/add_repo_input.xml
```diff
@@ -83,11 +81,8 @@
             android:layout_height="wrap_content"
             android:hint="@string/repository_url_hint"
             android:inputType="textUri"
-            android:nextFocusLeft="@id/apply_btt"
```
fire-light42, line 86: Why did you remove left and right?

Luna712: Because left and right is unexpected behavior when there is nothing to the left or right I don't expect it to jump to the bottom buttons if I accidentally press left or right, which is what it does. If you want it readded I can I just thought of it as behavior you don't expect.

fire-light42: CloudStream 3 uses that behavior in a lot of places to make it easier to navigate. So no, I dont think that is unexpected behavior.

https://github.com/search?q=repo%3Arecloudstream%2Fcloudstream%20nextFocusLeft%3D%22%40id%2Fapply_btt&type=code

Luna712: Readded

PR #1240, opened 2024-08-01
Fix jitpack build

CranberrySoup:
This makes `./gradlew build` be successful on a clean clone. 
Fixing this makes the library buildable on jitpack and thereby possible to be used as a dependency.


fire-light42 (approved):
👀

PR #1252, opened 2024-08-04
feat(TV UI): Navigation rail redesign

KingLucius:
- New icons (Home & Settings).
- Filled icons on focus & selecting (Home, Library & Settings).
- Account selection at bottom of the NavigrationRail (TV Only).

This PR is part of large change (notification center on TV) but decided to push this to make the review easier so you will find some comments (notification badge on updates) and a header on NavigationRail not used.

image


fire-light42 (changes requested):
`accounts` may be `arrayOf<Account>()` aka emptyList in case the user has never used the account feature. This makes the account profile pic empty when it should use the default account profile pic.

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1658,7 +1670,27 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
...
+
+            val navProfileRoot = findViewById<LinearLayout>(R.id.nav_footer_root)
+            val navProfilePic = findViewById<ImageView>(R.id.nav_footer_profile_pic)
+            val navProfileCard = findViewById<CardView>(R.id.nav_footer_profile_card)
+
+            navProfileCard?.setOnClickListener {
+                showAccountSelectLinear()
+            }
+
+            val profileImage = DataStoreHelper.getCurrentAccount()?.image ?: DataStoreHelper.getDefaultAccount(context).image
```
fire-light42, line 1683: No, why are you doing this here when you have the observe. Remove the homeViewModel.currentAccount shit and do everything here or use observe properly. I would recommend doing it here as accounts are not tied to the home view.

KingLucius: I am using the observe to update after the account management dialog and here to update on start up

KingLucius: You mean I should observe here to be able to update across all instead of home?

fire-light42: Observe updates instantly if you have set the value beforehand. Right now I am just worried that this code can lead to more bugs like this, as the account you observe does not include the default account.

fire-light42: So stick to one approach and not both, either make observe the only thing that changes the UI or only change UI with the callback event.

fire-light42 (changes requested):
I want `So stick to one approach and not both, either make observe the only thing that changes the UI or only change UI with the callback event.` to be fixed before merge.

KingLucius:
> I want `So stick to one approach and not both, either make observe the only thing that changes the UI or only change UI with the callback event.` to be fixed before merge.

I partly finished this, testing it before pushing

fire-light42 (approved):
Code looks good, will merge after testing 👍

PR #1256, opened 2024-08-06
feat(ui): settings for thumbnail on seekbar

int3debug:
jw8ljzopvab6kgvj_1


app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -482,11 +485,13 @@ abstract class AbstractPlayerFragment(
         )
 
         if (player is CS3IPlayer) {
+            val ctx = context ?: return
             // preview bar
             val progressBar: PreviewTimeBar? = playerView?.findViewById(R.id.exo_progress)
             val previewImageView: ImageView? = playerView?.findViewById(R.id.previewImageView)
             val previewFrameLayout: FrameLayout? = playerView?.findViewById(R.id.previewFrameLayout)
-            if (progressBar != null && previewImageView != null && previewFrameLayout != null) {
+            val previewSeekThumb = PreferenceManager.getDefaultSharedPreferences(ctx)?.getBoolean(ctx.getString(R.string.preview_seekbar_key),false) == true
```
fire-light42, line 493: This is somewhat incorrect, as what you are doing is simply hiding the UI. It will still be loaded in the background.

app/src/main/res/values/strings.xml
```diff
@@ -811,4 +811,7 @@
     <string name="sort_release_date_old">Release Date (Old to New)</string>
     <string name="hide_player_control_names_key" translatable="false">hide_player_control_names_key</string>
     <string name="hide_player_control_names">Hide names of the player\'s controls</string>
+    <string name="preview_seekbar">Seekbar preview</string>
+    <string name="preview_seekbar_desc">Enable preview thumbnail on seekbar</string>
+    <string name="preview_seekbar_key">false</string>
```
fire-light42, line 816: `preview_seekbar_key` should be `preview_seekbar_key`, not `false`

fire-light42: Also `translatable="false"`

app/src/main/res/xml/settings_player.xml
```diff
@@ -99,11 +99,11 @@
             android:summary="@string/rotate_video_desc"
             app:key="@string/rotate_video_key" />
         <SwitchPreference
```
fire-light42, line 101: You are removing an existing setting `screen_rotation`, readd it.

app/src/main/res/xml/settings_player.xml
```diff
@@ -99,11 +99,11 @@
...
-            android:icon="@drawable/screen_rotation"
-            android:title="@string/auto_rotate_video"
+            android:icon="@drawable/preview_seekbar_icon"
+            android:title="@string/preview_seekbar"
             app:defaultValue="false"
```
fire-light42, line 104: app:defaultValue should be true

app/src/debug/res/drawable/preview_seekbar_icon.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
fire-light42, line 1: preview_seekbar_icon.xml is in the wrong location and duplicated twice, it should be located in app/src/main.

fire-light42 (changes requested):
Nice PR 🥳 

I would want you to move your disable logic into the IPreviewGenerator::new function, as that will make sure it wont load any data if it is disabled.

```kotlin
        fun new(): IPreviewGenerator {
            val userDisabled = AcraApplication.context?.let { ctx ->
                PreferenceManager.getDefaultSharedPreferences(ctx)?.getBoolean(
                    ctx.getString(R.string.preview_seekbar_key),
                    true
                ) == false
            } ?: false
            /** because TV has low ram + not show we disable this for now */
            return if (isLayout(TV) || userDisabled) {
                empty()
            } else {
                PreviewGenerator()
            }
        }
```

I would also recommend you to hide it on TV, as it is not supported on TV.
```kotlin
getPref(R.string.preview_seekbar_key)?.hideOn(TV)
```

fire-light42 (approved):
Thank you for your PR. Dont worry about the quality as there is no such thing as a bad PR, and the review process normally takes a few rounds to complete.

PR #1258, opened 2024-08-08
Fix BackPressedCallback conflict between activities

Luna712:
There is another, more simple way to do this by just detatching the callback everytime it is pressed in DownloadedPlayerActivity, however, I figured this method was more future-proofed if more activities are added later so went with this, but if you want me to go to do the more simple method I can also.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadChildFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadChildFragment.kt
@@ -38,7 +38,7 @@ class DownloadChildFragment : Fragment() {
     }
 
     override fun onDestroyView() {
-        detachBackPressedCallback()
+        activity?.detachBackPressedCallback()
         binding = null
         super.onDestroyView()
     }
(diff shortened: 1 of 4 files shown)


fire-light42:
Will be merged after testing. 👍

fire-light42 (changes requested):
This breaks the current attachBackPressedCallback implementation by removing old callbacks. You can test this by pressing back on the mainpage on tv, then you will get a "do you want to exit dialog", however if you go into downloads then back this dialog wont appear.

Luna712:
> This breaks the current attachBackPressedCallback implementation by removing old callbacks. You can test this by pressing back on the mainpage on tv, then you will get a "do you want to exit dialog", however if you go into downloads then back this dialog wont appear.

@fire-light42 

This happens on current pre release also, I can fix it in this PR but it is unrelated to the purpose of this PR, which in my opinion fixes a much more annoying and consistent issue where back presses can stop working completely and you have force stop the app to fix. So I think it would make more sense to fix that in a follow-up. But it's up to you.

fire-light42 (approved):

PR #1264, opened 2024-08-10
Update StreamWishExtractor

b4byhuey:


fire-light42:
Not tested, but looks good to me. 👍

PR #1275, opened 2024-08-13
Update media dependency

CranberrySoup:
This bumps media to the latest version while retaining features such as sync subs.
I removed the 1us seek to forcefully refresh subtitles as it could cause 'infinite buffering' if the wrong decoder was selected.

The sync subs should be more future proof now, by not relying on any copied code.

The new parsing should also make it possible to 'view' upcoming subtitles to aid in subtitle syncing in the future.


fire-light42 (commented):
The code looks good, will make another review after testing 👍

PR #1279, opened 2024-08-16
Updated VidhideExtractor

misajimenezmx:


fire-light42:
Not tested, but looks good.

phisher98:
@fire-light42 I believe This PRbroke Vidhide Extractor so I have to create a inbuilt Extractor working before this PR and Pushed in Multimovies

misajimenezmx:
Added new pull request avoiding broke vidhide extractor, including VidHidePro extractor that needs another way to get the urls

https://github.com/recloudstream/cloudstream/pull/1286

PR #1281, opened 2024-08-16
ci: add dependabot

indrastorms:
Need to configure dependabot from repo security settings

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    labels: []
+    directory: /
+    target-branch: master
+    schedule:
+      interval: monthly
+
+  - package-ecosystem: npm
+    labels: []
+    directory: /
+    target-branch: master
+    schedule:
+      interval: monthly
+
+  - package-ecosystem: gradle
+    labels: []
+    directory: /
+    target-branch: master
+    schedule:
+      interval: monthly


IndusAryan:
This bot is not that great, mend renovate is better plus I update libs manually for cs3 as they are not straight forward like other apps, they fail all the time

fire-light42:
The reason why cs3 does not use dependabot, is because cs3 depends on ***very specific*** versions of packages to work. Such as bumping Jackson above 2.13.1 crashes on Android TV's and FireSticks. Other times updating requires changing a lot of the API, so any automated system creating dummy prs will just make my life harder. 

I will close this PR because this bot will just be unnecessary and annoying. @IndusAryan is a better dependabot.

indrastorms:
It's mainly for track updates if build fails we can try to fix or close the pr the bot will not raise alert until another update.
And if build pass we can manually bump or merge the pr.

fire-light42:
> It's mainly for track updates if build fails we can try to fix or close the pr the bot will not raise alert until another update. And if build pass we can manually bump or merge the pr.

Regressions and crashes are not caught in an automated environment. Each dependency update requires careful testing and consideration. Hence it is less bothersome and time consuming to do large periodic and well tested bumps rather than many small bumps.

indrastorms:
> > It's mainly for track updates if build fails we can try to fix or close the pr the bot will not raise alert until another update. And if build pass we can manually bump or merge the pr.
> 
> Regressions and crashes are not caught in an automated environment. Each dependency update requires careful testing and consideration. Hence it is less bothersome and time consuming to do large periodic and well tested bumps rather than many small bumps.

All hail to @IndusAryan bot! ♥️

PR #1286, opened 2024-08-19
Added VidHidePro Extractor

misajimenezmx:
Avoiding mix vidhide.com and vidhidepro.com with a new vid hide pro extractor.


fire-light42:
Nice 👍

PR #1287, opened 2024-08-20
update vidsrc2to encryption

RowdyRushya:


fire-light42 (approved):
Not tested, but looks good 👍

PR #1288, opened 2024-08-21
Fix StreamWishExtractor and Voe

misajimenezmx:
Fixed because it works randomly for  "streamwish.to" domain


fire-light42 (approved):
Not tested, but looks good 👍

PR #1289, opened 2024-08-21
Update Chillx Extractor

b4byhuey:


fire-light42 (approved):
Not tested, but looks good 👍

PR #1290, opened 2024-08-21
Added StreamSilk extractor.

misajimenezmx:
Also fixed JSHunter


fire-light42 (approved):
Not tested, but looks good.

PR #1303, opened 2024-08-26
Fix DoodExtractor

b4byhuey:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/DoodExtractor.kt
```diff
@@ -58,31 +60,62 @@ class DoodYtExtractor : DoodLaExtractor() {
...
+        )
 
     }
-}
\ No newline at end of file
+	
+private fun createHashTable(): String {
+        return buildString {
+            repeat(10) {
+                append(alphabet[Random.nextInt(alphabet.length)])
```
CranberrySoup, line 111: alphabet.random() is nice to have here :+1:

b4byhuey: done

PR #1305, opened 2024-08-27
Subtitle offset preview

CranberrySoup:
Uses the new subtitle parsing to sync subtitles
All lines are scrollable and clickable:

image

Also increases the clickable size of the ± buttons when syncing


fire-light42 (approved):
Very nice 👍

PR #1309, opened 2024-08-29
fix: check if range supported in header

int3debug:
Bugfix for this error
Screenshot_20240328_125912_CloudStream


fire-light42:
As this is not marked with ready for review yet I wont give a full review, but here is what I would request changed:
1. We cant use LongArray(1) { startByte } on a non resumable bytestream, so you would have to restart from LongArray(1) { 0 } (along with deleting the current downloaded file), do this in a separate if statement
2. The head response should only check for `Accept-Ranges` and if that is set to `none` we know it is a non resumable bytestream, and if it is set to `bytes` it is resumable 
3. If step 2 does not get a key from `Accept-Ranges`, then we do a `get` request that requests the first 1024 bytes or so `Range: bytes=0-1023` and if that returns a partial content = 206, then we know it is a resumable bytestream, otherwise we assert that it is not and continue to step 1. This step can also check for the `Content-Range` total length if the head-request content length did not work.

fire-light42:
> OK thank you Sir And I thought everything was right this time :(

All prs are appreciated, so dont feel down 👍. You are messing around in the downloaded, and that is quite a tricky area of code to navigate and change.

fire-light42:
> @fire-light42 I forgot when editing Then 1. was always wrong? I didn't change it

The original code assumed that all streams are resumable, and if that was the case the code was correct. However the original code completely breaks on both paths on the if statement in different ways on non resumable streams. So you more or less just have to double check beforehand if it is non resumable, and handle it separately. 

if non resumable:
delete the current downloaded file if exsists
val ranges = from 0 bytes
else:
old code for val ranges =

fire-light42:
> @fire-light42 Can I do this as a hotfix so that it doesn't exceed 100% and the file is correct and the other thing you request I do a new PR? Because what you requested takes longer because I am not yet so good at coding

No, this code will result in a bug when resuming downloads, and as such will not be accepted. However, thanks for bringing this issue to light, I will fix it.

fire-light42:
See https://github.com/recloudstream/cloudstream/commit/b95f8ee2110df8140ad6313395da2c2c9381a9f1

PR #1324, opened 2024-09-05
Added GamoVideo extractor, improved streamtape, fix mixdrop and Vidmoly

redblacker8:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidmoly.kt
```diff
@@ -26,15 +27,25 @@ open class Vidmoly : ExtractorApi() {
...
+        while (attemps < 10 && script.isNullOrEmpty()){
+            attemps++
+            script = app.get(
+                newUrl,
+                headers = headers,
+                referer = referer,
+            ).document.select("script")
+                .firstOrNull { it.data().contains("sources:") }?.data()
+            if(script.isNullOrEmpty())
+                Thread.sleep(500)
```
fire-light42, line 47: Thread.sleep? Use delay to not stop execution of concurrent requests. https://stackoverflow.com/questions/61345712/what-is-the-difference-between-delay-and-thread-sleep-in-kotlin

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidmoly.kt
redblacker8: Thanks for the review, already replaced by sleep

fire-light42 (approved):

Issue #379, opened 2023-02-18
Better aspect ratio/video resize required


jackuars7:
### Describe your suggested feature

Would appreciate if there are better aspect ratio and video resizing available in internal player. Because of a large number of sources that we are playing the videos on, many of them don't look good with the limited options available.

Can you support these as available in VLC:
1. Best Fit
2. Fit screen
3. Fill screen
4. 16:9
5. 16:10 and other aspect ratios

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

MonuGit9:
Can you assign this task to me

KaiKitKobold:
Hello! I'm just commenting here to declare that I'm currently working on this issue.

Akshay-1024:
Is anyone still working on this?

Akshay-1024:
I would like to learn and contribute let me know

KaiKitKobold:
Hi! @Akshay-1024  Sorry, Because of School, I couldn't work on this issue, I don't think anyone else in the discord server or community is working on this right now, So you can contribute to this! ^_^ 

Have fun!

Akshay-1024:
Sure Thanks.

Akshay-1024:
I have raised a PR here for this issue

https://github.com/recloudstream/cloudstream/pull/1335

PR #1335, opened 2024-09-12
Initial Changes For video aspect ratio changes

Akshay-1024:
https://github.com/recloudstream/cloudstream/issues/379

Made changes related to aspect ratio in the AbstractPlayerFragment.kt to handle for standard aspect ratios

Link for the video of Pop Up
https://drive.google.com/file/d/1frhCkOWAaTo3n1yJEPEYhnsaAIlfYhDy/view?usp=sharing

### Updated ScreenShot using internal SingleSelectionHelper

image

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
@@ -1,6 +1,7 @@
 package com.lagradost.cloudstream3.ui.player
 
 import android.annotation.SuppressLint
+import android.app.AlertDialog
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
@@ -615,13 +616,40 @@ abstract class AbstractPlayerFragment(
     }
 
     fun nextResize() {
-        resizeMode = (resizeMode + 1) % PlayerResize.entries.size
-        resize(resizeMode, true)
+//        resizeMode = (resizeMode + 1) % PlayerResize.entries.size
+//        resize(resizeMode, true)
+        showAspectRatioDialog()
     }
 
     fun resize(resize: Int, showToast: Boolean) {
         resize(PlayerResize.entries[resize], showToast)
     }
+
+    private fun showAspectRatioDialog() {
+        // Pause the player when the dialog is shown using the handleEvent method
+        player.handleEvent(CSPlayerEvent.Pause, PlayerEventSource.UI)
+
+        val aspectRatioOptions = PlayerResize.values().map { getString(it.nameRes) }.toTypedArray()
+
+        // Create and show the dialog
+        AlertDialog.Builder(requireContext())
+            .setTitle(getString(R.string.choose_aspect_ratio))
+            .setItems(aspectRatioOptions) { _, which ->
+                val selectedOption = PlayerResize.values()[which]
+                resize(selectedOption, true)
+
+                player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
+            }
+            .setOnCancelListener {
+                player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
+            }
+            .setNegativeButton(android.R.string.cancel) { _, _ ->
+                player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
+            }
+            .show()
+    }
+
+
 
     @SuppressLint("UnsafeOptInUsageError")
     fun resize(resize: PlayerResize, showToast: Boolean) {
@@ -633,17 +661,18 @@ abstract class AbstractPlayerFragment(
             PlayerResize.SixteenByNine -> 16.0f/9.0f
             PlayerResize.NineBySixteen -> 9.0f/16.0f
             PlayerResize.OneByOne -> 1.0f/1.0f
-            PlayerResize.IMAX -> 1.43f / 1.0f
-            PlayerResize.DolbyVision -> 2.39f / 1.0f
-            PlayerResize.UltraWide -> 21.0f / 9.0f
-            PlayerResize.Expanded -> 1.90f / 1.0f
+            PlayerResize.IMAX -> 1.43f/1.0f
+            PlayerResize.DolbyVision -> 2.39f/1.0f
+            PlayerResize.UltraWide -> 21.0f/9.0f
+            PlayerResize.Expanded -> 1.90f/1.0f
         }
         if(type is Float){
             playerView?.resizeMode = AspectRatioFrameLayout.RESIZE_MODE_FIT
             aspectRatioFrameLayout?.setAspectRatio(type)
-        }else {
-            playerView?.resizeMode = type as Int
+        }else if(type is Int) {
+            playerView?.resizeMode = type
         }
+
         if (showToast)
             showToast(resize.nameRes, Toast.LENGTH_SHORT)
     }
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
@@ -617,8 +617,6 @@ abstract class AbstractPlayerFragment(
     }
 
     fun nextResize() {
-//        resizeMode = (resizeMode + 1) % PlayerResize.entries.size
-//        resize(resizeMode, true)
         showAspectRatioDialog()
     }
 
@@ -627,19 +625,16 @@ abstract class AbstractPlayerFragment(
     }
 
     private fun showAspectRatioDialog() {
-        // Pause the player when the dialog is shown
         player.handleEvent(CSPlayerEvent.Pause, PlayerEventSource.UI)
 
         val aspectRatioOptions = PlayerResize.values().map { getString(it.nameRes) }
 
-        // Using activity?.showDialog with a safe call
         activity?.showDialog(
             aspectRatioOptions,
-            -1, // Use -1 if there's no pre-selected item
+            -1,
             getString(R.string.choose_aspect_ratio),
             false, // Immediate selection without an Apply button
             dismissCallback = {
-                // Resume the player when the dialog is dismissed without selection
                 player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
             },
             callback = { selectedIndex ->
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
@@ -638,7 +638,7 @@ abstract class AbstractPlayerFragment(
                 player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
             },
             callback = { selectedIndex ->
-                val selectedOption = PlayerResize.values()[selectedIndex]
+                val selectedOption = PlayerResize.entries[selectedIndex]
                 resize(selectedOption, true)
                 player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
             }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -622,9 +630,20 @@ abstract class AbstractPlayerFragment(
...
+            PlayerResize.IMAX -> 1.43f / 1.0f
+            PlayerResize.DolbyVision -> 2.39f / 1.0f
+            PlayerResize.UltraWide -> 21.0f / 9.0f
+            PlayerResize.Expanded -> 1.90f / 1.0f
+        }
+        if(type is Float){
+            playerView?.resizeMode = AspectRatioFrameLayout.RESIZE_MODE_FIT
+            aspectRatioFrameLayout?.setAspectRatio(type)
+        }else {
+            playerView?.resizeMode = type as Int
```
CranberrySoup, line 645: Avoid unchecked casts in Android, they will crash if not satisfied. Use `else if (type is Int)` for the exact same effect, but more future proof

app/src/main/res/values/strings.xml
```diff
@@ -433,6 +433,13 @@
     <string name="resize_fit">Fit to screen</string>
     <string name="resize_fill">Stretch</string>
     <string name="resize_zoom">Zoom</string>
+    <string name="resize_sixteen_by_nine">SixteenByNine</string>
```
CranberrySoup, line 436: I dont know if these are placeholders, but if they are not, then "Widescreen 16:9" and "Square 1:1" might be more readable and consistent

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -622,9 +630,20 @@ abstract class AbstractPlayerFragment(
             PlayerResize.Fill -> AspectRatioFrameLayout.RESIZE_MODE_FILL
             PlayerResize.Fit -> AspectRatioFrameLayout.RESIZE_MODE_FIT
             PlayerResize.Zoom -> AspectRatioFrameLayout.RESIZE_MODE_ZOOM
+            PlayerResize.SixteenByNine -> 16.0f/9.0f
+            PlayerResize.NineBySixteen -> 9.0f/16.0f
+            PlayerResize.OneByOne -> 1.0f/1.0f
+            PlayerResize.IMAX -> 1.43f / 1.0f
+            PlayerResize.DolbyVision -> 2.39f / 1.0f
+            PlayerResize.UltraWide -> 21.0f / 9.0f
+            PlayerResize.Expanded -> 1.90f / 1.0f
```
CranberrySoup, line 639: With this many options a popup might be preferable over pressing resize 9 times, but this is a great first framework!

CranberrySoup (commented):
Good job!
It's an early review so you might already be working on the details I mentioned, then you can just dismiss those comments.

You solved the aspect ratio problem very nicely :+1:

C10udburst:
btw, if you want to indicate you are still working on a PR and don't want it merged yet its good practise to mark it as draft. there should be a button on the right hand side to do so

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Akshay-1024: Yea, this is correct. I have Updated this code in my local

app/src/main/res/values/strings.xml
Akshay-1024: Updated them as needed

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Akshay-1024: I have made changes for the pop up, Attaching the video in the PR for reference let me know if I have to make any changes or suggestions

Akshay-1024:
> btw, if you want to indicate you are still working on a PR and don't want it merged yet its good practise to mark it as draft. there should be a button on the right hand side to do so

Yes, I would follow that from the next time. Thanks.

CranberrySoup (commented):
Looks good code wise, however I have not had time to test it yet

fire-light42:
Use SingleSelectionHelper.showDialog to show a dialog in the cs3 theme.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -607,23 +616,62 @@ abstract class AbstractPlayerFragment(
...
     }
 
+    private fun showAspectRatioDialog() {
+        // Pause the player when the dialog is shown using the handleEvent method
+        player.handleEvent(CSPlayerEvent.Pause, PlayerEventSource.UI)
+
+        val aspectRatioOptions = PlayerResize.values().map { getString(it.nameRes) }.toTypedArray()
+
+        // Create and show the dialog
+        AlertDialog.Builder(requireContext())
```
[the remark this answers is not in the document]
Akshay-1024: Using SingleSelectionHelper.showDialog i am failing it silently.

Akshay-1024:
> Use SingleSelectionHelper.showDialog to show a dialog in the cs3 theme.

Thanks Used it.

Akshay-1024:
Hi @C10udburst , @CranberrySoup , @fire-light42 any update on this?????

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -607,23 +617,57 @@ abstract class AbstractPlayerFragment(
...
     }
 
     fun resize(resize: Int, showToast: Boolean) {
         resize(PlayerResize.entries[resize], showToast)
     }
 
+    private fun showAspectRatioDialog() {
+        player.handleEvent(CSPlayerEvent.Pause, PlayerEventSource.UI)
+
+        val aspectRatioOptions = PlayerResize.values().map { getString(it.nameRes) }
```
CranberrySoup, line 630:
```suggestion
        val aspectRatioOptions = PlayerResize.entries.map { getString(it.nameRes) }
```
Better to use the recommended replacement for values

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -607,23 +617,57 @@ abstract class AbstractPlayerFragment(
...
+        activity?.showDialog(
+            aspectRatioOptions,
+            -1,
+            getString(R.string.choose_aspect_ratio),
+            false, // Immediate selection without an Apply button
+            dismissCallback = {
+                player.handleEvent(CSPlayerEvent.Play, PlayerEventSource.UI)
+            },
+            callback = { selectedIndex ->
+                val selectedOption = PlayerResize.values()[selectedIndex]
```
CranberrySoup, line 641:
```suggestion
                val selectedOption = PlayerResize.entries[selectedIndex]
```

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -607,23 +617,57 @@ abstract class AbstractPlayerFragment(
...
     }
 
+    private fun showAspectRatioDialog() {
+        player.handleEvent(CSPlayerEvent.Pause, PlayerEventSource.UI)
+
+        val aspectRatioOptions = PlayerResize.values().map { getString(it.nameRes) }
+
+        activity?.showDialog(
+            aspectRatioOptions,
+            -1,
```
CranberrySoup, line 634:
```suggestion
            DataStoreHelper.resizeMode,
```
Make the selected index what the user has previously picked

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -607,23 +617,57 @@ abstract class AbstractPlayerFragment(
...
+            PlayerResize.IMAX -> 1.43f/1.0f
+            PlayerResize.DolbyVision -> 2.39f/1.0f
+            PlayerResize.UltraWide -> 21.0f/9.0f
+            PlayerResize.Expanded -> 1.90f/1.0f
+        }
+        if(type is Float){
+            playerView?.resizeMode = AspectRatioFrameLayout.RESIZE_MODE_FIT
+            aspectRatioFrameLayout?.setAspectRatio(type)
+        }else if(type is Int) {
+            playerView?.resizeMode = type
```
CranberrySoup, line 669:
```suggestion
            playerView?.resizeMode = type
            aspectRatioFrameLayout?.setAspectRatio(0f) // Reset the aspect ratio when using built in resizing
```
Without this the implementation is bugged. This can easily be checked by switching from "Square" to "Fit to screen"

CranberrySoup (changes requested):
The bug is critical and the index should be fixed for app-wide consistency. You can just click and accept the changes.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Akshay-1024: Yes, I am also getting this while switching from Square to fit to screen, but after adding the setAspectRatio(0f) , after testing it was missing when switching from stretch to zoom, and other custom aspect ratio to predefined like fit, fill, zoom all behaving like fill. I am looking into that.

Akshay-1024: Once I will try this on real device

Akshay-1024: There maybe some issue in my code I will resolve it. I do agree with the idea setAspectRatio(0f) should fix the issue.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -607,23 +617,57 @@ abstract class AbstractPlayerFragment(
...
         }
-        playerView?.resizeMode = type
 
         if (showToast)
             showToast(resize.nameRes, Toast.LENGTH_SHORT)
```
C10udburst, line 673: Since after this change the aspect ratio is explicitly set by the user, it may be worth thinking if this toast is necessary.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
fire-light42: @Akshay-1024 I cannot merge without a fix to this bug. Should I apply the fix or are you working on something?

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Akshay-1024: Removed Toast from the Resize function.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Akshay-1024: @fire-light42 I have removed fit as its working same as widescreen, I have removed Zoom as well because after the changes Zoom is basically behaving same as fill So Kept only Fill for predefined and rest all are custom aspect ratios rest all are working fine, but if you have any fix for zoom as well let me know. I have Pushed the changes

fire-light42: I cant merge this if you have removed Fit and Zoom, all three original options act differently depending on screen size and video size. Just because they look the same on your phone, does not mean they function identically.

fire-light42 (changes requested):
After testing the code, I can say that this code is deeply flawed. The concept of overriding the aspect ratio via setAspectRatio overrides the aspect ratio information set by the playerView means that we cant recover the aspect ratio for RESIZE_MODE_FILL RESIZE_MODE_FIT and RESIZE_MODE_ZOOM if we choose to go back. 

By setting aspect ratio to 0.0f we dont fix anything as the aspect ratio view wont render using the resizemode. 

AspectRatioFrameLayout::onMeasure
```kotlin
if (videoAspectRatio <= 0) {
      // Aspect ratio not set.
      return;
}
``` 

Futhermore, the code is competing with contentFrame.setAspectRatio from playerview so even if you store the variables locally, the player might decide to override that aspect ratio as well.

PR #1338, opened 2024-09-17
Update MainAPI.kt

redowan99:
Adding getYearFromString()

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -1217,6 +1217,28 @@ interface LoadResponse {
         }
     }
 }
+/**
+ * Extracts a four-digit year from a string, prioritizing years in parentheses and ensuring no word characters follow.
+ *
+ * Example:
+ *
+ * "This is (2023) movie" -> 2023
+ *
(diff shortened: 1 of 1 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1217,6 +1217,28 @@ interface LoadResponse {
...
+ * "This is 2023 movie" -> 2023
+ *
+ * "This is 1999-2010 TvSeries" -> 1999
+ *
+ * @param check The input string.
+ * @return The year as an integer, or `null` if no match is found.
+ */
+fun getYearFromString(check: String?): Int? {
+    return check?.let {
+        "(?<=\\()\\d{4}(?=\\))".toRegex().find(it)?.value?.toIntOrNull()
```
fire-light42, line 1238: Regex construction makes this function take ~10x longer than it should. As such, I would like if you moved both of the .toRegex() into private variables in the global scope

fire-light42 (changes requested):

fire-light42:
Why was this closed?

C10udburst:
silly

redowan99:
I didn't know deleting the fork, would close the commit. I think I need to learn to work with git first🤦‍♂️.

PR #1339, opened 2024-09-17
Fixing some extractors

redowan99:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/DoodExtractor.kt
b4byhuey, line 1: I think you forget to added "DoodLiExtractor" to "[ExtractorApi.kt](https://github.com/recloudstream/cloudstream/blob/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt)"

PR #1343, opened 2024-09-20
Dependencies version catalog

KingLucius:
There was a PR already opened for this but I can't find it so I have read about it and here is a new PR.
You're free to close it if the old PR still exist.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,0 +1,112 @@
+[versions]
+acraCore = "5.11.3"
+appcompat = "1.7.0"
+autoServiceKsp = "1.2.0"
+biometric = "1.2.0-alpha05"
+colorpicker = "1.1.0"
+conscryptAndroid = "2.5.2"
+constraintlayout = "2.1.4"
+coreKtx = "1.13.1"
+desugar_jdk_libs_nio = "2.0.4"
+espressoCore = "3.6.1"
+fuzzywuzzy = "1.4.0"
+glide = "4.16.0"
+glideTransformations = "4.3.0"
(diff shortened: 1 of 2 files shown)


C10udburst:
Consider updating build scripts for library and docs as well

KingLucius:
> Consider updating build scripts for library and docs as well

Done

phisher98:
This goona be awesome i am feeling it

gradle/libs.versions.toml
```diff
@@ -0,0 +1,122 @@
...
+junitKtx = "1.2.1"
+junitVersion = "1.2.1"
+juniversalchardet = "2.5.0"
+kotlinGradlePlugin = "1.9.23"
+kotlinxCoroutinesCore = "1.8.0"
+ksp = "4.16.0"
+lifecycleLivedataKtx = "2.8.3"
+lifecycleViewmodelKtx = "2.8.3"
+material = "1.12.0"
+media3Cast = "1.4.0"
```
fire-light42, line 31: Why is every media3 dep separate entries here? Is not the entire point of this system to union the dependencies, so you only have to define a media3 ver and all media3Cast,media3Common,media3DatasourceOkhttp ect gets the same?

fire-light42 (commented):
Last PR was not merged due to me being unable to compile it

gradle/libs.versions.toml
KingLucius: Yes, it's up to us if we want to union them together no issue we can do this

KingLucius:
Any issue about that PR I should fix ?

Issue #1302, opened 2024-08-26
Don't ask again option add plesae


xcbly:
### Describe your suggested feature

![Screenshot_3](https://github.com/user-attachments/assets/5032ca95-c69c-4942-89b8-89aaa2cb20e8)

### Acknowledgements

- [X] My suggestion is **NOT** about adding a new provider
- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

zzjjaayy:
No one seems to be working on this, taking this up!

PR #1345, opened 2024-09-22
#1302 Option to not show exit confirmation dialog

zzjjaayy:
Fixes: [Issue#1302](https://github.com/recloudstream/cloudstream/issues/1302)
- Tested on Debug & Release Builds
- Added a setting to hide it or show it again {only where it is supported -  TV/EMULATOR}
- Checkbox in dialog to disable it
- Let me know if I am supposed to handle translations for newly added strings, i believe the weblate bot does that.
- NOTE: check SettingsFragment.kt:71-79, now returns null when not visible so further actions on the preference dont happen unnecessarily. Checked other usages, works fine.

Screenshot 2024-09-22 at 16 12 42
Screenshot 2024-09-22 at 16 12 26

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -125,6 +125,7 @@ import com.lagradost.cloudstream3.ui.settings.Globals.TV
 import com.lagradost.cloudstream3.ui.settings.Globals.isLayout
 import com.lagradost.cloudstream3.ui.settings.Globals.updateTv
 import com.lagradost.cloudstream3.ui.settings.SettingsGeneral
+import com.lagradost.cloudstream3.ui.settings.getExitDialogSupportedLayouts
 import com.lagradost.cloudstream3.ui.setup.HAS_DONE_SETUP_KEY
 import com.lagradost.cloudstream3.ui.setup.SetupFragmentExtensions
 import com.lagradost.cloudstream3.utils.ApkInstaller
@@ -1534,7 +1535,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
                 }
             }
 
-            if (isLayout(TV or EMULATOR)) {
+            if (isLayout(getExitDialogSupportedLayouts())) {
                 if (navDestination.matchDestination(R.id.navigation_home)) {
                     attachBackPressedCallback {
                         showConfirmExitDialog()
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -125,7 +125,7 @@ import com.lagradost.cloudstream3.ui.settings.Globals.TV
 import com.lagradost.cloudstream3.ui.settings.Globals.isLayout
 import com.lagradost.cloudstream3.ui.settings.Globals.updateTv
 import com.lagradost.cloudstream3.ui.settings.SettingsGeneral
-import com.lagradost.cloudstream3.ui.settings.getExitDialogSupportedLayouts
+import com.lagradost.cloudstream3.ui.settings.SettingsUI
 import com.lagradost.cloudstream3.ui.setup.HAS_DONE_SETUP_KEY
 import com.lagradost.cloudstream3.ui.setup.SetupFragmentExtensions
 import com.lagradost.cloudstream3.utils.ApkInstaller
@@ -616,7 +616,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
     @SuppressLint("ApplySharedPref") // commit since the op needs to be synchronous
     private fun showConfirmExitDialog() {
         val settingsManager = PreferenceManager.getDefaultSharedPreferences(this)
-        val confirmBeforeExit = settingsManager.getBoolean(getString(R.string.confirm_exit_key), true)
+        val confirmBeforeExit = settingsManager.getBoolean(getString(R.string.confirm_exit_key), SettingsUI.getConfirmExitDialogDefault())
         if (!confirmBeforeExit) {
             exitProcess(0)
             return
@@ -1535,16 +1535,14 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
                 }
             }
 
-            if (isLayout(getExitDialogSupportedLayouts())) {
-                if (navDestination.matchDestination(R.id.navigation_home)) {
-                    attachBackPressedCallback {
-                        showConfirmExitDialog()
-                        window?.navigationBarColor =
-                            colorFromAttribute(R.attr.primaryGrayBackground)
-                        updateLocale()
-                    }
-                } else detachBackPressedCallback()
-            }
+            if (navDestination.matchDestination(R.id.navigation_home)) {
+                attachBackPressedCallback {
+                    showConfirmExitDialog()
+                    window?.navigationBarColor =
+                        colorFromAttribute(R.attr.primaryGrayBackground)
+                    updateLocale()
+                }
+            } else detachBackPressedCallback()
         }
 
         //val navController = findNavController(R.id.nav_host_fragment)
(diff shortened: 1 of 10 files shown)


C10udburst:
Perhaps you could make the exit confirmation setting visible on phone but disabled by default.

C10udburst:
Also I think it would be better to put the setting in ui settings and not general settings.

zzjjaayy:
done.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -187,5 +194,12 @@ class SettingsUI : PreferenceFragmentCompat() {
             return@setOnPreferenceClickListener true
         }
 
+        setKey( // default value is layout specific
```
fire-light42, line 197: This is just wrong, as setKey will only set the first time you go into settings. If you want this behavior, then dont use a SwitchPreference, and instead use a bottom popup with the options `show`, `dont show`, and `auto` like we do App Layout.

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -724,14 +724,28 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
...
         }
+
+        val dialogView = layoutInflater.inflate(R.layout.confirm_exit_dialog, null)
+        val dontShowAgainCheck: CheckBox = dialogView.findViewById(R.id.checkboxDontShowAgain)
+        val builder: AlertDialog.Builder = AlertDialog.Builder(this)
+        builder.setView(dialogView)
+            .setTitle(R.string.confirm_exit_dialog)
+            .setNegativeButton(R.string.no) { _, _ -> /*NO-OP*/}
+            .setPositiveButton(R.string.yes) { _, _ ->
+                if(dontShowAgainCheck.isChecked) {
```
fire-light42, line 743: Only save on yes?

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -20,6 +24,9 @@ import com.lagradost.cloudstream3.utils.SingleSelectionHelper.showMultiDialog
 import com.lagradost.cloudstream3.utils.UIHelper.hideKeyboard
 
 class SettingsUI : PreferenceFragmentCompat() {
+    companion object {
+        fun getConfirmExitDialogDefault() = isLayout(TV or EMULATOR)
```
fire-light42, line 28: This is useless impo

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsFragment.kt
```diff
@@ -70,11 +70,12 @@ class SettingsFragment : Fragment() {
 
         /**
          * Hide the Preference on selected layouts.
+         * @return [Preference] if visible otherwise null
```
fire-light42, line 73: Please add *why* in the comment. I assume it is because of performance, but that is not stated here.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
fire-light42: A two state button cant capture a tri state system.

fire-light42 (changes requested):
Good work with the code, and I may give another one after testing this PR.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
zzjjaayy: I don't see this being a tri-state system, there is only `show` or `dont_show`. Only extra factor here being the default value of the switch only valuable when the `key` isnt in the sharedpref yet. `setKey` works in that case, only there because the default value cannot be set in the xml since it needs some logic. I agree on if there were to be more than 2 states, the sheet would have been a better idea, switch should work for us here.

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
zzjjaayy: yes! no seems like a dismissing action, usually users don't expect any change then.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
zzjjaayy: I would agree but since multiple people contribute this fun makes it difficult to introduce bugs, say when we decide to change the default value in a certain layou.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsFragment.kt
zzjjaayy: makes sense, will do.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
CranberrySoup: It *is* a tri-state system

### Clean install: 
No value is set for `confirm_exit_key`

```kt
val confirmBeforeExit = settingsManager.getBoolean(getString(R.string.confirm_exit_key), SettingsUI.getConfirmExitDialogDefault())
```
As no value is set it will default to:
```kt
SettingsUI.getConfirmExitDialogDefault()) 
```
Which equals to `isLayout(TV or EMULATOR)`, **which is dynamic and based on the current layout.**

### Visited settings once:
The setting behavior is **pinned** based on the current layout used when visiting settings.
This is done when doing 
```kt
setKey(
       getString(R.string.confirm_exit_key),
            settingsManager.getBoolean(
                getString(R.string.confirm_exit_key),
                getConfirmExitDialogDefault()
            )
        )
```
Which essentially evaluates to `setKey(getString(R.string.confirm_exit_key), isLayout(TV or EMULATOR))` on a clean install.

### Changed the setting manually:
The value is also pinned to another value, not based on layout.

--- 
So we have:
1. Based on layout
2. True
3. False

Whereas if you go to either (2) true or (3) false it is not possible to restore it to (3) dynamic layout.

To demonstrate the issue with this system we can imagine a off-brand Android TV which is not automatically recognized as a TV. In this case when the used goes into layout settings to change their UI this value will be set to **false**. This means that the user, while never doing anything related to this setting will have it permanently turned off, without even knowing it existed.

You **need** a tri-state option if you want the dialog to be able to be shown in phones, as opposed to the old behavior of only TVs.

CranberrySoup (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
zzjjaayy: I believe this makes sense. Thank you for the detailed analysis, will fix and get back!

zzjjaayy:
moved to a tri state system in the latest commit. I did some thinking and I still don't think this solves much for us. the "off brand TV" will still have the dialog not showing since the layout isn't recognised. The AUTO value will just work differently for them. Also have we considered using these flags? - https://stackoverflow.com/a/27138979 or use the screen size as a factor? 

Also attaching screenshots for the latest changes - 
image
image

CranberrySoup (approved):
Looks good and works well :+1:

fire-light42:
> The AUTO value will just work differently for them. Also have we considered using these flags? - https://stackoverflow.com/a/27138979 or use the screen size as a factor?

Auto uses UI_MODE_TYPE_TELEVISION to auto detect if the device is a TV. No one has created an issue that auto does not work so far. 
https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/ui/settings/Globals.kt#L24

zzjjaayy:
> Auto uses UI_MODE_TYPE_TELEVISION to auto detect if the device is a TV. No one has created an issue that auto does not work so far. 

If that's the case, do you think AUTO in confirm exit options makes sense? If we are always able to reliably tell if its a tv or not, we should not have AUTO, the dynamic default value works so much better. Also AUTO doesn't have any semantic value for the user, a switch works so much better there.

zzjjaayy:
Alright then, keeping it AUTO.

fire-light42 (approved):
Looks good to me

Issue #1152, opened 2024-06-23
Save backups in a subdirectory


Luna712:
### Describe your suggested feature

At the very least if we aren't doing a separate location selection, it would be nice if it saved backups in a subdirectory (such as "Backups") since everything else saves in sub directories. It would keep things more organized for me anyway.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

iswac:
I have already opened this issue https://github.com/recloudstream/cloudstream/issues/1106

Luna712:
> I have already opened this issue #1106

This is a different issue. This isn't asking for a custom directory just saving to a sub directory in selected downloads directory similar to how other download types do.

iswac:
yeah i missed up but both of us want similar feature

DestroyerBDT:
I like this

Issue #1106, opened 2024-05-26
Ask location when saving backup


iswac:
### Describe your suggested feature

When I try backup option, unlike other apps, it saves backup file every time under the Download/ folder.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

iswac:
It's especially for downloaded things. I don't want my movies and backups to get the same folder

PR #1354, opened 2024-09-26
feat: option to choose backup target directory

zzjjaayy:
Fixes: #1152 and #1106 
- Settings option to choose backup target directory only in PHONE {Currently saved in the open in Downloads/}
- I don't recommend having a different default directory since that fails in TVs and choosers are anyways not present in most of them.

Screenshot_2024-09-27-09-38-58-27_bba4627ecec1346e77f282610a5b88e8

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUpdates.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUpdates.kt
@@ -113,7 +113,7 @@ class SettingsUpdates : PreferenceFragmentCompat() {
             activity?.showBottomDialog(
                 dirs + listOf(getString(R.string.custom)),
                 dirs.indexOf(currentDir),
-                getString(R.string.backup_dir_key),
+                getString(R.string.backup_path_title),
                 true,
                 {}) {
                 // Last = custom
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUpdates.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUpdates.kt
@@ -128,9 +128,9 @@ class SettingsUpdates : PreferenceFragmentCompat() {
                     // path = used uri
                     // dir = dir path
                     settingsManager.edit()
-                        .putString(getString(R.string.backup_path_key), dirs[it]).apply()
-                    settingsManager.edit()
-                        .putString(getString(R.string.backup_dir_key), dirs[it]).apply()
+                        .putString(getString(R.string.backup_path_key), dirs[it])
+                        .putString(getString(R.string.backup_dir_key), dirs[it])
+                        .apply()
                 }
             }
             return@setOnPreferenceClickListener true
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUpdates.kt
```diff
@@ -47,19 +54,27 @@ class SettingsUpdates : PreferenceFragmentCompat() {
...
+                .putString(getString(R.string.backup_dir_key), it)
+                .apply()
+        }
+    }
+
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
         hideKeyboard()
         setPreferencesFromResource(R.xml.settings_updates, rootKey)
-        //val settingsManager = PreferenceManager.getDefaultSharedPreferences(requireContext())
+        val settingsManager = PreferenceManager.getDefaultSharedPreferences(requireContext())
```
zzjjaayy, line 70: using this top level declaration instead of in each listener. any reason to use separate calls?

app/src/main/java/com/lagradost/cloudstream3/utils/BackupUtils.kt
```diff
@@ -69,7 +77,12 @@ object BackupUtils {
 
         "biometric_key", // can lock down users if backup is shared on a incompatible device
         "nginx_user", // Nginx user key
-        "download_path_key" // No access rights after restore data from backup
+
+        // No access rights after restore data from backup
+        "download_path_key",
+        "download_path_pref",
```
zzjjaayy, line 83: added `download_path_pref` in ignoreForBackup list, check if that is something we left deliberately.

zzjjaayy: also check this value in `strings.xml`, why are we not keeping it `translatable=false`?

zzjjaayy (commented):
Questions/ThingsToLookAt for the reviewer. 
Thanks.

app/src/main/java/com/lagradost/cloudstream3/utils/BackupUtils.kt
CranberrySoup: Looks like a __big__ mistake, download_path_pref is used both as a display string and as a key, it should never be used as both.

CranberrySoup (commented):
Code looks good :+1:

app/src/main/java/com/lagradost/cloudstream3/utils/BackupUtils.kt
zzjjaayy: correct, I would fix it but would let one of the maintainers comment on this.
\

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUpdates.kt
zzjjaayy: Does it? if thats the case we could have a null check usage of `it.context`.

CranberrySoup: No, you misunderstand.

`requireContext()`: can crash, but no wet code
`it.context`: can't crash, but wet code

zzjjaayy:
fixed conflicts, clean for review again.

Issue #1356, opened 2024-09-28
Padding around systemBars for swipe detection in player


zzjjaayy:
### Describe your suggested feature

Need some padding around the system bars (navigation and status bar) for the swipe action detection. Whenever I try to pull down the notification shade from the player, the volume/brightness changes are triggered annoyingly. The same thing happens when I try to reach for the navigation bar which is hidden in the player.

### Acknowledgements

- [X] My suggestion is **NOT** about adding a new provider
- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

fire-light42:
That is already the case: https://github.com/recloudstream/cloudstream/blob/bd95ecf5ba0d30da76103597b40ebe77b80d2ad9/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt#L205

https://github.com/recloudstream/cloudstream/blob/bd95ecf5ba0d30da76103597b40ebe77b80d2ad9/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt#L940

zzjjaayy:
@fire-light42 I wanted "padding" around the system bars. Whenever i or any user tries to reach for these bars when the player is full screen we swipe down or up which often covers region larger than the bars themselves which triggers the setting. Happens almost always for me. Attaching a video which should explain the issue better, notice how i am dragging from the very bottom to reveal the navbar but it triggers the setting.

https://github.com/user-attachments/assets/fb6785a5-dc54-464e-88c3-ba7461201d19

fire-light42:
The code right now only accounts for top and right (three-button navigation and notifications bar) however I did not know that the bottom was a swipe action on some phones.

PR #1358, opened 2024-09-29
fix: accidental motion when accessing systemBars

zzjjaayy:
Fixes: #1356 
Deducts system insets from permissible region for motionEvents. 
Works on both below and above API Level 30 and using either button or gesture nav.

https://github.com/user-attachments/assets/19b3f703-6c9a-4184-8ef4-3f8e1586cc1d


zzjjaayy:
@reviewers CodeFactor suggests removal of some parenthesis, I feel they improve readability, let me know if I should remove them.

PR #1370, opened 2024-10-07
Added Filegram and Fix Vidguard Extractor

b4byhuey:


fire-light42 (changes requested):
You forgot to include it in the extractor list.

b4byhuey:
Added

fire-light42 (approved):

PR #1382, opened 2024-10-16
Fix opensubtitles login

CranberrySoup:
The previous logins did not work with special characters (such as '%') due to encoding. Changing the parameter fixes the encoding.


fire-light42:
Very nice!

PR #1392, opened 2024-10-24
refact(download): change download filename convention

int3debug:
reducing filename length for clarity

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -294,10 +294,10 @@
     <string name="acra_report_toast">Sorry, the application crashed. An anonymous bug report will be sent to the
         developers
     </string>
-    <string name="season">Season</string>
+    <string name="season">S</string>
     <string name="season_format">%1$s %2$d%3$s</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -294,10 +294,10 @@
     <string name="acra_report_toast">Sorry, the application crashed. An anonymous bug report will be sent to the
         developers
     </string>
-    <string name="season">Season</string>
+    <string name="season">S</string>
     <string name="season_format">%1$s %2$d%3$s</string>
     <string name="no_season">No Season</string>
-    <string name="episode">Episode</string>
+    <string name="episode">E</string>
```
CranberrySoup, line 300: This change would impact all other uses of this in the app (result UI and player UI), when changing things it is preferable to only change what you intend or need to change, otherwise it will introduce bugs or unwanted behavior. This change would also only affect English users and changed translations are less likely to be fixed that non-existing translations.

You are looking for episode_short and season_short, which already exists and is translated in other languages already.

fire-light42:
Close due to stale.

PR #1396, opened 2024-10-27
add singleton image loader, remove unused libs, increase disk & decrease memory cache

IndusAryan:
annotation processors are now needed if we shift to room sqlite or DI, kapt & ksp were for glide so as transformation lib

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
@@ -41,13 +41,13 @@ object ImageLoader {
             .build()
 
         return ImageLoader.Builder(context)
-            .crossfade(300)
+            .crossfade(250)
             .respectCacheHeaders(false)
             /** !Only use default placeholders and errors, if not using this instance for local
              * image buttons because when animating this will appear or in more cases **/
             //.placeholder(R.drawable.logo)
             //.error(R.drawable.logo)
-            //.allowHardware(true) // takes a toll on battery (only allow if app is like instagram or photos)
+            .allowHardware(false) // takes a toll on battery (only allow if app is like instagram or photos)
             .memoryCache {
                 MemoryCache.Builder(context)
                     .maxSizePercent(0.10) // Use 10 % of the app's available memory for caching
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
```diff
@@ -40,23 +41,23 @@ object ImageLoader {
...
         return ImageLoader.Builder(context)
-            .crossfade(true)
+            .crossfade(250)
             .respectCacheHeaders(false)
             /** !Only use default placeholders and errors, if not using this instance for local
              * image buttons because when animating this will appear or in more cases **/
             //.placeholder(R.drawable.logo)
             //.error(R.drawable.logo)
-            .allowHardware(true)
+            .allowHardware(false) // takes a toll on battery (only allow if app is like instagram or photos)
```
fire-light42, line 50: Is this really the case? I thought hardware bitmaps enabled the hardware to save energy by "offloading the work to dedicated co-processors". Can you link where you got this information or explain more in detail.

IndusAryan: the theory is hw decoding works well only in devices that are recent maybe like sdk int > 28, and it is only needed to render large bitmaps.
actually we are fucked instacart released coil 3 in which we dont need our own singleton instead we should inherit their factory in application class

fire-light42 (changes requested):
This does not work on image resources, like the source profiles.

IndusAryan:
fixed load bitmap from drawable, and for the hardware thing, we got software render error, because coil themselves have to solve that bitmap render issue on unsupported device

IndusAryan:
@fire-light42 should we use okhttp client from requests helper here instead of building?

fire-light42:
> @fire-light42 should we use okhttp client from requests helper here instead of building?

Yes

IndusAryan:
i tried to add it, the base client is very confusing, do we set there for each network request, i hope this works same

fire-light42 (approved):
I do not see any obvious issues, so I will merge this after testing 👍

PR #1415, opened 2024-11-22
Fix subtitle decoding on some subtitle files

dependencyrejection:
Patch proposed by @Blatzar and works on the example error file


fire-light42:
Very cool

PR #1423, opened 2024-12-12
Move to Dependencies version catalog

KingLucius:
Created new PR as I couldn't solve conflicts
- Used bundles for media3 to push all in one update.
- same catalog for app, lib & docs to simplify across updates.


fire-light42:
LGTM, lets hope this does not break everything.

PR #1429, opened 2024-12-17
refactor: overall navigation mechanism and split logic

IndusAryan:
added smol safedebounce to prevent intentional || unintentional rapid transitions
upgrade coil, navigation and desugar
properly name easter egg activity
dedicated function for opening activities outside nav graph 
support for adding custom tag in log error utility (optional)
better navigate function
the reason of replacing appcompat woth component is that it is slightly lightweight, although main activity cant be turned unless compose is used

---
this pr is like a general improvement


C10udburst:
Please update your fork first. You are missing the dependency catalog PR.

PR #1432, opened 2024-12-21
feat(DRM): License URL support for Extensions

KingLucius:
Extensions developers can use OnlineDrmExtractorLink to provide the Drm license URL.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
@@ -9,9 +9,16 @@ import com.lagradost.cloudstream3.extractors.AStreamHub
 import com.lagradost.cloudstream3.extractors.Acefile
 import com.lagradost.cloudstream3.extractors.Ahvsh
 import com.lagradost.cloudstream3.extractors.Aico
+import com.lagradost.cloudstream3.extractors.AnimesagaStream
+import com.lagradost.cloudstream3.extractors.Anplay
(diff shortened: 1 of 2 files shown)


fire-light42 (changes requested):
Why do you need an extra type, just extend DrmExtractorLink instead.

fire-light42:
This will also be blocked until stable like #1216

KingLucius:
> Why do you need an extra type, just extend DrmExtractorLink instead. 

We can go the other approach extending DrmExtractorLink and based on the existence of keyUrl will setup Media3 configuration online/offline

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -460,11 +459,12 @@ open class DrmExtractorLink private constructor(
...
         extractorData: String? = null,
-        kid: String,
-        key: String,
+        kid: String? = null,
+        key: String? = null,
         uuid: UUID = CLEARKEY_UUID,
-        kty: String = "oct",
+        kty: String? = "oct",
         keyRequestParameters: HashMap<String, String> = hashMapOf(),
+        licenseUrl: String? = null,
```
fire-light42, line 467: This will break the API, so this will be blocked until stable.

KingLucius: I saw that if we are going to wait stable anyway, I can add the parameter like this

LagradOst (approved):

PR #1434, opened 2024-12-22
update target sdk to 35 (android 15), gradle & exoplayer

IndusAryan:


fire-light42 (commented):
Looks fine, but will test before merge 👍

PR #1435, opened 2024-12-25
Fix subtitle deletion for scoped storage

Luna712:
The current way only works for non scoped storage directories therefore it doesn't work for the default Downloads directory. Using `context.contentResolver.delete` *should* make it work for scoped storage directories as well. We start with `SafeFile` and fallback to `contentResolver` to ensure both `file://` and `content://` schemes work.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/SubtitleUtils.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/SubtitleUtils.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.net.Uri
 import com.lagradost.api.Log
 import com.lagradost.cloudstream3.utils.VideoDownloadManager.getFolder
+import com.lagradost.safefile.SafeFile
 
 object SubtitleUtils {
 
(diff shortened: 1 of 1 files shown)


LagradOst:
Please make a PR to my upstream github instead, as this is not the only instance where we delete stuff https://github.com/LagradOst/SafeFile

Luna712:
> Please make a PR to my upstream github instead, as this is not the only instance where we delete stuff https://github.com/LagradOst/SafeFile

@LagradOst 

As far as I could tell this same thing is done elsewhere within this app here, no other instances where we actually delete things that use URI that doesn't do it like this, as the issue is with URI deletions being screwed with with scoped storage,

Luna712:
Also this doesn't have anything to do with SafeFile/UniFile (as contentResolver is different) so I figured it was not appropriate to do there which is why I did it here in the first place. I guess I can though if you absolutely want me to?

LagradOst:
Internally SafeFile is a wrapper to avoid using contentResolver explicitly as that was only relevant in later Android vers. In this case it is very important to fix this issue in SafeFile itself, as SafeFile *should* always work. It is supposed to be api agnostic, to solve these exact issues with the File hell of android.

LagradOst:
If you look at the src code, https://github.com/LagradOst/SafeFile/blob/edd10ad2e620c84fd76f90103b08f5ccbfbe1bdb/library/src/main/java/com/lagradost/safefile/MediaFile.kt#L347 SafeFile should call the exact same function, however it instead uses the UniFile to handle the uri https://github.com/LagradOst/SafeFile/blob/edd10ad2e620c84fd76f90103b08f5ccbfbe1bdb/library/src/main/java/com/lagradost/safefile/SafeFile.kt#L235C20-L235C34

Luna712:
That makes a ton of sense now, thanks!

PR #1438, opened 2024-12-27
Add go to downloads button in home if there is no connection

Luna712:


app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
```diff
@@ -590,7 +592,16 @@ class HomeFragment : Fragment() {
 
                     is Resource.Failure -> {
                         homeLoadingShimmer.stopShimmer()
-                        resultErrorText.text = data.errorString
+                        var errorString =  data.errorString
+                        if (context?.isNetworkAvailable() == false) {
```
fire-light42, line 596: Just use isNetworkError or extend the Resource.Failure and check the exception type of the error. This is because isNetworkAvailable is not always correct.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
Luna712: That actually seems to be more inaccurate though. IE if DNS fails but you actually have connection it is the same error as it is a network error, whereas what I currently use does it only if device has no service.

fire-light42:
Cant you just always show the "go to downloads instead" button, no matter the error?

Luna712:
> Cant you just always show the "go to downloads instead" button, no matter the error?

Could but what I wanted to avoid is the open in browser button if no connection as well.

app/src/main/res/layout/fragment_home.xml
```diff
@@ -177,6 +177,19 @@
             android:text="@string/result_open_in_browser"
             app:icon="@drawable/ic_baseline_public_24" />
 
+        <com.google.android.material.button.MaterialButton
```
fire-light42, line 180: This fails `testAllLayouts<FragmentHomeBinding>(activity, R.layout.fragment_home_tv, R.layout.fragment_home)` test, and should be implemented into the tv layout as well.

fire-light42: Otherwise we have no UI for tv at all.

fire-light42 (changes requested):

fire-light42:
I pushed this PR with minor modifications here: https://github.com/recloudstream/cloudstream/commit/244484f609e56d7b1bf95ef0855392ff5f54fb80

❤️

PR #1439, opened 2024-12-28
Fix path for AsianDrama downloads

Luna712:
It is defined as episode based and saves in the format of Episode X but not in a series sub folder which means it can easily conflict if you have 2 like that etc...


fire-light42:
Makes sense

PR #1440, opened 2024-12-29
Add StringUtils

Luna712:
This may be completely useless (but did plan to use it myself) so feel free to close if you feel it is absolutely unnecessary. This is meant for use by extensions for URI manipulation since the same method seems to be used across many providers and extractors ([search](https://github.com/search?q=cloudstream+String.encodeUri+language%3AKotlin&type=code)), so these methods were just meant to streamline that.

I also plan to potential expand it with more String methods later which is why it is StringX not URIX.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/StringUtils.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/StringUtils.kt
@@ -1,0 +1,14 @@
+package com.lagradost.cloudstream3.utils
+
+import java.net.URLDecoder
+import java.net.URLEncoder
+
+object StringUtils {
+    fun String.encodeUri(): String {
+        URLEncoder.encode(this, "UTF-8")
+    }
+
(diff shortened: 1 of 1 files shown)


fire-light42:
Why not just place it in AppUtils?

Luna712:
> Why not just place it in AppUtils?

I thought about it but figured it was cleaner to do a new one here, more clear that it is available and more String related methods could be added here later on.

IndusAryan:
you can put in text utils instead

https://github.com/recloudstream/cloudstream/blob/master/app%2Fsrc%2Fmain%2Fjava%2Fcom%2Flagradost%2Fcloudstream3%2Futils%2FTextUtil.kt

Luna712:
> you can put in text utils instead
> 
> https://github.com/recloudstream/cloudstream/blob/master/app%2Fsrc%2Fmain%2Fjava%2Fcom%2Flagradost%2Fcloudstream3%2Futils%2FTextUtil.kt

It needs to be library for extensions.

PR #1442, opened 2024-12-29
Add new TVTypes for generic Audio and Podcast

Luna712:
I also added `isAudioType` for an eventual change to player to better accommodate audio-only playback.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -681,8 +681,11 @@ enum class TvType(value: Int?) {
     Music(13),
     AudioBook(14),
 
-    /** Wont load the built in player, make your own interaction */
+    /** Won't load the built in player, make your own interaction */
     CustomMedia(15),
+
+    Audio(16),
+    Podcast(17),
 }
 
 public enum class AutoDownloadMode(val value: Int) {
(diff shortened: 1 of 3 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -681,8 +681,11 @@ enum class TvType(value: Int?) {
     Music(13),
     AudioBook(14),
 
-    /** Wont load the built in player, make your own interaction */
+    /** Won't load the built in player, make your own interaction */
     CustomMedia(15),
+
+    Audio(16),
```
fire-light42, line 687: We have music and audiobook, what exactly is Audio?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -698,9 +701,22 @@ public enum class AutoDownloadMode(val value: Int) {
     }
 }
 
-// IN CASE OF FUTURE ANIME MOVIE OR SMTH
 fun TvType.isMovieType(): Boolean {
-    return this == TvType.Movie || this == TvType.AnimeMovie || this == TvType.Torrent || this == TvType.Live
+    return this in listOf(
```
fire-light42, line 705: See other PR

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Something that doesnt cover any of it. My main reason was for https://github.com/Luna712/Luna712-CloudStream-Extensions/blob/6f45efaa365c9cf3a65f73a848d5cf2e10820fbd/InternetArchiveProvider/src/main/kotlin/com/luna712/InternetArchiveProvider.kt#L256-L257 where no other type other than a generic Audio actually makes true sense there. It isn't absolutely necessary by any means though so I suppose I can take it out if you really want?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: I *think* this is fixed now

PR #1444, opened 2024-12-29
Add volume booster above 100%

### LINKED ISSUE #1086: Allow increasing volume over 100%

A few videos have low volume. A toggle could be added to the player settings that removed the 100% limit. It would make the user experience better.

### Other details

Vlc cannot be used for most videos , because the https stream gets blocked on it. What i'm trying to say is "Just use vlc" is not a valid option.

### Acknowledgements

Luna712:
Resolves #1086


Luna712:
This *should* be ready for review now...

fire-light42 (changes requested):
I have only tested this briefly so more bugs might be found later. However at the moment you have two critical bugs that I require you to fix.
1. The showtoast is spammed when dragging over from under 100% to 100%. This might not seem like an issue to you depending on your OS, but for me it flickers. 
2. The color of the left bar even when under 100% is slightly more yellow/orange then the right brightness bar.

Luna712:
> I have only tested this briefly so more bugs might be found later. However at the moment you have two critical bugs that I require you to fix.
> 
> 1. The showtoast is spammed when dragging over from under 100% to 100%. This might not seem like an issue to you depending on your OS, but for me it flickers.
> 2. The color of the left bar even when under 100% is slightly more yellow/orange then the right brightness bar.

@fire-light42 

Both those *should* be fixed although I couldn't really reproduce the first one.

fire-light42:
Issue two is fixed, but issue one remain. I recommend to look at the log in android studio as showToast should print out repeatedly.

fire-light42:
While issue two is fixed, I would much rather you make the background progress of the orange bar invisible so it does not change the white background color at all above 100%.

Luna712:
> While issue two is fixed, I would much rather you make the background progress of the orange bar invisible so it does not change the white background color at all above 100%.

I tried numerous ways to do this and nothing seemed to work. Only thing that did work was a new drawable but I don't want to do that here,

Luna712:
> Issue two is fixed, but issue one remain. I recommend to look at the log in android studio as showToast should print out repeatedly.

I decided to just have toasts show once as they get annoying anyway if you change volume a lot, and I think you can get the point after the first time anyway.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1307,6 +1316,156 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+                }
+                else -> {
+                    // Volume spans base and first boost range
+                    level1ProgressBar.progress = 100
+                    level2ProgressBar.progress = currentVolumePercentage - 100
+                    level2ProgressBar.isVisible = true
+                }
+            }
+
+            val level1Color = Color.WHITE
```
fire-light42, line 1408: This entire part can be removed by adding `android:progressTint="@color/colorPrimaryOrange"`

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1273,6 +1270,18 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
                                     return true
                                 }
                             }
+
+                            KeyEvent.KEYCODE_VOLUME_DOWN,
+                            KeyEvent.KEYCODE_VOLUME_UP -> {
+                                if (isLayout(PHONE or EMULATOR)) {
```
fire-light42, line 1276: Why not tv, can you please explain why only the tv layout does not get volume boosting?

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -981,7 +997,10 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
                             val maxVolume =
                                 audioManager.getStreamMaxVolume(AudioManager.STREAM_MUSIC)
 
-                            currentRequestedVolume = currentVolume.toFloat() / maxVolume.toFloat()
+                            if (currentRequestedVolume == 0.0f) {
```
fire-light42, line 1000: This can cause a very bad experience for the user, and is why it always queried in the old code. To exemplify this, imagine that the user watch a movie on the max volume, pauses the app, and switched to another app. Under the course of a few hours lowers the volume, then comes back. 

In this case the volume would still be low, however the moment any volume is changed by only an single step the movie is back at maximum volume.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1307,6 +1316,156 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+            }
+            // Show the progress bar for 1.5 seconds
+            postDelayed(progressBarLeftHideRunnable, 1500)
+        }
+
+        // Max volume percentage including boost
+        var maxVolumePercentage = 200
+        if (verticalAddition >= 0f) {
+            val nextVolume = currentRequestedVolume + verticalAddition
+            if (isAdjustingVolume && currentRequestedVolume <= 1.0f && nextVolume >= 1.0f) {
```
fire-light42, line 1355: `isAdjustingVolume` just feels like a bad solution for this.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1307,6 +1316,156 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+        // Apply loudness enhancer for volumes > 100%
+        if (currentRequestedVolume > 1.0f) {
+            val boostFactor = ((currentRequestedVolume - 1.0f) * 1000).toInt()
+
+            // Show the toast only the first time the volume exceeds 100%
+            // or after it drops below 100% and goes above again.
+            // We only do this if the volume buttons are pressed as
+            // we handle sliding a bit different.
+            if (!hasShownVolumeToast && isVolumeUp && currentVolume < maxVolume) {
+                showToast(R.string.volume_exceeded_100)
```
fire-light42, line 1450: This does not show up if we start a video at exactly 100% volume.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: Because TVs don't need it as much and a lot might not support it. In essence it doesn't do anything and instead would cause the volume bar used on phones to appear on TV when overriding the volume key yet sometimes it doesn't do anything different. So decided to just disable it here.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: If it is I will probably just abandon this as I spent probably 4-5 hours trying to find some better solution lol. I don't have it in me to find another solution again unless you have some suggestion?

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: Whoops lol. That was a leftover from an old way this was done.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
fire-light42: If so, please note this with a comment.

fire-light42: Also, if some devices do not support this, is there a possibility to query this support? Then disable in the case of no support.

Luna712: There probably is I just didn't think it was all that useful on TVs so just disabled it on the layout. I will add a comment though.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
fire-light42: Well if you only want to have this lock feature, then just add a variable called `isVolumeLocked = currentVolumePercentage < 100.0f` or similar at the start of any swipe. 

Additionally with this you could reset hasShownVolumeToast by `if (isVolumeLocked) { hasShownVolumeToast = false; }` in the very next line to allow for more then 1 toast per player instance.

Luna712: That's a great idea actually.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: I *think* this should be fixed...

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: Now done.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: I *think* I fixed this issue.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: I added a comment for this now.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
Luna712: Fixed

fire-light42 (changes requested):
This unfortunately still have some bugs, namely that swiping is inconsistent with pressing the volume button. For example, you cant go above 100% with the vol button, and will simply be greeted with the "slide up again" toast repeatedly, but this does not happen when sliding. 

Additionally, if we go outside the application, lower the volume then come back, sliding does not affect the volume the same as pressing the vol button. I would personally reset the boost if the volume in onResume does not match up with the last requested volume in onPause.

fire-light42:
@Luna712 Do you need help finishing the PR?

Luna712:
> @Luna712 Do you need help finishing the PR?

@fire-light42 

So sorry I got very busy IRL if someone wants to take this over and finish it I'm absolutely okay with that otherwise I will finish it if/when I have time *eventually*

fire-light42 (approved):

PR #1445, opened 2024-12-29
PluginManager: capitalize folder and rename constant

Luna712:
The plugins directory already creates in the capital version of the folder, just rename it to be capital here for consistency, and to be consistent with the name of the app

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
@@ -163,10 +163,10 @@ object PluginManager {
         return getKey(PLUGINS_KEY_LOCAL) ?: emptyArray()
     }
 
-    private val CLOUD_STREAM_FOLDER =
-        Environment.getExternalStorageDirectory().absolutePath + "/Cloudstream3/"
+    private val CLOUDSTREAM3_FOLDER =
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -163,10 +163,10 @@ object PluginManager {
         return getKey(PLUGINS_KEY_LOCAL) ?: emptyArray()
     }
 
-    private val CLOUD_STREAM_FOLDER =
-        Environment.getExternalStorageDirectory().absolutePath + "/Cloudstream3/"
+    private val CLOUDSTREAM3_FOLDER =
+        Environment.getExternalStorageDirectory().absolutePath + "/CloudStream3/"
```
fire-light42, line 167: "Linux file system treats file and directory names as case-sensitive" so this will break all current plugins that are located in "Cloudstream3". Loop all folders that match Cloudstream3 in a case insensitive manner if you want to solve this.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Luna712: Makes sense.

Luna712:
I may do it later but for not I don't intend to put to much extra work into this minor change.

PR #1447, opened 2024-12-30
Add extractor for archive.org

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/InternetArchive.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    override suspend fun getUrl(
+        url: String,
+        referer: String?,
+        subtitleCallback: (SubtitleFile) -> Unit,
+        callback: (ExtractorLink) -> Unit
+    ) {
+        val document = archivedItems[url] ?: run {
+            try {
+                val doc = Jsoup.connect(url).get()
```
fire-light42, line 35: Is there a reason to use `Jsoup.connect(url).get()` as opposed to `app.get(url).document` ? 
Using the built in network client benefits from user settings (DNS) and caching.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/InternetArchive.kt
```diff
@@ -0,0 +1,112 @@
...
+import com.lagradost.cloudstream3.mvvm.logError
+import com.lagradost.cloudstream3.utils.ExtractorApi
+import com.lagradost.cloudstream3.utils.ExtractorLink
+import com.lagradost.cloudstream3.utils.Qualities
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import java.net.URLDecoder
+
+open class InternetArchive : ExtractorApi() {
+    override val mainUrl = "https://archive.org"
```
fire-light42, line 13: If archive.org has a stable video url scheme it would be beneficial to add a comment with an example video for future testing. Many already merged extractors are untestable due to unknown video urls. This is of course not required for merge.

fire-light42 (changes requested):
Code looks good, but `Jsoup.connect` preferably needs to be changed or justified (in a code comment) before merge.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/InternetArchive.kt
Luna712: I can't remember why now lol. I created the extractor about 6 months ago before pushing here from where it was originally in my extension, and now I forgot why. I can try the other way.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/InternetArchive.kt
Luna712: Done

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/InternetArchive.kt
Luna712: Done

fire-light42 (approved):

PR #1448, opened 2024-12-30
fix: downloads and services affected by android 15 support

IndusAryan:


Luna712:
It still crashes with this patch.

app/src/main/java/com/lagradost/cloudstream3/services/VideoDownloadService.kt
```diff
@@ -34,13 +44,19 @@ class VideoDownloadService : Service() {
             }
         }
 
+        VideoDownloadManager.startServiceForeground(this)
```
fire-light42, line 47: The video download manager is already running using the worker foreground service:
```
Started foreground service Intent { act=ACTION_START_FOREGROUND cmp=com.lagradost.cloudstream3.prerelease.debug/androidx.work.impl.foreground.SystemForegroundService (has extras) }
```
Starting another service (without proper termination) is unnecessary

fire-light42: I will remove personally due to the urgency of a fix

app/src/main/AndroidManifest.xml
```diff
@@ -207,18 +207,25 @@
...
+            android:foregroundServiceType="dataSync"
+            tools:node="merge" />
+
         <activity
             android:name=".ui.ControllerActivity"
             android:exported="false" />
 
         <service
+            android:name=".services.PackageInstallerService"
+            android:enabled="false"
```
fire-light42, line 227: Why disabled?

fire-light42:
Let us hope it works this time

Luna712:
@fire-light42 seems PackageInstaller is still broke:
```
Fatal exception on thread main (2) java.lang.RuntimeException: Unable to stop service com.lagradost.cloudstream3 .services.PackageInstallerService@af1a8d7: java.lang.IllegalArgumentException:

Receiver not registered: com.lagradost cloudstream3.utils. Apkinstaller$installActio nReceiver$1@8356a64
```

PR #1449, opened 2024-12-31
Fix PiP new targetsdk

Luna712:


IndusAryan:
if this will fix it then you can use the registerbroadcastreciever fucntion from intent helper that is added yesterday

PR #1450, opened 2024-12-31
Use FLAG_IMMUTABLE creating pending intent in PackageInstaller

Luna712:
This *might* fix the PackageInstaller but I can't test that until it is live and there is another update to test on after it.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/services/PackageInstallerService.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/services/PackageInstallerService.kt
@@ -29,7 +29,7 @@ class PackageInstallerService : Service() {
 
     private val baseNotification by lazy {
         val flag = if (SDK_INT >= Build.VERSION_CODES.S) {
-            PendingIntent.FLAG_MUTABLE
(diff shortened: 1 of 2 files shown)


Luna712:
@fire-light42

app/src/main/java/com/lagradost/cloudstream3/services/PackageInstallerService.kt
```diff
@@ -29,7 +29,7 @@ class PackageInstallerService : Service() {
 
     private val baseNotification by lazy {
         val flag = if (SDK_INT >= Build.VERSION_CODES.S) {
-            PendingIntent.FLAG_MUTABLE
+            PendingIntent.FLAG_IMMUTABLE
```
Luna712, line 32: Just noting I'm not positive if this one here needs changed also or not.

Luna712: This should work though from examining things. Intent is never changed si IMMUTABLE should be the correct option here, theoretically...

Luna712:
This finally fixed PackageInstaller!

PR #1453, opened 2024-12-31
Chillx Fix

phisher98:


fire-light42 (changes requested):
Dont use android libs in extractors, as the core of cloudstream should be cross platform. If I remember correctly importing android will break the auto build.

phisher98:
@fire-light42 I have removed 
import android.annotation.TargetApi
import android.os.Build

still giving error

phisher98:
@fire-light42 fixed

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Chillx.kt
```diff
@@ -132,45 +111,55 @@ open class Chillx : ExtractorApi() {
...
+            val keyValue = xorKey[i % xorKey.size]  
+            val xorResult = byteValue xor keyValue
+            result.append(xorResult.toChar())
         }
+
+        return result.toString()
+    }
+
+    private suspend fun fetchKey(): String? {
+        return app.get("https://raw.githubusercontent.com/Rowdy-Avocado/multi-keys/refs/heads/keys/index.html")
```
fire-light42, line 160: Cache this value

fire-light42 (commented):
Looking at this, it looks like there is 3 different locations that get the same url. Can you please unify them in a companion object that all extractors that use these keys share? https://github.com/recloudstream/cloudstream/blob/678ddb6b0a1bc1a05c81ae65bf9d03614ce27f74/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/VidSrcTo.kt#L25

phisher98:
@fire-light42 Why its failing its working in local

phisher98:
@fire-light42 Hope this works now

fire-light42 (approved):

PR #1441, opened 2024-12-29
Fix to define Cartoon as episode based

Luna712:
Supposed to be based on https://github.com/recloudstream/cloudstream/blob/282b25b665fd2fe9f7a596e3d6c9067ea540730c/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt#L643

PR #1455, opened 2024-12-31
Bump version to 4.4.3

Luna712:
Just doing to trigger a new prerelease to test PackageInstaller.

Feel free to close if you want to do something else.


Luna712:
@fire-light42

Luna712:
If it's able to and you don't want to merge this #1441 can be merged instead to trigger it though...

PR #1463, opened 2025-01-02
MainAPI: add missing builder methods

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -819,6 +819,25 @@ interface SearchResponse {
     var quality: SearchQuality?
 }
 
+fun MainAPI.newTorrentSearchResponse(
+    name: String,
+    url: String,
+    type: TvType? = null,
+    posterUrl: String? = null,
```
fire-light42, line 826: Do not add any non required fields in the new constructor as that is the initializers job.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -832,6 +851,23 @@ fun MainAPI.newMovieSearchResponse(
     return builder
 }
 
+fun MainAPI.newLiveSearchResponse(
+    name: String,
+    url: String,
+    type: TvType? = null,
```
fire-light42, line 857: No, this should not be nullable.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1431,6 +1466,27 @@ data class TorrentLoadResponse(
     )
 }
 
+suspend fun MainAPI.newTorrentLoadResponse(
+    name: String,
+    url: String,
+    magnet: String? = null,
+    torrent: String? = null,
+    plot: String? = null,
```
fire-light42, line 1474: This is a non required field, remove it.

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: It is a required field.
> No value passed for parameter 'plot'

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: It is a required field.
> No value passed for parameter 'posterUrl'

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: `override var plot: String?` This means that the initializer should handle it.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: I was just using what the default was in LiveSearchResponse, otherwise it unexpectedly changes this from the default to the new default used here. I can change if you want though?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: The keyword is **var** here

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: The name LiveSearchResponse implies that it of type TvType.Live like the other new constructors.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: How do I let the builder use it then without breaking existing constructor by changing it when it errors if the constructor is not passed it directly?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: It is not, all new constructors should be as minimal as possible as it is created to replace the build in constructors. If we do not need that field to function it should not be in the constructor. `posterUrl` is simply a visual aid, but all require a name and url, otherwise they cant function at all.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Changed

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: Pass null

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Fixed

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Well I am an idiot for not thinking of the obvious. Thanks! Fixed.

fire-light42 (approved):

PR #1465, opened 2025-01-03
 Implement click functionality on cast names to initiate a Google search

### LINKED ISSUE #861: Do a Google search when clicking on movie/series Cast with real name

Right now clicking on movie/series cast names do nothing 
I want it to do Google search or open their Wikipedia page or something like that so we can get information on them

### Other details

_No response_

### Acknowledgements

bilal96aslam:
This change will resolve #861 improvement Do a Google search when clicking on movie/series Cast with real name

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ActorAdaptor.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ActorAdaptor.kt
@@ -1,18 +1,19 @@
 package com.lagradost.cloudstream3.ui.result
 
-import android.net.Uri
+import android.app.SearchManager
+import android.content.Intent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.RecyclerView
-import com.lagradost.api.getContext
 import com.lagradost.cloudstream3.ActorData
 import com.lagradost.cloudstream3.ActorRole
 import com.lagradost.cloudstream3.R
 import com.lagradost.cloudstream3.databinding.CastItemBinding
-import com.lagradost.cloudstream3.utils.AppContextUtils.openBrowser
+import com.lagradost.cloudstream3.ui.settings.Globals.PHONE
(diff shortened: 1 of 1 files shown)


AvocadoWasSeized:
This won't work, especially for anime as when you click the actor it swaps to their character. It would also be very annoying for the people who accidentally click on the actors when scrolling down, this has happened to me a lot. I think if this were to be implemented there should be a setting to enable/disable due to these issues, mainly the anime one though.

KingLucius:
Better to use internal search for character related search but this is not supported yet in extensions apis

bilal96aslam:
@AvocadoWasSeized Thanks for your feedback! I've implemented the functionality on long-press instead of click to avoid accidental activations while scrolling. I’ve also tested it for anime cases, and it works similarly to TV/movies. Let me know if there are further improvements you’d suggest!

fire-light42:
> This won't work, especially for anime as when you click the actor it swaps to their character. It would also be very annoying for the people who accidentally click on the actors when scrolling down, this has happened to me a lot. I think if this were to be implemented there should be a setting to enable/disable due to these issues, mainly the anime one though.

This is `LongClick` not click.

fire-light42 (approved):
I am open to merging this pr while we add better actor support. However I am unsure if hard-coding google is the best option. Is there no way to get the default search engine on android?

bilal96aslam:
To avoid hard-coding the Google URL and use the user's default search engine:
itemView.setOnLongClickListener {
    val intent = Intent(Intent.ACTION_WEB_SEARCH).apply {
        putExtra(SearchManager.QUERY, actor.actor.name)
    }
    intent.resolveActivity(itemView.context.packageManager)?.let {
        itemView.context.startActivity(intent)
    } ?: itemView.context.openBrowser(url = "https://www.google.com/search?q=${Uri.encode(actor.actor.name)}")
    true
}
This checks for a web search app and falls back to Google if none is available.

IndusAryan:
I think this should be only available for mobile.

fire-light42:
> I think this should only be available for mobile with view intent of browser instead of url.

@bilal96aslam I would want you to both only limit this to mobile (use for example islayout) as well as adding Intent.ACTION_WEB_SEARCH before I merge this 👍

PR #1469, opened 2025-01-05
Migrate to non-transitive R classes

Luna712:


fire-light42:
Lets hope this does not break extensions.

PR #1472, opened 2025-01-05
Use MainAPI to get folder prefix

Luna712:
Rather than having to keep getFolder() and isEpisodeBased in sync, we can automate using a sub-folder if TvType is episode-nased. Also adds some documentation to `TvType?.isEpisodeBased()` and new `TvType.getFolderPrefix()`


fire-light42:
I will have to double check this, but if we want to refactor this I would suggest the viewmodel to decide if it is episode based, because it has additional context like `is EpisodeResponse`. https://github.com/recloudstream/cloudstream/blob/9c31d0a61e662ed446ff536bd31f6ab39cfccee6/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt#L1320

Luna712:
> I will have to double check this, but if we want to refactor this I would suggest the viewmodel to decide if it is episode based, because it has additional context like `is EpisodeResponse`.
> 
> https://github.com/recloudstream/cloudstream/blob/9c31d0a61e662ed446ff536bd31f6ab39cfccee6/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt#L1320

Only LoadResponse version has is EpisodeResponse. I did it like this, at least at the moment just so it kept the *exact* same behavior as before but I can change it if you want?

PR #1473, opened 2025-01-06
Update Turkish Translation

KerimDemirkaynak:
In Turkish, plural forms are generally not used with numbers. For example, "10 Bölümler" ("10 Episodes") is incorrect, while "10 Bölüm" ("10 Episode") is the correct form. The "episodes" string value was updated from "Bölümler" to "Bölüm" to align with Turkish grammar rules. This change ensures linguistic accuracy and improves the user interface for Turkish-speaking users.


fire-light42:
Please use https://hosted.weblate.org/engage/cloudstream/ instead

keyiflerolsun:
https://hosted.weblate.org/projects/cloudstream/-/tr/

PR #1478, opened 2025-01-08
Fix many build warnings/deprecations

Luna712:
Also suppresses deprecations that have no replacement/that we can't ever fix while maintaining current functionality support.

Also only deprecate constructors in MainAPI, we don't want type hints to warn.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/JsUnpacker.kt
```diff
@@ -56,12 +56,12 @@ class JsUnpacker(packedJS: String?) {
...
                 while (m.find()) {
                     val word = m.group(0)
-                    val x = unbase.unbase(word)
+                    val x = if (word == null) 0 else unbase.unbase(word)
                     var value: String? = null
                     if (x < symtab.size && x >= 0) {
                         value = symtab[x]
                     }
-                    if (value != null && value.isNotEmpty()) {
+                    if (!value.isNullOrEmpty() && !word.isNullOrEmpty()) {
```
fire-light42, line 64: I have not checked this in the IDE yet, but it looks like you are adding extra logic by `!word.isNullOrEmpty()`

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/JsUnpacker.kt
```diff
@@ -26,25 +26,25 @@ class JsUnpacker(packedJS: String?) {
...
                 try {
-                    radix = radixStr.toInt()
-                } catch (e: Exception) {
+                    radix = radixStr?.toInt() ?: radix
+                } catch (_: Exception) {
                 }
                 try {
-                    count = countStr.toInt()
-                } catch (e: Exception) {
+                    count = countStr?.toInt() ?: 0
```
fire-light42, line 46: If you are going to refactor like this, use toIntOrNull

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -814,17 +816,17 @@ class CS3IPlayer : IPlayer {
                     factory.createMediaSource(item.mediaItem)
                 }
             } else {
-                val source = ConcatenatingMediaSource()
+                val source = ConcatenatingMediaSource2.Builder()
                 mediaItemSlices.map { item ->
-                    source.addMediaSource(
-                        // The duration MUST be known for it to work properly, see https://github.com/google/ExoPlayer/issues/4727
+                    source.add(
+                        // The duration MUST be known for it to work properly
```
fire-light42, line 822: Why remove the comment?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/JsUnpacker.kt
Luna712: Because it warned about word being nullable and figured it could just align with value check exactly

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
Luna712: Accident i originally removed this and did something else and comment got lost in that... my bad...

Luna712: Fixed

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/JsUnpacker.kt
Luna712: Done

app/src/main/java/com/lagradost/cloudstream3/utils/AppContextUtils.kt
```diff
@@ -848,20 +874,27 @@ object AppContextUtils {
         val isCastApiAvailable =
             GoogleApiAvailability.getInstance()
                 .isGooglePlayServicesAvailable(applicationContext) == ConnectionResult.SUCCESS
+
+        val executor: Executor = Executors.newSingleThreadExecutor()
+
         try {
-            applicationContext?.let { CastContext.getSharedInstance(it) }
+            applicationContext?.let {
+                CastContext.getSharedInstance(it, executor)
```
fire-light42, line 882: I have not tested but I belive this is a bug. The old code ran the code to check for an exception, but this uses an executor. This means that the future might not get resolved to throw the error that we want to catch.

fire-light42 (changes requested):
Untested, but most of it seems to be in order.

app/src/main/java/com/lagradost/cloudstream3/utils/AppContextUtils.kt
Luna712: I have an idea of but may not be entirely sure what you mean. But I will take a look and confirm if we need to handle this differently if you want... I did test basically every aspect of the app for this (whether the change directly touches it or not) from downloads to fcast to normal cast etc... and didn't find any issues though but like I've seen before other api levels can definitely have different issues so worth looking if this needs to be handled a different way now with this.

fire-light42: Imagine it like this, ` CastContext.getSharedInstance(it) ` runs the code directly, so if it cant do something it will throw an error. However `CastContext.getSharedInstance(it, executor)` instead returns an Task, this means that after this function has returned it has not actually done any work. A task is simply a lazy-ish calculation, that you can await to completion to get a result. Imagine it like a more complicated way to do suspended in java.  

Right now we are interested if getSharedInstance throws an error, and if none of the work is done to throw an error then no error can be thrown.

Luna712: I'm kinda confused though because why do we need to catch an error at all if it's not going to cause the app to crash at all, isn't the fact it wouldn't throw an exception a good thing? I am still learning a bit of how all the executor stuff works so forgive me if I am being a total idiot with this...

fire-light42: The entire point of this function is to catch the Exception, because if it throws anything it will return isCastApiAvailable as false. We want to check if the function throws anything because we then use that as a condition. 

isCastApiAvailable = getSharedInstance does not throw any exception

Luna712: Oh that makes sense.

Luna712: According to https://developers.google.com/android/reference/com/google/android/gms/cast/framework/CastContext#public-static-taskcastcontext-getsharedinstance-context-context,-executor-executor this still throws the same error for the same reasons unless I am missing something?

fire-light42: It is still a task, await it properly with `.result`.

Luna712: So should I add an OnFailureListener or something with .result connected check? Still a bit confused about this all. I apologize.

fire-light42: Just do `@Suppress("DEPRECATION")` or add `.result`

Luna712:
@fire-light42 I think this is done now. Also I made only constructors in MainAPI deprecated as otherwise it warns about helper methods and type hints which we don't want.

app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
```diff
@@ -255,10 +257,12 @@ object CommonActivity {
                 try {
                     enterPictureInPictureMode(PictureInPictureParams.Builder().build())
                 } catch (e: Exception) {
-                    enterPictureInPictureMode()
+                    Log.e(TAG, "Failed to enter picture in picture mode.")
```
fire-light42, line 260: Just found this now, but why? We specifically call `enterPictureInPictureMode()` if `enterPictureInPictureMode(parms)` fails and you remove the backup function? It is catched anyway, so why remove it.

fire-light42 (changes requested):
Also check out https://stackoverflow.com/questions/52202810/how-does-deprecated-replacewith-work-for-kotlin-in-intellij

app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
Luna712: I did it because it's deprecated so I didn't think we should be using it even in a catch without a version check, which is done below. So instead I made it log that it fails so we can rather have reports if it's now working and fix underlying cause rather than a fallback to a deprecated method in versions where it is actually deprecated. I can change it if you want I just thought it was better not to rely on deprecated methods if it's avoidable at all which I saw this as avoidable.

fire-light42: I would actually prefer using the deprecated version as a backup to not using it at all.

Luna712: Done

Luna712:
@fire-light42 is this good now or does it need something else (code review wise, I understand if it can't be merged as not tested by you yet)?

fire-light42 (approved):

PR #1479, opened 2025-01-08
refactor: make app class proper & remove unused libs

IndusAryan:


fire-light42:
Can you please check if any plugins are using `AcraApplication.Companion` in any way. Such as `setKey` or `context`. Because if they are we should also add a backwards compatibility helper.

CranberrySoup:
AcraApplication.Companion.setKey is used in plugins, at least [here](https://github.com/CranberrySoup/AniyomiCompat/blob/master/app/src/main/java/com/lagradost/aniyomicompat/AniyomiPlugin.kt#L17-L21)

IndusAryan:
I looked at some popular extensions, they don't seem to be using acra imports, so changing the imports in aniyomi compat itself after merging would be good for everyone.

Luna712:
> I looked at some popular extensions, they don't seem to be using acra imports, so changing the imports in aniyomi compat itself after merging would be good for everyone.

Can't we just do something like `class AcraApplication : CloudStreamApp()` as a temporary backcompat wrapper that can be removed after plugins are updated as to not break any plugins during the transition phase? Maybe not exactly like that but I think there should be at least temporary backcompat.

IndusAryan:
changing compatibility 2 times in 2 updates will anyway crash in one pre release and this is the only extension imo which uses this.

Luna712:
> changing compatibility 2 times in 2 updates will anyway crash in one pre release and this is the only extension imo which uses this.

There are many more.

Luna712:
https://github.com/search?q=cloudstream3+acraapplication+-path%3Acom%2Flagradost+-path%3Agradle+language%3AKotlin&type=code

Luna712:
> changing compatibility 2 times in 2 updates will anyway crash in one pre release and this is the only extension imo which uses this.

That's why you do it with stable releases. You keep back compat, update all extensions then finally remove back compat when all extensions update it which may be awhile but there is not any real disadvantage of leaving the back compat code for awhile either.

Luna712:
We accidentally broke so many extensions with the build stuff lately (which we've been working on  fixing) I'm especially not happy if we end up purposely breaking many extensions by changing the main class they would use for setting plugin preferences etc... in the settings popup.

IndusAryan:
We can do backCompat wrapper likely, but sorry these extensions are dead, unmaintained and unused except ultimaStorage and aniwave. But ultima is huge so it may rely on wrapper but as i said it will crash on many devices once wrapper is removed either in stable or pre and things are broken now already so maybe best time is this only.

IndusAryan:
@fire-light42 what to do here, afaik ULTIMA is the only public extension that uses set key

Luna712:
@IndusAryan is this still planned to be finished? Otherwise I was gonna work on it, but to maintain some back compat with deprecations for now if that is still necessary. My main reason is to remove Acra when we aren't using it (unless we don't want to remove it, @fire-light42?).

IndusAryan:
> @IndusAryan is this still planned
Hi @Luna712 , Unfortunately I rarely get time to do anything due to my work that is in chaos since past few months due to a massive project with a short timeline, You can continue this and my other PRs as well.
I will close them when you open their counterparts.

IndusAryan:
> > @IndusAryan is this still planned

Hi @Luna712 , Unfortunately I rarely get time to do anything due to my work that is in chaos since past few months due to a massive project with a short timeline, You can continue this and my other PRs as well.
I will close them when you open their counterparts.

PR #1481, opened 2025-01-09
add DoH DNS servers

salman731:
DNS.sb
Canadian Shield
Next DNS

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/network/DohProviders.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/network/DohProviders.kt
@@ -89,7 +89,7 @@ fun OkHttpClient.Builder.addQuad9Dns() = (
 fun OkHttpClient.Builder.addDnsSbDns() = (
         addGenericDns(
             "https://doh.dns.sb/dns-query",
-
+            //https://dns.sb/guide/
             listOf(
                 "185.222.222.222",
                 "45.11.45.11",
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/network/DohProviders.kt
```diff
@@ -84,4 +84,36 @@ fun OkHttpClient.Builder.addQuad9Dns() = (
...
+
+            listOf(
+                "149.112.121.10",
+                "149.112.122.10",
+            )
+        ))
+
+fun OkHttpClient.Builder.addNextDns() = (
+        addGenericDns(
+            "https://dns.nextdns.io/457cc1",
```
CranberrySoup, line 111: This does not look like a generic domain, is /457cc1 public, or is it tied to an account?

app/src/main/java/com/lagradost/cloudstream3/network/DohProviders.kt
```diff
@@ -84,4 +84,36 @@ fun OkHttpClient.Builder.addQuad9Dns() = (
...
-    ))
\ No newline at end of file
+    ))
+
+fun OkHttpClient.Builder.addDnsSbDns() = (
+        addGenericDns(
+            "https://doh.dns.sb/dns-query",
+
+            listOf(
+                "185.222.222.222",
```
CranberrySoup, line 94: Please add comments where these IPs are found for easier maintenance and verification

app/src/main/java/com/lagradost/cloudstream3/network/DohProviders.kt
CranberrySoup: Having it in someones account would allow for logs and unwanted traffic controls

salman731: It is tied to account. I have to remove

app/src/main/java/com/lagradost/cloudstream3/network/DohProviders.kt
salman731: okay

Datj0000:
please use nextdns public https://dns.nextdns.io

PR #1488, opened 2025-01-14
Add Not Interested WatchType

Luna712:


Luna712:
NOTE: this was not yet tested at all I just needed to push it while I reset my Android Studio so I didn't loose it.

fire-light42:
1. Why?
2. As far as I know this will break stuff. This is because internal Id is only meant to act as an unchanging serializable identifier, and not sorting order.

Luna712:
> 1. Why?
> 2. As far as I know this will break stuff. This is because internal Id is only meant to act as an unchanging serializable identifier, and not sorting order.

Indeed but I like it to be in order, was a personal preference but can change if you want... Changing the ID for none doesn't break anything though since none is never stored only removed when changing to none. In order to avoid breaking anything I just used -2 for the new one so that it doesn't matter. As for the answer to why, because it is different then dropped and adding things to dropped isn't ideal when you don't even want to start something (which I end up doing a lot)

fire-light42:
As far as I know, no syncing service currently offered has not interested as a watch type. If I remember correctly rewatching gets remapped to watching on some services where rewatching does not exists, however not interested has no replacement. 

Additionally for the build in lib, we do not currently offer any blocking service, so just tagging it not interested without that backend code seams wasteful and useless. So as of this moment, I do not see a need nor a good reason for this inclusion to the watch types.

Luna712:
> As far as I know, no syncing service currently offered has not interested as a watch type. If I remember correctly rewatching gets remapped to watching on some services where rewatching does not exists, however not interested has no replacement.
> 
> Additionally for the build in lib, we do not currently offer any blocking service, so just tagging it not interested without that backend code seams wasteful and useless. So as of this moment, I do not see a need nor a good reason for this inclusion to the watch types.

That's why I didn't add it to any to be used. However it will be added with MyDramaList eventually so while adding to local I added to SyncApi too since it could be used there when that is done which KingLucius said they were working on also.

I don't know why you say this doesn't do anything though it adds not interested to the local library? Which has all the implementation like showing it on both the home page and in the library and in the bookmark dialog options with full handling, so what do you mean it doesn't have the backend code? I'm confused, I apologize if I am misunderstanding you.

Luna712:
I can change the ids though if that is what you don't like, the using -2 has the ID for this.

fire-light42:
> I don't know why you say this doesn't do anything though it adds not interested to the local library? Which has all the implementation like showing it on both the home page and in the library and in the bookmark dialog options with full handling, so what do you mean it doesn't have the backend code? I'm confused, I apologize if I am misunderstanding you.

I mean like an anti-favorite/block feature, that removes an entry from the homepage or search.

Luna712:
> > I don't know why you say this doesn't do anything though it adds not interested to the local library? Which has all the implementation like showing it on both the home page and in the library and in the bookmark dialog options with full handling, so what do you mean it doesn't have the backend code? I'm confused, I apologize if I am misunderstanding you.
> 
> I mean like an anti-favorite/block feature, that removes an entry from the homepage or search.

I'm just extremely confused lol. That's not what this is, Not Interested is not like Favorites where you can have in another watchlist too, it's a valid single option that none of the other, Watching, Completed, Dropped, etc... would appear in. Additionally it's a bookmark list you'd want to keep track of and find sometimes so can also appear in library. Although I suppose that's actually not a bad idea at all to implement but that's actually a different thing I never even thought of and different then what this was meant to do.

fire-light42:
> I'm just extremely confused lol. That's not what this is, Not Interested is not like Favorites where you can have in another watchlist too, it's a valid single option that none of the other, Watching, Completed, Dropped, etc... would appear in. Additionally it's a bookmark list you'd want to keep track of and find sometimes so can also appear in library. Although I suppose that's actually not a bad idea at all to implement but that's actually a different thing I never even thought of and different then what this was meant to do.

Yes, that is understandable. However what my main argument was, is that without a feature attached the notation itself would be useless. If you are not interested in something, why bother bookmarking it? I simply do not understand the motivation behind adding this feature.

Luna712:
> > I'm just extremely confused lol. That's not what this is, Not Interested is not like Favorites where you can have in another watchlist too, it's a valid single option that none of the other, Watching, Completed, Dropped, etc... would appear in. Additionally it's a bookmark list you'd want to keep track of and find sometimes so can also appear in library. Although I suppose that's actually not a bad idea at all to implement but that's actually a different thing I never even thought of and different then what this was meant to do.
> 
> Yes, that is understandable. However what my main argument was, is that without a feature attached the notation itself would be useless. If you are not interested in something, why bother bookmarking it? I simply do not understand the motivation behind adding this feature.

To track it so you don't keep going back to the same thing. I do that a lot. I currently use Dropped for this purpose, but it doesn't really fit there either as that has another use as well.

PR #1490, opened 2025-01-15
Migrate dokka to 2.0.0

Luna712:
This was tested and works locally. This fixes a lot of compatability issues with kotlin 2 also, such as it currently giving `<Error class: unknown class>` for things like `listOf()`, `mutableListOf()`, and `typealias()`. It also adds support for configuration cache in Dokka (although still may need a few changes on our end for it which I may try to do in a follow-up patch) which will be required in a future gradle release, as well as has many improvements to the actual generated HTML docs.

### DIFF
--- a/.github/workflows/generate_dokka.yml
+++ b/.github/workflows/generate_dokka.yml
@@ -55,7 +55,7 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE/src/
           chmod +x gradlew
-          ./gradlew docs:dokkaHtml
+          ./gradlew docs:dokkaGeneratePublicationHtml
 
       - name: Copy Dokka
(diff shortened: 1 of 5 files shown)


Luna712:
I have one request with this, if it causes dokka build to fail (it shouldn't though as I tested locally) please don't revert and give me a chance to fix it which I will do as soon as I notice. Thanks!

fire-light42:
Thank you for making sure the documentation is working, but please motivate any gradle updates in the future.

Luna712:
> Thank you for making sure the documentation is working, but please motivate any gradle updates in the future.

Thank you for merging, but what do you mean "please motivate any gradle updates"? Not to sure what you mean about that...

PR #1493, opened 2025-01-16
Dokka: fix compatibility with multiplatform and configuration cache

Luna712:
Fixes many `<Error class: unknown class>` such as with LiveData and RecyclerView as well as potential conflicts in class names between modules

TODO:

- [x] ~~Fix `<Error class: unknown class>` for DataBinding classes~~
- [x] ~~Fix `<Error class: unknown class>` R class~~
- [x] Fix `<Error class: unknown class>` for some jackson and nicehttp classes in library only
- [x] Fix missing source link in App module
- [x] Fix overall project to be Cloudstream again

I had DataBinding and R fixed locally by adding the directories the generated files are in to sourceRoots, however that means the dokka script action would also need to build the entire app to get there so decided not to push here, We may be able to find a solution later but testing this locally it seems to fix every other remaining `<Error class: unknown class>` issue.

This also significantly simplifies the dokka configuration and removes the need to maintain the `dokkaImplementation` dependencies separately.

Removes libs that were only used in docs as well.

Also, this combined with #1494 adds full support for configuration and build cache across cloudstream entirely.


.github/workflows/generate_dokka.yml
```diff
@@ -41,6 +41,8 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE/dokka/
           rm -rf "./-cloudstream"
```
Luna712, line 43: **NOTE**: This can be removed in a follow-up but it needs to be kept for the next run to remove the old directory in recloudstream/dokka.

PR #1494, opened 2025-01-17
Support configuration cache for getting git commit

Luna712:
This will be required in a future version of gradle so we may as well prepare for it and support it here.

Without this:
```
Starting an external process 'git rev-parse --short HEAD' during configuration time is unsupported.
```
I am not sure if there is another solution here though but this is the only one I managed to get working.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,7 +1,6 @@
 import com.android.build.gradle.internal.cxx.configure.gradleLocalProperties
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
-import java.io.ByteArrayOutputStream
 
 plugins {
     id("com.android.application")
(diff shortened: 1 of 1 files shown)


app/build.gradle.kts
```diff
@@ -13,22 +12,20 @@ val javaTarget = JvmTarget.fromTarget(libs.versions.jvmTarget.get())
...
-
-    return if (exitValue == 0) {
-        output.toString().trim()
-    } else null
+fun getGitCommitHash(): String {
+    val headFile = file("${project.rootDir}/.git/HEAD")
+
+    // Read the commit hash from .git/HEAD
+    return if (headFile.exists()) {
+        val headContent = headFile.readText().trim()
```
fire-light42, line 20: Even if you check with exists(), this may throw an error. I would recommend try catch to not crash the build process.

fire-light42 (changes requested):

app/build.gradle.kts
Luna712: Done

fire-light42 (approved):
Not tested, but looks good 👍

PR #1495, opened 2025-01-17
gradle: enable build and configuration cache

Luna712:
#1493 and #1494 adds full support for this.

This shouldn't break any extensions or anything but the next step is adding support for configuration cache for plugins themselves in the [Cloudstream Gradle Plugin](/recloudstream/gradle).


Luna712:
This should be good to merge now. It's been attempted twice before but this time everything should support it properly.

PR #1496, opened 2025-01-18
Changed YoutubeExtractor

doGior:
Now it uses the an m3u8 playlist from newpipe so the quality should depend on the connection


fire-light42 (changes requested):
This both removes the caching system as well as subtitles. Why???

doGior:
Sorry for the caching system, I didn't understand where those variables were used, but the subtitles are included in the hls stream.

doGior:
The subtitles are included in the hls stream as I said, but I put back the old code commented just in case. Hls streams should also support different audio tracks

fire-light42:
> but I put back the old code commented just in case. Hls streams should also support different audio tracks

No subtitles are not always embedded.

fire-light42 (approved):

PR #1497, opened 2025-01-18
Bump some dependencies

Luna712:
rhino could also be bumped to 1.8.0 but it has many many more changes and needs much more extensive testing so left it out of this PR. I will do it separately after doing much more testing on it.


Luna712:
Do not merge this yet. May have found a serious bug on some devices that causes app to continually crash.

PR #1498, opened 2025-01-21
fix: don't attempt to fix urls that are JSON arrays

Bnyro:
I noticed that some providers pass JSON arrays as data to `loadLinks`, which will be URL-fixed and thus such providers are broken.

Apart from that it can come in handy to pass the stream links directly as a list via JSON arrays instead of wrapping it into a JSON dictionary.


fire-light42:
Very good catch 👍

PR #1509, opened 2025-01-26
Fix isCastApiAvailable

Luna712:


Luna712:
@fire-light42 seems it was inconsistent the other way (sometimes works but more often then not doesn't); this seems to fix it.

PR #1523, opened 2025-01-29
Respect "User-Agent" header in ExtractorLink

KingLucius:
Some sites blocks default user agent value in CS, I have encountered one while creating extension and I couldn't pass another user agent via headers

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -633,16 +633,17 @@ class CS3IPlayer : IPlayer {
         private fun createOnlineSource(link: ExtractorLink): HttpDataSource.Factory {
             val provider = getApiFromNameNull(link.source)
             val interceptor = provider?.getVideoInterceptor(link)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -633,16 +633,17 @@ class CS3IPlayer : IPlayer {
         private fun createOnlineSource(link: ExtractorLink): HttpDataSource.Factory {
             val provider = getApiFromNameNull(link.source)
             val interceptor = provider?.getVideoInterceptor(link)
+            val userAgent = link.headers.getOrDefault("User-Agent", null)
```
fire-light42, line 636: Getting "User-Agent" by this is case sensitive, but headers are not.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
KingLucius: What's your suggestion ? Lower case only or what?

PR #1524, opened 2025-01-29
feat(Extensions): addTrailer with headers support

KingLucius:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1419,6 +1420,17 @@ interface LoadResponse {
             trailers.addAll(newTrailers.map { TrailerData(listOf(it)) })
         }*/
 
+        @Suppress("RedundantSuspendModifier")
+        suspend fun LoadResponse.addTrailer(
```
fire-light42, line 1424: add `@Prerelease`

fire-light42 (changes requested):

fire-light42 (approved):

PR #1527, opened 2025-01-30
Make the plugin part crossplatform

CranberrySoup:
The purpose of this is to make it possible to remove the cloudstream apk dependency from plugins.
```kt
// This can now be removed and replaced with the library dependency
cloudstream("com.lagradost:cloudstream3:pre-release")
```
and within the plugin change it to:
```kt
@CloudstreamPlugin
class TwitchPlugin: BasePlugin() {
    override fun load() {
        // All providers should be added in this manner. Please don't edit the providers list directly.
        registerMainAPI(TwitchProvider())
        registerExtractorAPI(TwitchProvider.TwitchExtractor())
    }
}
```
This commit is fully backwards compatible from my testing. 
No old or new plugins break (unless new plugins use `BasePlugin` on an old app version).

BasePlugin is missing registerVideoClickAction, resources and openSettings, all of which are hard to un-bundle from android. Those will have to be considered in the future.

My aim with this is to make plugins independent from the app with the express purpose of a future PC version.


CranberrySoup:
Note to plugin developers: It is already possible to test plugins locally

```kt
suspend fun main() {
    val results = TwitchProvider().search("test")
    println(results)
}
```
You may need 
```kt
implementation("com.fasterxml.jackson.module:jackson-module-kotlin:2.13.1")
implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.1")
```

PR #1533, opened 2025-01-31
Fix installer on some older devices

Luna712:
Not 100% sure when this starts working however I am 100% sure it works on SDK 34 & 35 so we just start there and use the old behavior on older APIs just to maintain that it will work for certain.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/PackageInstaller.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/PackageInstaller.kt
@@ -111,17 +111,20 @@ class ApkInstaller(private val service: PackageInstallerService) {
                 }
 
             // We must create an explicit intent or it will fail on Android 15+
-            val installIntent = Intent(service, PackageInstallerService::class.java).apply {
-                action = INSTALL_ACTION
+            val installIntent = Intent(
+                service, PackageInstallerService::class.java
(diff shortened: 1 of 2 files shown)


Luna712:
Maybe this isn't needed?

Luna712:
Let's do it just to be safe

app/src/main/java/com/lagradost/cloudstream3/utils/PackageInstaller.kt
```diff
@@ -111,17 +111,20 @@ class ApkInstaller(private val service: PackageInstallerService) {
                 }
 
             // We must create an explicit intent or it will fail on Android 15+
-            val installIntent = Intent(service, PackageInstallerService::class.java).apply {
-                action = INSTALL_ACTION
+            val installIntent = Intent(
```
fire-light42, line 114: Dont overcomplicate this, just use an if else statement instead of takeIf and apply.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/utils/PackageInstaller.kt
Luna712: Done

fire-light42 (approved):

PR #1534, opened 2025-02-01
Add reviews support to API

Luna712:


Luna712:
This has been fully tested to work on mobile, still need to add UI for TV though.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -945,6 +979,36 @@ fun MainAPI.updateUrl(url: String): String {
     }
 }
 
+@Prerelease
+@ConsistentCopyVisibility
+data class UserReview internal constructor(
+    var review: String? = null,
+    var reviewTitle: String? = null,
+    var username: String? = null,
+    var reviewDate: String? = null,
```
fire-light42, line 988: Store date as unixtime or similar.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -945,6 +979,36 @@ fun MainAPI.updateUrl(url: String): String {
...
+    var avatarUrl: String? = null,
```
fire-light42, line 989: + avatarHeaders

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -945,6 +979,36 @@ fun MainAPI.updateUrl(url: String): String {
...
+    var rating: Int? = null,
```
fire-light42, line 990: No documentation on the range here (0-10?) Some sites allow for partial score so this should also allow for scores like 3.2/10.0.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -945,6 +979,36 @@ fun MainAPI.updateUrl(url: String): String {
...
+    var ratings: List<Pair<Int, String>>? = null,
+) {
+    fun new(initializer: UserReview.() -> Unit = {}): UserReview {
+        return UserReview().apply {
```
fire-light42, line 994: What is this? This is unlike any other initializer, and is prone to error when a new field is added.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -519,6 +519,40 @@ abstract class MainAPI {
         throw NotImplementedError()
     }
 
+    open val hasReviews: Boolean = false
+    open suspend fun loadReviews(
+        url: String,
+        page: Int,
+        showSpoilers: Boolean = false
```
fire-light42, line 526: While this api might be functional, we might want to also add sort by or filter by.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -945,6 +979,36 @@ fun MainAPI.updateUrl(url: String): String {
     }
 }
 
+@Prerelease
+@ConsistentCopyVisibility
+data class UserReview internal constructor(
```
fire-light42, line 984: This class does not allow for sites that only have boolean ratings (positive/negative), like steam.

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
```diff
@@ -194,4 +195,19 @@ class APIRepository(val api: MainAPI) {
...
     }
+
+    suspend fun loadReviews(
+        url: String,
+        page: Int,
+        showSpoilers: Boolean = false
+    ): Resource<List<UserReview>> {
+        return safeApiCall {
+            val response = api.loadReviews(url, page, showSpoilers)
+            val reviewsPerPage = 20
```
fire-light42, line 206: What is this?

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
```diff
@@ -194,4 +195,19 @@ class APIRepository(val api: MainAPI) {
...
         }
     }
+
+    suspend fun loadReviews(
+        url: String,
+        page: Int,
+        showSpoilers: Boolean = false
+    ): Resource<List<UserReview>> {
+        return safeApiCall {
+            val response = api.loadReviews(url, page, showSpoilers)
```
fire-light42, line 205: No caching?

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
```diff
@@ -194,4 +195,19 @@ class APIRepository(val api: MainAPI) {
...
+            val reviewsPerPage = 20
+            val totalReviews = response
+            val startIndex = (page - 1) * reviewsPerPage
+            val endIndex = startIndex + reviewsPerPage
+            totalReviews.subList(startIndex, minOf(endIndex, totalReviews.size))
```
fire-light42, line 210: This makes no sense.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ReviewAdapter.kt
```diff
@@ -0,0 +1,122 @@
...
+                    card.ratings?.forEach { (a, b) ->
+                        val chip = Chip(context)
+                        val chipDrawable = ChipDrawable.createFromAttributes(
+                            context,
+                            null,
+                            0,
+                            R.style.ChipReview
+                        )
+                        chip.setChipDrawable(chipDrawable)
+                        chip.text = "$b ${a / 200}★"
```
fire-light42, line 100: This is /200 but not `card.rating`?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -519,6 +519,40 @@ abstract class MainAPI {
         throw NotImplementedError()
     }
 
+    open val hasReviews: Boolean = false
+    open suspend fun loadReviews(
+        url: String,
```
fire-light42, line 524: This can lead to duplicate requests of the page, and should instead be a variable passed from LoadResponse.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -519,6 +519,40 @@ abstract class MainAPI {
...
+        page: Int,
+        showSpoilers: Boolean = false
+    ): List<UserReview> {
```
fire-light42, line 527: Unlike HomePageResponse, this lacks a hasNext making it hard to know if we should re-fetch or not if it fails.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -486,6 +488,70 @@ class ResultViewModel2 : ViewModel() {
...
+    private fun loadMoreReviews(url: String) {
+        viewModelScope.launch {
+            if (loadMoreReviewsMutex.isLocked) return@launch
+            loadMoreReviewsMutex.withLock {
+                val loadPage = (reviewPage.value ?: 0) + 1
+                if (loadPage == 1) {
+                    reviews.postValue(Resource.Loading())
+                }
+                val response = currentResponse ?: return@launch
+                val api =
```
fire-light42, line 518: Use `currentRepo`.

fire-light42 (changes requested):
I only a reviewed the code itself for API changes, and did not test it. You will get a tested review after the draft period. Keep in mind that the mainapi has very strict requirements as it is hard to change after introducing it.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: This is to replace .copy() so we can future proof it and allow to make constructor (and thus copy()) internal so only the newUserReview builder can be used, however, due to JVM it needed to use it.copy() somewhere else so this was done as an alternative as just initializing it another way didn't work. But I will look into a better solution. The code here is very much still a work in progress.

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
Luna712: This is to limit to loading certain amounts of responses at once to UI for performance so as you scroll it can loud more responses rather than all at once. But this is also still very much WiP.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ReviewAdapter.kt
Luna712: Accident and still a lot of work planned to let API have more control over the format of rating it provides.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: A bit confused what you mean here? url is a variable passed from loadresponse from ResultViewModel2.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Wdym? rename avatarUrl to avatarHeaders or add a new avatar Headers? And should it except a list or something? Kinda confused here...

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
fire-light42: `UI for performance` No, a proper recycleview wont render out of view items and as such this should have zero impact on performance.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: Well this assume that a page has a link to the reviews with the same format as the mainurl, like

https://www.imdb.com/title/tt1016150/ has a link on its page to https://www.imdb.com/title/tt1016150/reviews/ so you can just append /reviews/.

However instead imagine that `https://www.imdb.com/title/tt1016150/` generated a token via js or similar to an api request, then each time we would want to call `loadReviews` we would have to do `get(https://www.imdb.com/title/tt1016150/)`, generate the token, then the real `get(reviewsUrl)` request, making it load twice as long because of a bad API. `MainAPI::load` should instead be able to do the work that is shared between all requests, similar to how `MainAPI::loadLinks` does not take the main url+episode, but instead an url/json given by `load`.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: Just call `this.initializer(); return this;` ?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: There is no headers for the image to get loaded, like referer, cookies or authentication. Add the field `avatarHeaders` or similar.

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
Luna712: I am looking into why as I don't understand why it does, but it does has a lot of performance issues without this... UI lags alot.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Done

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
Luna712: Removed

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: I think this should now support this using ratingFormat

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: This is now documented, and in addition, now supports many other score options using RatingFormat

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: I gave spoiler support control to the API so it is no longer done here at all, can pass isSpoiler to UserReview and then it has a UI for spoilers (which will be improved) as for sort I guess we could add it if you think we need to?

app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
Luna712: What do you mean? How would we cache it here?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: If I understood correctly, this is now done.

Luna712:
I put this on hold for now. I am go back to it eventually but this has become much more work than I originally bargained for...

PR #1541, opened 2025-02-05
Move metaproviders to library

Luna712:
So that providers can extend them when using the library dependency.

#1544 should be merged first since the tmdb package was implicitly loading gson which is used in PluginManager and this moves the tmdb package to commonMain in library so it is no longer readable in PluginManager.


app/build.gradle.kts
```diff
@@ -205,7 +205,7 @@ dependencies {
     implementation(libs.quickjs)
     implementation(libs.fuzzywuzzy) // Library/Ext Searching with Levenshtein Distance
     implementation(libs.safefile) // To Prevent the URI File Fu*kery
-    implementation(libs.tmdb.java) // TMDB API v3 Wrapper Made with RetroFit
+    implementation(libs.gson) // Only Used in PluginManager
```
fire-light42, line 208: I do not understand why we need another JSON parser, can you please clarify why jackson is not used instead?

fire-light42 (changes requested):

app/build.gradle.kts
Luna712: I have no idea, for some reason PluginManager uses gson so I just kept it, before it seemed to pull gson from the tmdb library since that is moved I changed this to be explicit so it just keeps the exact same behavior as before.

Luna712: @fire-light42 #1544 replaces the use of gson in PluginManager, so once that is merged I will update this PR to remove that part. I did it separately since it was totally unrelated to this PR.

Luna712: Removed gson here but build will fail until #1544 is merged so that should be done first.

Luna712:
@fire-light42 this should be okay to merge now I think?

fire-light42 (approved):

PR #1542, opened 2025-02-05
Move SubtitleHelper to library

Luna712:
So that providers can use it when using the library dependency.


fire-light42:
Good idea.

PR #1545, opened 2025-02-07
Include Dub/Sub selection for MovieLoadResponse

JustRelaxable:
This is not meant to be merged. Please check the related issue.


fire-light42 (changes requested):
While this is a good idea, I do not think this is the correct implementation. I understand that the current system might not have true multilingual support. However that could be taken care of on an extension level, by creating 2 extensions or adding settings to the extensions. The only reason why anime has dub/sub/raw is because that medium of media often has this reflected on the sites themself, while movie sites tend to be monolingual.

PR #1549, opened 2025-02-08
Fix typos in unimplimented methods in MainAPI

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1465,18 +1465,18 @@ interface LoadResponse {
...
         fun LoadResponse.addImdbId(id: String?) {
-            // TODO add imdb sync
+            // TODO add IMDb sync
             this.addSimklId(SimklSyncServices.Imdb, id)
         }
 
         @Suppress("UNUSED_PARAMETER")
-        fun LoadResponse.addTrackId(id: String?) {
-            // TODO add trackt sync
+        fun LoadResponse.addTraktId(id: String?) {
```
fire-light42, line 1473: While this might seam like an easy fix, it is not. Modifying the api means that until a new stable, either stable or pre is going to crash, even if the implementation is empty. To fix this, please add back the old functions too, but mark them as `@Deprecated("Outdated API", ReplaceWith("addTraktId(id)"))`, and make them call the new functions.

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: Dont forget `@Prerelease` tag. Or instead add `@JvmName("oldjvmnamehere")`.

Luna712: Oh right thanks. Good point.

fire-light42 (approved):

PR #1555, opened 2025-02-11
refactor: replace PendingIntent#getActivity with PendingIntentCompat#getActivity

Bnyro:
`PendingIntentCompat#getActivity` does exactly the same as the previous version code check does, so we can get rid of that extra sdk version checks here.


fire-light42 (approved):
Nice work!

PR #1568, opened 2025-02-23
bugfix(subdl): allow special-chars in password

int3debug:


app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/Subdl.kt
```diff
@@ -140,7 +144,9 @@ class SubDlApi(index: Int) : InAppAuthAPIManager(index), AbstractSubApi {
             url = "$APIURL/login",
             data = mapOf(
                 "email" to useremail,
-                "password" to password
+                "password" to withContext(Dispatchers.IO) {
+                    URLEncoder.encode(password, "UTF-8")
```
fire-light42, line 148: Just use `json = mapOf(...)` so we dont have to use Dispatchers or URLEncoder. 
It would also work in case the email was weird too.

fire-light42 (changes requested):

fire-light42 (approved):

PR #1575, opened 2025-03-01
hotfix(postEpisodes) add default fallback to NUMBER_ASC

int3debug:
r5ybn5ro760yi3ux
km9uttju74cf18nw

Small bug in: https://github.com/recloudstream/cloudstream/pull/1565


fire-light42:
This will fix it, however does not respect `DataStoreHelper.resultsSortingMode` and that is a bug also. However thank you for pointing it out, I never tested seasons switching to seasons with exactly 1 episode.

fire-light42:
https://github.com/recloudstream/cloudstream/commit/0a38434dec2a5f6fc2bf91f76f6351163fafc40d

PR #1578, opened 2025-03-01
refactor: Use kotlin based new fuzzy lib (levenshtein distance)

IndusAryan:
image

This is new and updated lib written in kotlin based on xDropFuzzy and has fixes over it. This is slightly smaller too 

https://github.com/jens-muenker/fuzzywuzzy-kotlin?tab=readme-ov-file#changelog

This is draft as i am unable to test it, as the library module is giving error, I have imported this in syncapi and plugin viewmodel but it is not importing in ExtractorAPI 

@KingLucius 
@Blatzar

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -203,7 +203,7 @@ dependencies {
     // Extensions & Other Libs
     implementation(libs.rhino) // Run JavaScript
     implementation(libs.quickjs)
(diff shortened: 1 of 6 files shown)


IndusAryan:
ya, it's not that popular and the current one works

fire-light42:
Closing as Blatzar makes a good argument.

PR #1585, opened 2025-03-05
Feat: Added player notification and play in background via PIP.

fire-light42:
This adds the PlayAsAudio event to the player, to play the video entirely in the background. Right now this is only available as an option after going into PIP, however this can later be added as an alternative to PIP as well. To be able to control it, the player always emits a notification that is able to control the playback.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
@@ -27,6 +27,7 @@ import androidx.annotation.MainThread
 import androidx.annotation.OptIn
 import androidx.core.animation.addListener
 import androidx.core.app.NotificationCompat
+import androidx.core.app.PendingIntentCompat
 import androidx.core.content.ContextCompat
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
@@ -39,7 +40,6 @@ import androidx.media3.common.util.UnstableApi
 import androidx.media3.common.util.Util
 import androidx.media3.exoplayer.ExoPlayer
 import androidx.media3.ui.PlayerNotificationManager
-import androidx.media3.ui.PlayerNotificationManager.ACTION_STOP
 import androidx.media3.ui.PlayerNotificationManager.EXTRA_INSTANCE_ID
 import androidx.media3.ui.PlayerNotificationManager.MediaDescriptionAdapter
 import androidx.preference.PreferenceManager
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -230,6 +249,200 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+    ): PendingIntent {
+        val intent: Intent = Intent(action).setPackage(context.packageName)
+        intent.putExtra(EXTRA_INSTANCE_ID, instanceId)
+        val pendingFlags = if (Util.SDK_INT >= 23) {
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+        } else {
+            PendingIntent.FLAG_UPDATE_CURRENT
+        }
+
+        return PendingIntent.getBroadcast(context, instanceId, intent, pendingFlags)
```
Bnyro, line 266: You could use `PendingIntentCompat#getBroadcast` here, similar to https://github.com/recloudstream/cloudstream/pull/1555.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -230,6 +249,200 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+                        is ExtractorUri -> {
+                            meta.headerName ?: meta.name
+                        }
+
+                        else -> null
+                    } ?: "Unknown"
+                }
+
+                override fun createCurrentContentIntent(player: Player): PendingIntent? {
+                    // When we click the notification, should we open the app?
```
Bnyro, line 301: I think that makes sense, most other apps do so as well.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -230,6 +249,200 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+                    return null
+                }
+
+                override fun getCurrentContentText(player: Player): CharSequence? {
```
Bnyro, line 305: In theory the notification should display what's set inside the `MediaMetadata` of the player's media item, I assume CloudStream doesn't set that? (must be done when creating the media source, e.g. via `MediaItem.Builder()`

That would probably be an improvement for the future, as that means that the proper current metadata would be linked to the media session, i.e. that also allows Android to show the currently playing music metadata (title, author) on the lock screen

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
fire-light42: Makes sense and can be changed, however this was code pulled straight out of androids media3 PlayerNotificationManager, and I am therefore more interested in keeping up the parity.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
fire-light42: Ye, right now cs3 does not set any metadata on the media session as it simply has not been needed. However you are right, but I do think that would fit another PR. #1591

PR #1589, opened 2025-03-06
feature(result) support soft/hard subtitle and custom name

int3debug:
pebor41uxk9pd5kj
4aztora1ehn6giig

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
@@ -155,15 +155,19 @@ enum class EpisodeSortType {
 }
 
 fun txt(status: DubStatus?): UiText? {
-    return txt(
-        when (status) {
-            DubStatus.Dubbed -> R.string.app_dubbed_text
-            DubStatus.Subbed -> R.string.app_subbed_text
-            DubStatus.SoftSub -> R.string.app_softsub_text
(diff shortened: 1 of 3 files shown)


AvocadoWasSeized:
Looks great, really useful. One thing I'd recommend, not really needed, is instead of calling it HardSub you can keep it as the original Sub, and for soft sub you can do Soft-Sub or S-Sub for short. You don't have to do this but I do think that having one named hard sub would be redundant of you then specify soft sub

AvocadoWasSeized:
I see you added the ahrd subbed not changed the original subbed, my bad.

I think you should remove the Hard subbed then to remove redundancy and clean the code so for example the new view model would look like this

        DubStatus.Dubbed -> txt(R.string.app_dubbed_text)
        DubStatus.Subbed -> txt(R.string.app_subbed_text)
        DubStatus.SoftSub -> txt(R.string.app_softsub_text)
        DubStatus.Custom -> {
            val customText = status.tagName ?: "NoName"
            txt(R.string.app_custom_text, customText)

fire-light42:
Im sorry, but while your code may look correct it is fundamentally flawed because of how enums work in Kotlin. To better understand it, imagine that all enums are a `companion object` and when you change a variable of one enum you change it on all instances of the same type. This means that you can not have two different "Custom" enum values, as creating the second will override the first. Because of this you cant make a custom DubStatus with an enum, no matter what you tried. 

You can try it out yourself at: https://pl.kotl.in/nBXWev8EA

PR #1604, opened 2025-03-10
Fix: Added ID_SEEK_HEAD parsing for MKV container.

### LINKED ISSUE #1584: MKV Seek / Better non seeking support

The current upstream MatroskaExtractor from media3 is really bad at finding the seek points. The issue here explains why https://github.com/androidx/media/issues/1143. I have verified this myself on mkv files with mkvinfo, and this is a bug in the upstream media3 repo. 

I suggest resolving this with a local extractor (by ExtractorsFactory) or pushing upstream. It is frankly ridiculous that google cant parse a fucking file correctly. They even say this is the "intended" in issues like https://github.com/google/ExoPlayer/issues/9901, but clearly they are coping hard when every other major video player can parse this format. 

Due to this `extractorOutput.seekMap(new SeekMap.Unseekable(durationUs));` is called and the media is not seekable. Solutions like setConstantBitrateSeekingEnabled does nothing to fix this, as they only apply to other formats. 

In the meantime I also suggest adding better support for non seekable streams by disallowing any form of seeking via actions. 

### Other details

_No response_

### Acknowledgements

fire-light42:
Note due to media3 fuckery we have to copy paste the extractorfactory and mkv parser to edit anything. To make it less of a pain they are also auto translated into Kotlin.

This just added `seekForSeekContent`, `seekHeadContentPosition` and `seekPositionAfterSeekingForHead`. Notably this only supports a single recursive query, so no nested KaxSeekHead.

To detail what happens and why this is needed, the current impl only seeks to cues (timestamps) if the cue header is on the top of the file. What many real mkv editors do is to append the metadata (including cues) to the end of the file, presumably to not have to move like 1gb of media like 20kb forwards. This PR allows the header to include a SeekHead.

This means that when we parse an MKV file what really happens is:

1. find KaxSeekHead
2. seek to KaxSeekHead position
3. find KaxCues
4. seek to KaxCues position
5. parse cues
6. seek back from KaxCues
7. seek back from KaxSeekHead
8. parse rest normally

What happened before was:
1. find KaxCues
2. seek to KaxCues position
3. parse cues
4. seek back from KaxCues
5. parse rest normally

This resolves #1584

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -844,10 +844,13 @@ class CS3IPlayer : IPlayer {
                             ).build()
                     )
 
+            // Because "Java rules" the media3 team hates to do open classes so we have to copy paste the entire thing to add a custom extractor
+            // This includes the updated MKV extractor that enabled seeking in formats where the seek information is at the back of the file
+            val extractorFactor = UpdatedDefaultExtractorsFactory()
 
             val factory =
-                if (cacheFactory == null) DefaultMediaSourceFactory(context)
-                else DefaultMediaSourceFactory(cacheFactory)
+                if (cacheFactory == null) DefaultMediaSourceFactory(context, extractorFactor)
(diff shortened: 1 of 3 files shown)


KingLucius:
Please try to push it upstream after testing
For me, I have tried and it works

PR #1622, opened 2025-03-24
feat(Extensions): MyDramaList Meta provider

KingLucius:
Extension developers can build their extension over this meta provider by inheriting MyDramaListDemo() instead of MainApi()

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
@@ -1,6 +1,7 @@
 package com.lagradost.cloudstream3.metaproviders
 
 import com.fasterxml.jackson.annotation.JsonProperty
+import com.lagradost.api.BuildConfig
 import com.lagradost.cloudstream3.APIHolder.unixTimeMS
 import com.lagradost.cloudstream3.Actor
 import com.lagradost.cloudstream3.ActorData
@@ -44,7 +45,7 @@ open class MyDramaListDemo : MainAPI() {
 
     companion object {
         const val TAG = "MyDramaList"
-        const val API_KEY = "bac0925df628dce7841a1d4e8d474c2e63fb818b"
+        val API_KEY: String = BuildConfig.MDL_API_KEY
         const val API_HOST = "https://api.mydramalist.com/v1"
         const val SITE_HOST = "https://mydramalist.com"
         private val headerInterceptor = MyDramaListInterceptor()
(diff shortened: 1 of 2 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
```diff
@@ -0,0 +1,466 @@
...
+    override val providerType = ProviderType.MetaProvider
+    override val supportedTypes = setOf(
+        TvType.Movie,
+        TvType.TvSeries,
+        TvType.AsianDrama,
+    )
+
+    companion object {
+        const val TAG = "MyDramaList"
+        const val API_KEY = "bac0925df628dce7841a1d4e8d474c2e63fb818b"
```
fire-light42, line 47: Are you fine with a public API key? It can be included as a BuildConfig.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
```diff
@@ -0,0 +1,466 @@
...
+                date = released,
+            )
+
+            newMovieLoadResponse(
+                name = title,
+                url = data.toJson(),
+                type = TvType.Movie,
+                dataUrl = link.toJson(),
+            ) {
+                this.name = title
```
fire-light42, line 154: This is a lot of code duplication, I would suggest adding a function like function that applies on any load response directly.
```kotlin

open suspend fun Media.applyMedia(response:LoadResponse) // applies all the metadata from Media on the response. 
```
And call it on both branches.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
```diff
@@ -0,0 +1,466 @@
...
+    data class TrailerNode(
+        @JsonProperty("trailer") val trailerDetails: TrailerDetails,
+    )
+
+    data class TrailerDetails(
+        @JsonProperty("id") val id: Long,
+        @JsonProperty("source") val source: String,
+    )
+
+    data class LinkData(
```
fire-light42, line 454: No JsonProperty? You are calling .toJson on this.

fire-light42 (commented):
Just minor code quality comments, I have not tested it yet, but everything looks fine ❤️

library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
KingLucius: Please check if I have right implementation

fire-light42:
It looks correct, but lets hope the lib build is successful

PR #1628, opened 2025-03-29
Group subtitles with the same name

CranberrySoup:
This pull request merges all subtitles with the same name to one group within the player. 
Specific subtitles within the group can be selected using an additional ListView.


fire-light42 (approved):
🔥🔥🔥 Will be merged after testing

phisher98:
image


Okay its Working but I have seen a Bug maybe in English Lang Selection one of option doesn't contain any Text 
image

CranberrySoup:
Embedded subtitles and online subtitles do not have suffixes. I assumed they would not have the same names as normal subtitles and therefore not be shown.

Should I make string/embedded and string/online?

CranberrySoup:
I fixed empty names by checking the subtitle origin as opposed to inserting strings everywhere as this solution provides some extra flexibility when writing subtitle sources.

There can still be multiple embedded subtitles using the same name but properly fixing that requires reworking the entire numbering system.

fire-light42:
🔥

PR #1631, opened 2025-03-30
Chore: Optimize .gitignore

Pk-web6936:
@fire-light42


fire-light42:
I do not see a good reason to merge this considering that most of this is not applicable, like `TypeScript v1 declaration files` when CloudStream does not use TypeScript at all.

PR #1632, opened 2025-03-30
feat(Extensions): newExtractorLink for developers

KingLucius:
- Deprecating `ExtractorLink `in favor of `newExtractorLink`.
- Deprecating `DrmExtractorLink `in favor of `newDrmExtractorLink`.
- Porting all internal extractors to `newExtractorLink`.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
@@ -521,13 +521,11 @@ open class DrmExtractorLink private constructor(
         url: String,
         referer: String? = null,
         quality: Int? = null,
-        isM3u8: Boolean? = false,
         /** the type of the media, use INFER_TYPE if you want to auto infer the type from the url */
         type: ExtractorLinkType? = INFER_TYPE,
         headers: Map<String, String> = mapOf(),
         /** Used for getExtractorVerifierJob() */
         extractorData: String? = null,
-        isDash: Boolean? = false,
         kid: String? = null,
         key: String? = null,
         uuid: UUID = CLEARKEY_UUID,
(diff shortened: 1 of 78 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -450,83 +476,24 @@ val PLAYREADY_UUID = UUID(-0x65fb0f8667bfbd7aL, -0x546d19a41f77a06bL)
  * @property kty Key type "oct" (octet sequence) by default
  * @property keyRequestParameters Parameters that will used to request the key.
  * */
-open class DrmExtractorLink private constructor(
-    override val source: String,
-    override val name: String,
+@Deprecated("Use newExtractorLink", level = DeprecationLevel.WARNING)
```
Luna712, line 479: Just deprecate the constructor not the whole class so types don't become deprecated

Luna712 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
Luna712: See how I did it in MainAPI for this.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -518,15 +604,15 @@ open class ExtractorLink constructor(
...
+    override var headers: Map<String, String> = mapOf(),
     /** Used for getExtractorVerifierJob() */
-    open val extractorData: String? = null,
-    open val type: ExtractorLinkType,
+    open var extractorData: String? = null,
+    open var type: ExtractorLinkType,
 ) : IDownloadableMinimum {
-    val isM3u8: Boolean get() = type == ExtractorLinkType.M3U8
-    val isDash: Boolean get() = type == ExtractorLinkType.DASH
+    var isM3u8: Boolean = false
```
fire-light42, line 614: What? No? This is so wrong. isM3u8 get() was a backport from when we used a bool, however we switched to type because otherwise it would be nonsense when you had several flags like isDash+isM3u8. Reimplement `get() = type == ExtractorLinkType.M3U8` as well as isDash. All occurrences of `isM3u8 = true` should be  `type=ExtractorLinkType.M3U8` because we do not have a reasonable way to implement the setter.

fire-light42 (changes requested):

LagradOst (approved):
It is stable time 🐴

PR #1637, opened 2025-04-01
Fix gradient overlay on loading

CranberrySoup:
Currently when loading the result page there is a gradient overlay from the poster gradient. 
By reordering the views the issue disappears.

Example of the issue
file
Notice the difference between episode 1 and episode 2


fire-light42:
🔥

PR #1642, opened 2025-04-04
Fix MyDramaList Api key

KingLucius:
Currently the key passed inside the request with "", this PR should fix the request


LagradOst (approved):

LagradOst:
"1000 request for changes"

PR #1650, opened 2025-04-08
feat: add bigwarp extractors

Bnyro:
can be tested with https://github.com/Bnyro/GermanProviders/tree/master/Xcine


fire-light42:
🔥

PR #1654, opened 2025-04-09
feat: add secvideo (fsst, dsst, csst) extractor

Bnyro:
Example Link: https://fsst.online/embed/882885/

Used by Kinoger.com (and probably others)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/SecvideoOnline.kt
```diff
@@ -0,0 +1,84 @@
...
+                        url = getRealUrl(languageAndSourceMatch.groupValues[2])
+                    )
+                )
+            }
+        }
+    }
+
+    private suspend fun getRealUrl(url: String): String {
+        // follow redirects
+        return app.get(url).url
```
fire-light42, line 67: Please do a head request instead, to really clarify that you do not want to consume the body. But are you sure you need to do this at all? The player should handle redirects.

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/SecvideoOnline.kt
Bnyro: Good catch. I wasn't sure if the player handles it, I will look if it works without that code when I have time 👍

fire-light42 (approved):
🔥

PR #1655, opened 2025-04-09
refact: unify torrent error messages

int3debug:
eea9bo21yktl6xkd

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -1576,7 +1576,8 @@ class CS3IPlayer : IPlayer {
 
                                 DialogInterface.BUTTON_NEGATIVE -> {
                                     Torrent.hasAcceptedTorrentForThisSession = false
-                                    event(ErrorEvent(ErrorLoadingException("Not accepted torrent")))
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1510,6 +1510,7 @@ class CS3IPlayer : IPlayer {
     @MainThread
     private fun loadOnlinePlayer(context: Context, link: ExtractorLink, retry: Boolean = false) {
         Log.i(TAG, "loadOnlinePlayer $link")
+        val torrentNotAccepted = context.getString(R.string.torrent_preferred_media)
```
fire-light42, line 1513: torrent_not_accepted and torrent_preferred_media is different errors

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
fire-light42: Show torrent_preferred_media if it is not enabled in settings, otherwise we show torrent_not_accepted if the user has manually not accepted the torrent dialog.

fire-light42 (approved):
❤️

PR #1658, opened 2025-04-11
feat(TV) : Remove all continue watching items

KingLucius:
Adds "Remove all" for continue watching item's long press on TV UI


fire-light42:
Not sure this is the correct way to do it, from a pure UX/UI perspective.

KingLucius:
> Not sure this is the correct way to do it, from a pure UX/UI perspective.

Any suggestions? I thought about the UX part for long time, thought about adding it to the recycler itself but felt like stupid idea, on TV UI everything is limited 😂

fire-light42:
> thought about adding it to the recycler itself but felt like stupid idea

No I do not think so ❤️ 

It is what I did in my implementation of this PR: https://github.com/recloudstream/cloudstream/commit/bdb1e12f3bf20555ee82f7b24e0636c21c8c0ae3

PR #1667, opened 2025-04-20
Fix Voe Extractor

phisher98:


Bnyro:
Out of interest, how did you figure out how to fix this @phisher98 ? Do you know some other projects that implemented it or do you have some tools that help debugging the JavaScript code, and if so which ones?

phisher98:
I had some help with People which have Expertise in JS

Slavko-P:
> Out of interest, how did you figure out how to fix this @phisher98 ? Do you know some other projects that implemented it or do you have some tools that help debugging the JavaScript code, and if so which ones?

I was interested to find that out too, so after almost two hours of wadding through the debugger I found this code:
image
Basically you have to check the Network tab to see which script invokes the decoded URL. Then you go to the Debugger and override that script with a "beautified" one, and start digging around. Afterwards just analyze and implement whatever the script is doing.

Bnyro:
> Basically you have to check the Network tab to see which script invokes the decoded URL. Then you go to the Debugger and override that script with a "beautified" one, and start digging around. Afterwards just analyze and implement whatever the script is doing.
>
That's a good hint, thank you for the explanation!

PR #1668, opened 2025-04-20
fix: fixed coil headers again again 💀 💀

IndusAryan:
This general improvement does the following:

1. Fixes a very sneaky bug in production caused by my friend - Kum- alt @fire-light42 .
2. Uses a debug logger for better verbosity and even shows if the image is being loaded from memory cache, disk cache (LRU) or the network.
3. Exploits Kotlin's semantics to set up the logger.
4. Function Block body removed.
5. Cache is increased to 512MB.

6. Media3 ExoPlayer incremented to 1.6.1.
7. AGP bumped to 8.9.1 so as CoreKTX.


fire-light42:
❤️

PR #1677, opened 2025-04-30
fix(TraktProvider): use trakt provided ids in search response for consistent keys

valhn:
The search response url is used to generate keys in the app and the current implementation of trakt passes a data class that contains keys such as "comment_count","rating","votes" and others which change too frequently. This pr uses just the ids provided by trakt and the media type as the url for the search response.


fire-light42:
Are you [lephx](https://github.com/lephx)? Because #1646 is an exact duplicate. Are you trying to bump the PR?

fire-light42:
You found a core issue with the current API approach that made it impossible to do JSON correctly, and that is why I choose to solve it in the API instead instead. So while this was a very good pull request, I instead added a uniqueUrl that can be used as the unique consistent key https://github.com/recloudstream/cloudstream/commit/df158135409ef3371d1a0b73fb874a545ea0e88a . ❤️

valhn:
Yh, the API handling it is a much better solution.

fire-light42:
> Yh, the API handling it is a much better solution.

Because it will slow down the app if every call to load will take an extra get request. By adding another field, extension developers still have the flexibility to put anything in the url field without worrying about the persistent key.

PR #1684, opened 2025-05-08
Build Config Hash and Build time

rahulDNS:
This update adds two new BuildConfig fields — COMMIT_HASH and BUILD_TIME — to embed the current Git commit hash and the build timestamp directly into the application at compile time. This enhancement improves traceability, allowing developers and QA teams to easily identify the source and timing of a specific build. It’s especially useful for debugging, crash reporting, and release tracking across CI/CD environments.


rahulDNS:
This update adds two new BuildConfig fields — COMMIT_HASH and BUILD_TIME — to embed the current Git commit hash and the build timestamp directly into the application at compile time.

LagradOst:
Is this an LLM generated PR? Because both these are already visible in app and present like 5 lines above the code you inserted.

Issue #999, opened 2024-03-20
The " Resize Button" and Black Bar Problem 


JimHopper0:
### Describe your suggested feature

the resize button has only three option
1) Zoom
2) Fit to Screen
3 ) sketch

sometimes the black bar is still visible even after trying those three option of resizing , in order to fix this problem add more resizing option or **add a feature like nPlayer where we can zoom the video using two finger** 

I hope other users also find this feature helpful 
_have a nice day cloudstream team_

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Issue #987, opened 2024-03-14
Some features to enhance User Experience 


rusunny123:
### Describe your suggested feature

1) Two fingers zoom like MXplayer
2) Play with extenal player ( currently app allows only VLC and mpv player only )
3) Make the app access storage video and act as a video player 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Pentaphon:
I don't see the need for this. Just use another player like https://github.com/anilbeesetti/nextplayer

adezukZ:
But you don't add next player in preferred video player

PR #1685, opened 2025-05-10
Add feature pinch to zoom

### LINKED ISSUE #1619: pinch-to-zoom

When i want to zoom in in pan very large zone in the frame I want to use pinch-to-zoom function to make detailed in the movie easy to see

### Other details

_No response_

### Acknowledgements

jolojo500:
This pull reuquest should close #1619 . It adds A class that wraps the activity_main.xml layout, successfully enabling gesture based zooming. It also has smooth transitioning back to the view's original size to not mess with the core and feel of the ui. Was tested on both an emulator and a real device.

Below is a demonstration of it working.
https://github.com/user-attachments/assets/5a03d286-67c8-4b6b-abac-13096a827218

### DIFF
--- a/.idea/AndroidProjectSystem.xml
+++ b/.idea/AndroidProjectSystem.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AndroidProjectSystem">
+    <option name="providerId" value="com.android.tools.idea.GradleProjectSystem" />
+  </component>
+</project>
--- a/.idea/artifacts/library_jvm_1_0_0.xml
+++ b/.idea/artifacts/library_jvm_1_0_0.xml
@@ -1,0 +1,8 @@
+<component name="ArtifactManager">
+  <artifact type="jar" name="library-jvm-1.0.0">
+    <output-path>$PROJECT_DIR$/library/build/libs</output-path>
+    <root id="archive" name="library-jvm-1.0.0.jar">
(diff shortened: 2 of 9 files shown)


fire-light42:
While I really appreciate this pull request, I meant it more like a pinch-to-zoom like it is on YouTube and not app wide zoom. An app wide zoom can be provided by the operating systems accessibility settings, so I do not believe this implementation is fit for CloudStream. I now see the discussion in the issue, and realize that I misinterpreted the original intent, but see #987 #999

fire-light42:
I believe that you can still continue with this pull request if you move the zoom into the videoplayer xmls, and instead of resetting when releasing only reset when you zoom out. That way users can set a zoom for a movie without having to hold it. Also, it might be interesting to only allow zoom changes when the videoplayer is not locked, https://github.com/recloudstream/cloudstream/blob/4217a716cd0a973dd26fa7c633451114da8308bb/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt#L107

StupidRepo:
> While I really appreciate this pull request, I meant it more like a pinch-to-zoom like it is on YouTube and not app wide zoom. An app wide zoom can be provided by the operating systems accessibility settings, so I do not believe this implementation is fit for CloudStream. I now see the discussion in the issue, and realize that I misinterpreted the original intent, but see #987 #999

Yeah. I mentioned this to them in a reply on the original issue, but I guess they didn't see it: https://github.com/recloudstream/cloudstream/issues/1619#issuecomment-2867989910

PR #1705, opened 2025-06-10
Odnoklassniki (Minor Fix)

Kraptor123:
"Changing from ExtractorLinkType.M3U8 to INFER_TYPE resolves the issue with Odnoklassniki videos not playing."


phisher98:
@Kraptor123 Confirmed its Working 

Good Work

Issue #1721, opened 2025-06-28
MediaCodecAudioRenderer error


NivinCNC:
### Note

Works with external player [VLC for android] by changing player. Its an internal player issue.

### Steps to reproduce

- Install the MovieBoxProvider Repo From [here](https://cutt.ly/qrQa38ja)
- Play a movie ACE(Tamil) and then from _source_ select TamilDub - 1080p
- You will get the error as  **_MediaCodecAudioRenderer error, index=1, format=Format(2, null, video/mp4, audio/mp4a-latm, mp4a.40.2, 394587, ta, [-1, -1, -1.0, null], [0, 48000]), format_supported=YES_**
- 

https://github.com/user-attachments/assets/15db7a7c-b84c-4dfc-a1ab-5db73ea1cf89

### Expected behavior

Should play 

### Actual behavior

Got error and fallbacked to another source

### Cloudstream version and commit hash

4.5.2 d9594158

### Android version

Android 11

### Logcat

```java
06-28 14:47:34.322  4834  5006 I PreviewImgMp4: Loading with url = https://macdn.hakunaymatata.com/resource/3f7f7f456d0aa48bda15176a6e28dd72.mp4?Expires=1751792859&Signature=7tzIr505QTxS6Xr-UnJqSUeTZ4t6phVICjYi0TTnoK0S2zfAG4qvcwOC77MC0meTqpysujUyE9uENm9dbChbtTp5DI0VbCNX6GqNnFDdjMpj8~JUrCJDv1UCvBkBnnWB7xfQ3fP8m8QonpeDa5A9M~f2HLchhizNIFzif2MYV45XjCpg-UtPHdv-lFq0Da6D3OZsBXM88a-I9fnzZR9qxlOmrH7UHnZlKcLEd1orJasPMBYRQ1fw7t3WyFLicc~tMxbfF19iQL1tpNWPtrRcNJ9MBMDnwBgOmqTPe3nHwJeo-sQpTABHMa1MvszJ7CMd03wcwZ7yU7a2y6rB0pwAeA__&Key-Pair-Id=KMHN1LQ1HEUPL headers = {Referer=https://api.inmoviebox.com, Cookie=}
06-28 14:47:34.322  4834  5006 V MediaHTTPService: MediaHTTPService(android.media.MediaHTTPService@3607c46): Cookies: null
[...]

PR #1723, opened 2025-06-29
Add 'Always ask' option for video player selection

NivinCNC:
Introduces an 'Always ask' option in player settings, allowing users to choose a player each time an video is played. Updates the settings UI, string resources, and episode handling logic to support this feature.

Purpose: I got issues in some mp4 files in which a audioDecoder error occurs in internal player . For that sake i used to switch to vlc in settings. But some Dash files that requires cookies didnt work in vlc but worked in internal player. Its been a headache between switching players in settings for each media everytime. So i made a "Always Ask" In prefered video player which allows users to selecte player while they play video on the go..

This could be a temp solution for the issue [#1721 ](https://github.com/recloudstream/cloudstream/issues/1721)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/actions/AlwaysAskAction.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/actions/AlwaysAskAction.kt
@@ -1,0 +1,29 @@
+package com.lagradost.cloudstream3.actions
+
+import android.content.Context
+import com.lagradost.cloudstream3.R
+import com.lagradost.cloudstream3.ui.result.LinkLoadingResult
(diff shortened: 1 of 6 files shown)


int3debug:
You can press and hold on "play movie" and then select your player

fire-light42 (changes requested):
While this will work, the preferred solution in this case is to add it as a "VideoClickAction" because that system is more modular and easy to maintain.

NivinCNC:
Added as "VideoClickAction" . Kindly checkout and leave comments.

NivinCNC:
> You can press and hold on "play movie" and then select your player

In my pov "Always Ask" is the most convenient way

PR #1726, opened 2025-06-29
Fix subtitle overlap

CranberrySoup:
This pull request tries to fix all subtitle overlap by simply combining all subtitle texts into a single text. It could also be fixed by pinning extra subtitles to the top of the screen. That solution runs into multiple UI issues which is why this solution was used instead.

The only potential issue with this would be a small subtitle pinned to a location of the screen while the rest of the text is supposed to be at the bottom. I am not sure if this is a plausible scenario. A lot of real world testing is needed.

The most basic overlapping vtt subtitle looks like the following:
```
WEBVTT

00:00:01.000 --> 00:00:05.000
Lorem ipsum dolor sit amet consectetur adipiscing elit. Amet consectetur adipiscing elit quisque faucibus ex sapien. Quisque faucibus ex sapien vitae pellentesque sem placerat. Vitae pellentesque sem placerat in id cursus mi.

00:00:03.500 --> 00:00:06.000
Dui felis venenatis ultrices proin libero feugiat tristique. Sem placerat in id cursus mi pretium tellus. Taciti sociosqu ad litora torquent per conubia nostra. Mauris pharetra vestibulum fusce dictum risus blandit quis. Justo lectus commodo augue arcu dignissim velit aliquam.

00:00:06.500 --> 00:00:09.000
Senectus netus suscipit auctor curabitur facilisi cubilia curae. Sed diam urna tempor pulvinar vivamus fringilla lacus. Et magnis dis parturient montes nascetur ridiculus mus.

00:00:07.500 --> 00:00:10.000
nte condimentum neque at luctus nibh finibus facilisis.

00:00:11.000 --> 00:00:14.000
Rutrum gravida cras eleifend turpis fames primis vulputate. Dolor sit amet consectetur adipiscing elit quisque faucibus.

00:00:12.500 --> 00:00:15.000
Nisl malesuada lacinia integer nunc posuere ut hendrerit.
```

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CustomSubtitleDecoderFactory.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CustomSubtitleDecoderFactory.kt
@@ -242,10 +242,25 @@ class CustomDecoder(private val fallbackFormat: Format?) : SubtitleParser {
         val currentStyle = style
         val customOutput = Consumer<CuesWithTiming> { cue ->
             // fixed style and filter on the cues
+
+            val combinedCueText = cue.cues.joinToString("\n") {
+                it.text?.toString() ?: ""
+            }
+
+            // Move all cues into one single one
+            // This is to prevent text overlap in vtt (and potentially other) subtitle files
+            val combinedCues = cue.cues
+                .firstOrNull()
+                ?.buildUpon()
+                ?.fixSubtitleAlignment()
+                ?.applyStyle(currentStyle)
+                ?.setText(combinedCueText)
+                ?.build()
+                ?.let { mutableListOf(it) } ?: mutableListOf()
+
+
             val newCue =
-                CuesWithTiming(cue.cues.map { c ->
-                    c.buildUpon().fixSubtitleAlignment().applyStyle(currentStyle).build()
-                }, cue.startTimeUs, cue.durationUs)
+                CuesWithTiming(combinedCues, cue.startTimeUs, cue.durationUs)
 
             // do not apply the offset to the currentSubtitleCues as those are then used for sync subs
             currentSubtitleCues.add(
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/CustomSubtitleDecoderFactory.kt
```diff
@@ -242,10 +242,25 @@ class CustomDecoder(private val fallbackFormat: Format?) : SubtitleParser {
...
         val customOutput = Consumer<CuesWithTiming> { cue ->
             // fixed style and filter on the cues
+
+            val combinedCueText = cue.cues.joinToString("\n") {
+                it.text?.toString() ?: ""
+            }
+
+            // Move all cues into one single one
+            // This is to prevent text overlap in vtt (and potentially other) subtitle files
+            val combinedCues = cue.cues
```
fire-light42: This looks good, but ideally it should also do `.groupBy { it.lineAnchor to it.positionAnchor }` to avoid grouping subtitles that actually align on different locations. However it might be better to only group by lineAnchor as horizontal subtitles locations might not matter. Therefore something like `it.lineAnchor to it.position.times(1000.0f).toInt()` might be used instead.

```kotlin
            val combinedCues = cue.cues.groupBy {
                it.lineAnchor to it.position.times(1000.0f).toInt()
            }.mapNotNull {
                val combinedCueText = cue.cues.joinToString("\n") {
                    it.text?.toString() ?: ""
                }

                // Move all cues into one single one
                // This is to prevent text overlap in vtt (and potentially other) subtitle files
                cue.cues
                    .firstOrNull()
                    ?.buildUpon()
                    ?.fixSubtitleAlignment()
                    ?.applyStyle(currentStyle)
                    ?.setText(combinedCueText)
                    ?.build()
            }
```

PR #1731, opened 2025-07-01
Paginated search

BlipBlob:
Initial draft for paginated search. 
More app integration to come soon

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -422,19 +422,17 @@ fun newHomePageResponse(list: List<HomePageList>, hasNext: Boolean? = null): Hom
 
 @Prerelease
 fun newSearchResponseList(
-    list: List<SearchResponse>?,
+    list: List<SearchResponse>,
     hasNext: Boolean? = null,
 ): SearchResponseList {
-    val newList = list ?: emptyList()
-    @Suppress("DEPRECATION")
     return SearchResponseList(
-        newList,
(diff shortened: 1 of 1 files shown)


BlipBlob:
Tested and works with example extensions, also backwards compatible. Ready for review @fire-light42

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -420,6 +420,24 @@ fun newHomePageResponse(list: List<HomePageList>, hasNext: Boolean? = null): Hom
     return HomePageResponse(list, hasNext = hasNext ?: list.any { it.list.isNotEmpty() })
 }
 
+@Prerelease
+fun newSearchResponseList(
+    list: List<SearchResponse>?,
+    hasNext: Boolean? = null,
+): SearchResponseList {
+    val newList = list ?: emptyList()
```
fire-light42: As you take a List<>? we might actually want to return SearchResponseList?, or else make list non-null. Same with toNewSearchResponseList.

fire-light42 (commented):
Works well when testing!

Kraptor123:
I tried it, it’s awesome, thank you! But I couldn’t figure out the quick search part, and it only works in the multi-search section. Are you planning to implement it in the extensions’ own search section as well?

BlipBlob:
> But I couldn’t figure out the quick search part

Quick search was not intended to be paginated.

>  Are you planning to implement it in the extensions’ own search section as well?

Yes, that was an something I missed. Extensions' own search works differently, which means the fix is not that simple.

PR #1745, opened 2025-07-15
Added a rating preview on the poster

Sarlay:
Here is a preview:
IMG_20250715_195851

PR #1748, opened 2025-07-17
Added new class Score

fire-light42:
This is primary inspired by https://github.com/recloudstream/cloudstream/pull/1745 as CloudStream currently has several different ways of storing score, 0-10000, 0-100 and 0-10. This seeks to unify them into a single primitive that is easier to use. 

The new syntax would be `score = Score.from10("7.5")` or `score = Score.from5(4.1)` or `score = Score.from100(84)` or `addScore("7.5")` instead of `rating = text?.toFloat()?.times(1000)?.toInt()`

As this is primarily to make the API easier, what do some of you extension developers think of this @phisher98 @Luna712 @Sarlay @KingLucius ?

Anyone, comment if you have any input on this.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/syncproviders/SyncApi.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/syncproviders/SyncApi.kt
@@ -80,8 +80,7 @@ interface SyncAPI : OAuth2API {
         var totalEpisodes: Int? = null,
 
         var title: String? = null,
-        /**1-1000*/
-        var publicScore: Int? = null,
+        var publicScore: Score? = null,
         /**In minutes*/
         var duration: Int? = null,
         var synopsis: String? = null,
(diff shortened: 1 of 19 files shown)


KingLucius:
Totally agree 👍

fire-light42:
> Would this mean that for scoring on sync, such as with mal or anilist, will decimals finally work? I tried to implement it myself but I couldn't, and it always bothered me since I use 10 decimal system.
> 
> Or will it be easier to implement now?

I guess, but only by a negligible amount. This pull request is mostly to make it easier for first time extension developers to make sense of the scoring system, and make it easier to write for long time extension developers.

Kawagetsuu:
> > Would this mean that for scoring on sync, such as with mal or anilist, will decimals finally work? I tried to implement it myself but I couldn't, and it always bothered me since I use 10 decimal system.
> > Or will it be easier to implement now?
> 
> I guess, but only by a negligible amount. This pull request is mostly to make it easier for first time extension developers to make sense of the scoring system, and make it easier to write for long time extension developers.

Ahh I see, thanks for letting me know, I'll look into this feature again. I was just curious, thanks for the reply

Luna712:
I absolutely agree with this as well.

phisher98:
It's pretty good

PR #1755, opened 2025-07-19
feat: increase player UI hide delay to 5 seconds

int3debug:


Luna712:
This PR kinda reminds me, not sure if that is the reason for it, but my opinion is while a touch is ongoing on controls on mobile, it should not hide at all, because, especially in vertical mode, it makes sliding to the further controls when many are enabled hard as as you are sliding the controls just disappear on you. This would probably help that but its just my opinion that it would be better to just not hide it while a touch is ongoing and reser the delay during that time so that it hides from certain amount of seconds after you arent touching it at all.

I'm probably explaining this very poorly though.

PR #1756, opened 2025-07-19
feat: enhance player error logging with url, headers, and referer

int3debug:


fire-light42 (changes requested):
Use `override fun playerError(exception: Throwable) {` in GeneratorPlayer along with currentSelectedLink, instead of adding more functions to the interface.

int3debug:
> Use `override fun playerError(exception: Throwable) {` in GeneratorPlayer along with currentSelectedLink, instead of adding more functions to the interface.

one day I received a PR without any corrections

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1503,7 +1503,20 @@ class GeneratorPlayer : FullScreenPlayer() {
 
 
     override fun playerError(exception: Throwable) {
-        Log.i(TAG, "playerError = $currentSelectedLink")
+        val currentUrl =
+            currentSelectedLink?.let { it.first?.url ?: it.second?.uri?.toString() } ?: "unknown"
+        val headers = currentSelectedLink?.first?.headers?.toString() ?: "none"
+        val referer = currentSelectedLink?.first?.referer ?: "none"
+        Log.e(
```
fire-light42, line 1510: This is really tricky, but this code may crash the entire application. I almost missed this bug when reviewing, so do not take it as hard critique. 

A crash can happen when using a torrent, if playerError is a SocketTimeoutException. A SocketTimeoutException means that playerError is thrown on another thread that is not the main UI thread. Normally this would be fine just for logging, however  getPosition, getDuration will throw an IllegalStateException as verifyApplicationThread is checked, as exoplayer can only be interacted with on the main thread. 

You have two options for solving this:
1. Wrap `eventHandler?.invoke` or `mainCallback` in runOnUiThread/runOnMainThread
2. Do not use getDuration and getIsPlaying

I believe that making the event run on the main thread should be the expected behavior so others do not make the same mistake.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
int3debug: Then it was better to add it as the first commit. I need to take a closer look at this. I don't know about this main thread and crash. Because I couldn't reproduce the error
I need some time for this change. Please forgive me

PR #1763, opened 2025-07-21
Add quick scroll-to-top functionality via long press on Home button

onurcvnoglu:
When users scroll down extensively on long pages, returning to the top can be inconvenient. With this update, users can now quickly return to the top of the main page by long-pressing the Home button, improving overall navigation and user experience.

### DIFF
--- a/.idea/AndroidProjectSystem.xml
+++ b/.idea/AndroidProjectSystem.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AndroidProjectSystem">
+    <option name="providerId" value="com.android.tools.idea.GradleProjectSystem" />
+  </component>
+</project>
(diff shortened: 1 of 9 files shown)


onurcvnoglu:
I reviewed and updated the codes. Thank you.

fire-light42:
> I reviewed and updated the codes. Thank you.

That is good, however you still have auto-generated files from your IDE in the .idea and .kotlin folder. Please remove those files as they are not relevant to other people. 👍

onurcvnoglu:
You're right. Thank you for your help and information. I've updated the code again.

fire-light42 (approved):

Aurora3rs:
This is a great feature, however it's only applicable to the Homepage, it would be better if it was the same for Library, especially since Library would normally be a long page and contain a lot of Movies/Series, in the "completed" section for e.g.
Thanks for your work.

PR #1768, opened 2025-07-21
Add DRM request headers support in player

NivinCNC:
Headers from drm.keyRequestParameters and currentLink.headers are now added to DRM license requests for CLEARKEY and PLAYREADY. This improves compatibility with DRM servers requiring custom headers.

This fixes custum header requirement for DRM protected contents.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -1063,13 +1063,9 @@ class CS3IPlayer : IPlayer {
             item.drm?.let { drm ->
                 when (drm.uuid) {
                     CLEARKEY_UUID -> {
-                        val client =
-                            OkHttpDataSource.Factory(app.baseClient).setUserAgent(USER_AGENT)
-                        // Add headers from drm.keyRequestParameters (if any) and from the link.headers
-                        val headers = drm.keyRequestParameters + (currentLink?.headers ?: emptyMap())
-                        if (headers.isNotEmpty()) {
-                            client.setDefaultRequestProperties(headers)
-                        }
+                        // Use createOnlineSource to properly apply headers from the ExtractorLink
+                        val client = currentLink?.let { createOnlineSource(it) }
+                            ?: OkHttpDataSource.Factory(app.baseClient).setUserAgent(USER_AGENT)             
                         val drmCallback =
                             LocalMediaDrmCallback("{\"keys\":[{\"kty\":\"${drm.kty}\",\"k\":\"${drm.key}\",\"kid\":\"${drm.kid}\"}],\"type\":\"temporary\"}".toByteArray())
                         val manager = DefaultDrmSessionManager.Builder()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1065,6 +1065,11 @@ class CS3IPlayer : IPlayer {
                     CLEARKEY_UUID -> {
                         val client =
                             OkHttpDataSource.Factory(app.baseClient).setUserAgent(USER_AGENT)
+                        // Add headers from drm.keyRequestParameters (if any) and from the link.headers
+                        val headers = drm.keyRequestParameters + (currentLink?.headers ?: emptyMap())
```
fire-light42: I am unsure if keyRequestParameters is correct here given that it is already used in `setKeyRequestParameters(drm.keyRequestParameters)`

Can you double check that headers must be the union of both maps?

NivinCNC: > I am unsure if keyRequestParameters is correct here given that it is already used in `setKeyRequestParameters(drm.keyRequestParameters)`
> 
> Can you double check that headers must be the union of both maps?

According to me, they actually serve different purposes, so it's not redundant.

In this case:

```
val headers = drm.keyRequestParameters + (currentLink?.headers ?: emptyMap())
client.setDefaultRequestProperties(headers)
```
The merged headers map is used to configure the HTTP client (i.e., DefaultHttpDataSource) which sends requests — including potentially the DRM license request, depending on how ExoPlayer routes it.

Later, this call:

`.setKeyRequestParameters(drm.keyRequestParameters)`

is passed to DefaultDrmSessionManager.Builder, and is specifically used to configure the DRM license request body or headers — depending on the underlying DRM implementation (e.g., Widevine). These parameters are not automatically injected into the HTTP request by the player or client unless explicitly passed.

So while the keys may overlap, they are used in different contexts:

setDefaultRequestProperties(...) → sets headers for all HTTP requests made by the client.

setKeyRequestParameters(...) → passes parameters to the DRM session manager for use in license acquisition.

Hence, if the license server or content stream requires both sets of headers, then yes — headers must be the union of drm.keyRequestParameters and currentLink?.headers. Otherwise, key headers may be missing from the request sent by the HTTP client.

fire-light42: "According to me, they actually serve different purposes, so it's not redundant." Yes, and because of that I do not see a need to union them for the HTTP headers. In this case you remove the choice of what headers the extractor could give as the "setDefaultRequestProperties". So why include KeyRequestParameters in HTTP requests? If that is really needed just include it into the currentLink.headers in the extractor too. 

@KingLucius You have more experience with DRM content, can you double check this.

NivinCNC: > "According to me, they actually serve different purposes, so it's not redundant." Yes, and because of that I do not see a need to union them for the HTTP headers. In this case you remove the choice of what headers the extractor could give as the "setDefaultRequestProperties". So why include KeyRequestParameters in HTTP requests? If that is really needed just include it into the currentLink.headers in the extractor too.
> 
> @KingLucius You have more experience with DRM content, can you double check this.

I’ve seen some DRM files where keyRequestParameters are expected to be passed as headers, which is why I initially merged them. But you're right — it's cleaner to let the extractor handle all headers through currentLink.headers. If any of the keyRequestParameters are actually needed as headers, they can be added there directly. I've updated the code to reflect that — thanks for pointing it out!

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1063,8 +1063,9 @@ class CS3IPlayer : IPlayer {
             item.drm?.let { drm ->
                 when (drm.uuid) {
                     CLEARKEY_UUID -> {
-                        val client =
-                            OkHttpDataSource.Factory(app.baseClient).setUserAgent(USER_AGENT)
+                        // Use createOnlineSource to properly apply headers from the ExtractorLink
+                        val client = currentLink?.let { createOnlineSource(it) }
```
fire-light42: I would actually add the link (or headers) to the DrmMetadata class to avoid using the player class variables like this.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
NivinCNC: > I would actually add the link (or headers) to the DrmMetadata class to avoid using the player class variables like this.

Updated.

fire-light42 (approved):
❤️

PR #1770, opened 2025-07-22
Show estimated download time in notification

rockhero1234:


app/src/main/java/com/lagradost/cloudstream3/utils/VideoDownloadManager.kt
```diff
@@ -274,7 +274,19 @@ object VideoDownloadManager {
             return null
         }
     }
-
+    //calculate the time
+    private fun getEstimatedTimeLeft(bytesPerSecond: Long, progress: Long, total: Long):String{
```
fire-light42: This looks a bit ad hoc. Can you please look at how we already handle nextAiringDate with next_episode_time_day_format, next_episode_time_hour_format and next_episode_time_min_format. You could add some formatted string for more precise time format.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/utils/VideoDownloadManager.kt
rockhero1234: Ok

rockhero1234:
@fire-light42 is this good?

fire-light42 (approved):
This looks good, will merge after testing!

Issue #1779, opened 2025-07-23
Feat: Make rating previews on homepage poster optional


Aurora3rs:
### Describe your suggested feature

Add a toggle for the Rating feature in:
Settings -> Layout -> Toggle UI elements on poster

![Image](https://github.com/user-attachments/assets/b6d405db-b25e-4367-b176-8c0d8cf695dc)

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

rockhero1234:
I also think so

PR #1782, opened 2025-07-24
Toggle rating label

rockhero1234:
fix to this #1779

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
@@ -81,7 +81,7 @@ object SearchResultBuilder {
         val showDub = showCache[textIsDub?.context?.getString(R.string.show_dub_key)] ?: false
         val showTitle = showCache[cardText?.context?.getString(R.string.show_title_key)] ?: false
         val showHd = showCache[textQuality?.context?.getString(R.string.show_hd_key)] ?: false
-
+        val showRating = showCache[textQuality?.context?.getString(R.string.show_rating_key)] ?: false
         if(card is SyncAPI.LibraryItem) {
             val showRating = (card.personalRating ?: 0) != 0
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/array.xml
```diff
@@ -202,13 +202,15 @@
         <item>@string/show_hd</item>
         <item>@string/show_dub</item>
         <item>@string/show_sub</item>
+        <item>@string/show_rating</item>
```
fire-light42, line 205: There exists more than one array.xml file that needs to be updated.

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
```diff
@@ -81,7 +81,7 @@ object SearchResultBuilder {
         val showDub = showCache[textIsDub?.context?.getString(R.string.show_dub_key)] ?: false
         val showTitle = showCache[cardText?.context?.getString(R.string.show_title_key)] ?: false
         val showHd = showCache[textQuality?.context?.getString(R.string.show_hd_key)] ?: false
-
+        val showRating = showCache[textQuality?.context?.getString(R.string.show_rating_key)] ?: false
```
fire-light42, line 84: This does not include SyncAPI.LibraryItem?

rockhero1234: Is this needed to be done? That is fine be always display

app/src/main/res/values/array.xml
rockhero1234: done

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
fire-light42: Alright

fire-light42 (approved):

PR #1803, opened 2025-08-03
Initial migration to stateless auth

### LINKED ISSUE #1801: Update the AuthAPI, OAuth2API and AccountManager API

The current code works, but is unmaintainable. These apis should be redesigned to not rely on setkey/getkey, and instead be stateless like MainAPI. Additionally, score(...) should use an enum instead of `0 -> Watching ect`. It should then be possible to open this up to extensions as well. 

However I believe this will also log out everyone, but that should not be a major issue. 

### Other details

_No response_

### Acknowledgements

fire-light42:
This closes #1801 by replacing all auth with stateless classes. Authentication was not maintainable, so this PR replaces them with a new a inheritance structure that tries to follow the same ideas of MainAPI. 

Structure: 
1. Every implementation related of authentication/authorization inherits from AuthAPI, like MainAPI. This should then be used with SyncAPI or SubtitleAPI.
2. AuthAPI has a wrapper called AuthRepo for safe access, that is implemented for SubtitleRepo and SyncRepo. 
3. Every call to any function takes in a AuthToken for authentication instead of storing it in the AuthAPI. The AuthRepo instead stores the current account, meaning that AuthAPI is stateless. 

Important notes: 
1. After looking at some repositories I found that even if the account system is not supposed to be exposed, it is still used @phisher98 👀. Therefore I added *some* backwards compatibility but not much. 
2. While I tested this, I do not trust myself that everything is 100% correct, so it would be appreciated if anyone that has an account on some of the services in CloudStream tested this PR. 
3. All current accounts will be logged out

Any feedback is welcome ❤️ 
@Luna712 @phisher98 @int3debug @KingLucius @CranberrySoup


int3debug:
1. I apologize :)
cpllndxyi0te2py3

2. I will create accounts and test this PR

KingLucius:
Wow, that's seems a very nice addition ❤️ will test & feedback tomorrow

KingLucius:
MAL: signed in and library loaded successfully
Opensubtitles: login 👍
SubDL: can't login on this PR debug, can login on pre

phisher98:
I could change it just let me know what to change in my repo for auth

fire-light42:
> I could change it just let me know what to change in my repo for auth

That would break for everyone on stable, so hold off on that until the next stable release.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsAccount.kt
```diff
@@ -119,247 +119,266 @@ class SettingsAccount : PreferenceFragmentCompat(), BiometricCallback {
...
+            binding.applyBtt.id.let {
+                displayedItems.lastOrNull()?.nextFocusDownId = it
+            }
+
+            binding.text1.text = api.name
+
+            binding.applyBtt.setOnClickListener {
+                val loginData = AuthLoginResponse(
+                    username = if (req.username) binding.loginUsernameInput.text?.toString() else null,
+                    password = if (req.username) binding.loginPasswordInput.text?.toString() else null,
```
fire-light42: @KingLucius silly mistake here, as password is null if no username.

KingLucius: Fixed, thanks

fire-light42:
I updated accounts to also switch when switching the CloudStream account, that is to say what your active account is for every service. You do not need to reauthenticate for the external service when creating a new CloudStream account, however you need to select the account in settings again. This should make it easier when two (or more) people share a device, and use different sync services.

PR #1813, opened 2025-08-07
Feat: Initial setup for trakt, login only

fire-light42:
This is my attempt of adding Trakt sync to CloudStream. However this will be abandoned as the API of trakt is painfully backwards. 

If someone wants to continue this pull request then update `redirectUrlIdentifier`, `id` and `secret` to your local testing app. I have only implemented the login, but I cant get my head around the rest of the actual sync. For debugging purposes TraktProvider is added as a built in provider. 

These are the reasons I dropped this PR:
1. TraktID is not unique?
`Lookup items by their Trakt, IMDB, TMDB, or TVDB ID. If you use the search url without a type it might return multiple items if the id_type is not globally unique. Specify the type of results by sending a single value or a comma delimited string for multiple types.`
2. No watch status like `WATCHING`/`ONHOLD` ect
3. Almost every api requires more metadata then we have available, for example for `https://trakt.docs.apiary.io/#reference/sync/add-ratings/add-new-ratings`, making it very tedious, same with get rating. There is no clear or clean way to implement this as far as I know.


KingLucius:
to be honest, trakt api structure is very stupid and limited that's why it took me a lot of time to get around it and finally my motivation is very down to finalize it
I almost stopped on same step, only implemented pin login & partly library support (shocked by third point).

now you get it bro xD

KingLucius:
I remembered that I used another list names that fits trakt not CS.
I don't know if this can be OK with the sync implementation of CS

Issue #1739, opened 2025-07-08
Support Android 13+ Per-App Language Selection


SittingDuck999:
### Describe your suggested feature

Currently, language selection is handled in-app. On Android 13 (API 33) and above, users can set the app’s language independently in system settings (Settings > Apps > [App Name] > Language).

### Other details

**Benefits**:  
- Aligns with Android best practices.
- Prevents the app defaulting to Chinese on some Chinese ROMs.
- Improves user experience by letting users set language from device settings.
- 

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

PR #1815, opened 2025-08-10
WIP use IETF BCP 47 conformant tag, starts language per-app

### LINKED ISSUE #1791: Filtering repository plug-ins by language doesn't work
### Steps to reproduce

1. Open a repository to list plug-ins
2. Select languages 🇵🇹 Portuguese, 🇧🇷 Brazilian Portuguese
3. Apply and get an empty list

Image

### Expected behavior

List plug-ins matching language

### Actual behavior

Empty list

4.5.4 91652da

### Android version

12

### Logcat

```java

```

### Other details

_No response_

### Acknowledgements

diogob003:
* Set IETF BCP 47 language tag as default
* Fix filter plugins by language
* Show localized language name in plugin details
* Show flag on player subtitles download / auto select language
* Initial changes to support android 13+ language per-app

Fix #1791
Start #1739


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/HlsPlaylistParser.kt
```diff
@@ -1450,7 +1450,16 @@ object HlsPlaylistParser {
          */
         //val labels: List<Label> = emptyList(),
 
-        /** The language as an IETF BCP 47 conformant tag, or null if unknown or not applicable. */
+        /**
```
fire-light42, line 1453: This is media3 code, so it is unrelated to CloudStream.

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+    "recommendations": [
```
fire-light42: Please remove any .vscode/* stuff as that is your local config and not something that should be pushed upstream.

fire-light42 (changes requested):
Can you please split this pull request up into individual changes, because this modify a lot of stuff in a lot of different and unrelated places, making it hard to test and review. 

Right now you modify:
1. localeconfig
2. SubtitleHelper
3. strings.xml
4. libs.versions.toml

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/HlsPlaylistParser.kt
diogob003: You're right, but I didn't make any changes except to include links `IETF BCP 47` language tags

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1083,13 +1095,16 @@ fun TvType.isAnimeOp(): Boolean {
...
-    var lang: String,
+    var lang: String, // most extractors use for language names
     var url: String,
     var headers: Map<String, String>?
 ) {
     /** Backwards compatible constructor, mark this as deprecated when new stable comes out */
     constructor(lang: String, url: String) : this(lang = lang, url = url, headers = null)
 
+    /** Language code to properly filter auto select / download subtitles */
+    var langTag: String? = fromCodeToLangTagIETF(lang) ?: fromLanguageToTagIETF(lang, true)
```
fire-light42: Right now you can set langTag after you have initalized SubtitleFile? This should be a `get() = ` instead, as it is derived from lang.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -42,7 +46,7 @@ object SubtitleHelper {
         val ISO_639_3: String,      // ISO 639-6 missing as it's intended to differentiate specific dialects and variants
         val openSubtitles: String, // inconsistent codes that do not conform ISO 639
     ) {
-        internal fun localizedName(localizedTo: String? = null): String {
+        fun localizedName(localizedTo: String? = null): String {
```
fire-light42, line 50: Prerelease tag?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -60,7 +64,7 @@ object SubtitleHelper {
                 sysLocalizedName
         }
 
-        internal fun nameNextToFlagEmoji(localizedTo: String? = null): String {
+        fun nameNextToFlagEmoji(localizedTo: String? = null): String {
```
fire-light42, line 69: Prerelease tag?

app/src/main/java/com/lagradost/cloudstream3/ui/player/RepoLinkGenerator.kt
```diff
@@ -125,7 +125,7 @@ class RepoLinkGenerator(
 
                 currentSubsNames.add(name)
                 val updatedFile =
-                    correctFile.copy(originalName = fixedName, nameSuffix = "$count")
+                    correctFile.copy(nameSuffix = "$count")
```
fire-light42: Why did you remove this `originalName = fixedName`?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: If it is supposed to be a different variable then make it null as default, otherwise you make weird dependencies.

fire-light42: Additionally, tag it Prerelease.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1670,27 +1669,25 @@ class GeneratorPlayer : FullScreenPlayer() {
...
             }
         }
 
-        sortSubs(subtitles).firstOrNull { sub ->
-            val t = sub.name.replace(Regex("[^A-Za-z]"), " ").trim()
-            settings && t == lang || t.startsWith(lang) || t == langCode
-        }?.let { sub ->
-            return sub
-        }
+        if (settings == false) return null
```
fire-light42: `!settings`

app/src/main/java/com/lagradost/cloudstream3/ui/player/RepoLinkGenerator.kt
diogob003: Because `name` is initialized to `$originalName $nameSuffix` so no need 2 properties `originalName` and `name` to make "all names unique for UX".
I noticed that `originalName` is been used elsewhere to group subtitles, filter and match subtitles. And it's called `originalName`. But I'm not sure why this was being done, if you think this might break something I can undo it.

fire-light42: It is a bit complicated, however that is not relevant here. The fixedName is just the originalName unmangled because some extensions report subtitles with html styling or url encoding, for example `%3C%3E` = `<>`. That is why we do `correctFile.name.html().toString().trim()` but in reality it acts like `correctFile.originalName.html().toString().trim()` as the nameSuffix is not set yet.

diogob003: It makes a lot of sense now that you explained it. Fixed

app/src/main/java/com/lagradost/cloudstream3/ui/player/RepoLinkGenerator.kt
```diff
@@ -113,19 +113,15 @@ class RepoLinkGenerator(
...
-                while (currentSubsNames.contains(name)) {
-                    count++
-                    name =
-                        SubtitleData.constructName(originalName = fixedName, nameSuffix = "$count")
-                }
 
-                currentSubsNames.add(name)
+                val nameDecoded = correctFile.originalName.html().toString().trim() // `%3Ch1%3Esub%20name…` → `<h1>sub name…` → `sub name…`
+
+                val suffixCount = currentSubsNames.count { it.contains(nameDecoded) } +1
```
fire-light42: This is very inefficent, and may have some unintentional effects like "ABCEnglish".contains("English") = true. 

It would make more sense to do a hashmap.
```kotlin
val suffixCount = currentSubsNames.getOrDefault(nameDecoded, 1)
currentSubsNames[nameDecoded] = suffixCount + 1
```

diogob003: wow, good point!
I was doing the same as before, but using `.count` instead of `while`.
Do you think it needs to be `HashMap` or can it be a more generic `Map` and let the compiler choose? I think in this case it doesn't matter.

fire-light42: > wow, good point! I was doing the same as before, but using `.count` instead of `while`. Do you think it needs to be `HashMap` or can it be a more generic `Map` and let the compiler choose? I think in this case it doesn't matter.

The actual map implementation does not matter that much as you will not really gain anything from using the insertion order provided by `LinkedHashMap` (the default), in comparison to `HashMap`. Both works fine and we are not going to use some custom map for this. I just wanted to be explicit about using the correct data structure and not the higher order function by the same name (`.map { ... }`).

The main difference between `count` and `while` is that `while` will only loop for the amount of entries of a single language, while `count` will loop once for every entry that exists even if they had different languages. So while the previous behavior was bad, using count would be worse, and that is why I reacted.

PR #1818, opened 2025-08-10
Fix: unclosed file descriptor, bump libs min version

diogob003:
* Fix: unclose inputPFD:ParcelFileDescriptor in AppContextUtils
* Security: bump libs minor version


.gitignore
```diff
@@ -96,11 +96,6 @@ replay_pid*
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
```
fire-light42, line 99: I dont quite understand why you want to remove these from the .gitignore

fire-light42 (changes requested):
Good pull request, but I want some clarification on the .gitignore part 👍

.gitignore
Luna712: Just assuming here btw
Per your comment in another PR:
"Please remove any .vscode/* stuff as that is your local config and not something that should be pushed upstream."

removing them here makes sure they dont get pushed when there is an ! in the front it overrides the .gitignore from above meaning they aren't ignored.

diogob003: @Luna712 is right
You asked me to remove **any** `.vscode/*` stuff.

`.vscode/*` -> remove everything in .vscode
`!.vscode/settings.json` -> `!` make and exception  in the rules forcing the inclusion of `.vscode/settings.json`

So I just removed the exceptions, and now everything is gone

gradle/libs.versions.toml
```diff
@@ -12,7 +12,7 @@ desugar_jdk_libs_nio = "2.1.5"
 dokkaGradlePlugin = "2.0.0"
 espressoCore = "3.6.1"
 fuzzywuzzy = "1.4.0"
-gradle = "8.9.2"
+gradle = "8.9.3"
```
diogob003, line 15: @Fire-light42 what is the project's dependency update policy? Cause newer version 8.12.0 was reported by the lint tool, but I do not want to make drastic changes on my first pull request

.gitignore
fire-light42: My bad, this makes sense as I missed the "!".

gradle/libs.versions.toml
fire-light42: All other libs can be updated if possible, but android in all their incompetence makes gradle version very tied to the IDE version of Android Studio. Due to this, gradle should lag behind to not force everyone to constantly update, and at the same time make it possible to use with IDEs like intellij, that cant use the newest versions of gradle.

PR #1820, opened 2025-08-11
Always show pinned providers

rockhero1234:


app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
```diff
@@ -415,7 +415,7 @@ class HomeFragment : Fragment() {
                     pinnedphashset = pinnedp.toHashSet()
                     arrayAdapter.clear()
                     val sortedApis = validAPIs
-                        .filter { it.hasMainPage && it.supportedTypes.any(preSelectedTypes::contains) }
+                        .filter {it.hasMainPage && pinnedphashset.contains(it.name) ||  it.supportedTypes.any(preSelectedTypes::contains) }
```
fire-light42: I believe this needs parentheses as you presumably mean:
```kotlin
it.hasMainPage && ( pinnedphashset.contains(it.name) || it.supportedTypes.any(preSelectedTypes::contains) )
```
However this statements gets parsed as:
```kotlin
( it.hasMainPage && pinnedphashset.contains(it.name) ) || it.supportedTypes.any(preSelectedTypes::contains) 
```
Meaning it will match providers that does not have a mainpage given that:
```kotlin
false && false || true // => true
```

rockhero1234:
Done

fire-light42 (approved):

PR #1822, opened 2025-08-11
Fix: Migrate to authdata 

fire-light42:
I made this into a pull request instead of a commit to master, mainly to fully document the changes, and reasoning in a more visible manner. 

1. Better set/get key security
Currently access-token was used, this leaks the access-tokens not by the value but by the key. To fix this in an easy and expandable way, every function now uses AuthData to access the user field. No need to pass by payload.
3. Split up auth into files.
The current structure placed everything in AuthAPI, this should make it easier to navigate. 
4. Added preliminary BackupAPI, will be expanded upon in a future pull request.

5. Use Result instead of Resource.
More correct, and can be wrapped into resource by the caller and not the callee.


fire-light42:
It is also important to note that no developer API used in extensions should break, given that the SyncRepo and AuthAPI is unchanged.

PR #1823, opened 2025-08-12
Upgrade gradle wrapper

diogob003:
* Move plugins from legacy `buildscript` block to `plugins`
* Centralize repository management in `settings.gradle.kts` (default for new A.S. projects)
* Use `alias()` to import plugins info from version catalog `libs.versions.toml`
* Upgrade wrapper 8.13 -> 9.1.0 `gradle wrapper --gradle-version 9.1.0`


gradle/libs.versions.toml
```diff
@@ -1,49 +1,53 @@
+# https://docs.gradle.org/current/userguide/plugins.html#sec:version_catalog_plugin_application
+# https://docs.gradle.org/current/userguide/dependency_versions.html#sec:strict-version
 [versions]
 acraCore = "5.12.0"
+androidGradlePlugin = { strictly = "8.9.3" } # "android in all their incompetence makes gradle version very tied to the IDE version of Android Studio. Due to this, gradle should lag behind"
+androidToolsUtp = "31.12.0"
```
fire-light42: What is this?

build.gradle.kts
```diff
@@ -1,23 +1,30 @@
...
+}
 
-    dependencies {
-        classpath(libs.gradle)
-        classpath(libs.jetbrains.kotlin.gradle.plugin)
-        classpath(libs.dokka.gradle.plugin)
-        // Universal build config
-        classpath(libs.buildkonfig.gradle.plugin)
+// Task to list dependencies updates // https://github.com/ben-manes/gradle-versions-plugin
+tasks.withType<com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask> {
```
fire-light42: Why is this included? Android studio already includes this.

gradle/libs.versions.toml
diogob003: I forgot to delete the comment. "androidToolsUtp" it's `adb` android wifi debugging

build.gradle.kts
diogob003: I don't use Android Studio and I think there are other people who don't either. But I can't remove it without any problems :)

Dexbuddy:
Why isn't this merged yet?

fire-light42:
> Why isn't this merged yet?

There is a lot of changes, and no one else have looked at it. I do not have the extensive knowledge about the build system to merge this in good faith.

diogob003:
@fire-light42 I can try to split it even further. I'll create a new pr that only modifies ```libs.versions.toml```, and let this pull request for changes on the build scripts and Gradle Wrapper.
I did not make any major changes to the build scripts, as most of it was autogenerated when I upgraded the Gradle wrapper version.

Luna712:
For the record I did most of this throughout a few different PRs and was also planning to do the plugins in the version catalog part as well, which I think might be the only piece that is actually still missing from this. I didn't fully realize it was all done in this PR but in my opinion it's also sometimes better to split things to multiple PRs so its better to test and review also.

PR #1825, opened 2025-08-12
feat:added repo icon

rockhero1234:


app/src/main/java/com/lagradost/cloudstream3/ui/settings/extensions/ExtensionsViewModel.kt
```diff
@@ -17,6 +17,7 @@ import com.lagradost.cloudstream3.utils.txt
 import com.lagradost.cloudstream3.utils.Coroutines.ioSafe
 
 data class RepositoryData(
+    @JsonProperty("iconUrl") val iconUrl: String,
```
fire-light42: I think this should be nullable given that it is already nullable in Repository

app/src/main/java/com/lagradost/cloudstream3/ui/settings/extensions/RepoAdapter.kt
```diff
@@ -89,6 +91,11 @@ class RepoAdapter(
                         }
                         mainText.text = repositoryData.name
                         subText.text = repositoryData.url
+                        if(repositoryData.iconUrl.isNotEmpty()){
```
fire-light42: As this is a recycleview entryIcon will needed to be cleared to not keep the old icon. See: https://github.com/recloudstream/cloudstream/blob/1a36bb45238f94ca0de5e0c0444a2ad63d5c6c94/app/src/main/java/com/lagradost/cloudstream3/ui/settings/extensions/PluginAdapter.kt#L90

fire-light42 (changes requested):
Good pull request!

rockhero1234:
Done

Dexbuddy:
What is the reason that only some repo icons are appearing and others' corners are not getting rounded?

IMG_20250813_225149

rockhero1234:
@Dexbuddy There’s no XML styling applied to make the icons round. If rounded appearance is desired, the icons themselves must be designed with rounded shapes and transparent backgrounds

PR #1828, opened 2025-08-14
feat:mark as watched up to this episode

rockhero1234:


app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1640,17 +1663,26 @@ class ResultViewModel2 : ViewModel() {
...
                     setVideoWatchState(click.data.id, VideoWatchState.None)
                 } else {
                     setVideoWatchState(click.data.id, VideoWatchState.Watched)
                 }
-
                 // Kinda dirty to reload all episodes :(
                 reloadEpisodes()
             }
 
+            ACTION_MARK_WATCHED_UP_TO_THIS_EPISODE -> {
```
fire-light42: No option for batch unmark as watches.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1424,6 +1424,29 @@ class ResultViewModel2 : ViewModel() {
...
+    fun markEpisodesAsWatched(episodeIds: Array<Int>) {
+        episodeIds.forEach{id->
+            if (getVideoWatchState(id) != VideoWatchState.Watched) {
+                setVideoWatchState(id, VideoWatchState.Watched)
+            }
+        }
+    }
+
+    fun getEpisodesBySeason(season: Int): HashMap<Int, Array<Int>> {
+        return HashMap(
```
fire-light42: This will return both dub and sub ids for the same entry, is this intended? As "Mark as watched" separates these two.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1424,6 +1424,29 @@ class ResultViewModel2 : ViewModel() {
         _episodeSynopsis.postValue(null)
     }
 
+    fun markEpisodesAsWatched(episodeIds: Array<Int>) {
+        episodeIds.forEach{id->
+            if (getVideoWatchState(id) != VideoWatchState.Watched) {
+                setVideoWatchState(id, VideoWatchState.Watched)
```
fire-light42: This may look fine, but if you mark a long running tv-show with 1k+ episodes this may crash the device, or at least give a huge mem spike. The current setKey and getKey will do a commit on the database for every call, meaning that a hot loop like this will cause many commits. Backup had this same problems before we moved to the Editor class to batch apply values.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1640,17 +1663,26 @@ class ResultViewModel2 : ViewModel() {
...
                 }
-
                 // Kinda dirty to reload all episodes :(
                 reloadEpisodes()
             }
 
+            ACTION_MARK_WATCHED_UP_TO_THIS_EPISODE -> {
+                val (clickSeason,clickEpisode) = click.data.let { (it.season ?: 0) to it.episode }
+                val seasons = getEpisodesBySeason(clickSeason)
+                val seasonsToMark = if (clickSeason == 0) listOf(0) else clickSeason downTo 1
```
fire-light42: The logic here is a bit ad hoc, and hard to follow. Why cant getEpisodesBySeason simply filter this too?

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1424,6 +1424,29 @@ class ResultViewModel2 : ViewModel() {
...
 
+    fun markEpisodesAsWatched(episodeIds: Array<Int>) {
+        episodeIds.forEach{id->
+            if (getVideoWatchState(id) != VideoWatchState.Watched) {
+                setVideoWatchState(id, VideoWatchState.Watched)
+            }
+        }
+    }
+
+    fun getEpisodesBySeason(season: Int): HashMap<Int, Array<Int>> {
```
fire-light42: getEpisodesBySeason takes in a season parameter, but filters on it. There is no obvious reason for this nor is it understood by the function signature. Additionally the name getEpisodesBySeason would presumably mean ResultEpisodes not random integers, so getEpisodesIdsBySeason is more clear.

fire-light42 (changes requested):
While it looks like your code works well, I have some UX, performance and clean code concerns.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
rockhero1234: Added

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
rockhero1234: Fixed that was not intended

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
rockhero1234: I Used Editor class this time

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
rockhero1234: Done

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
rockhero1234: Changed that

fire-light42:
Looks good on a cursory glance, will merge after testing 👍

rockhero1234:
Ok I will check

rockhero1234:
@fire-light42 i think now it will work

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1424,6 +1428,35 @@ class ResultViewModel2 : ViewModel() {
         _episodeSynopsis.postValue(null)
     }
 
+    private fun markEpisodes(context: Context?,episodeIds: Array<String>,watchState: VideoWatchState) {
```
fire-light42: Sorry, but this function still needs to be optimized. markEpisodes will cause ~100ms delay, and while this is ok for things with like 2 seasons, series with 30 seasons will cause the screen to freeze for ~3s. The main reason is again the apply. I would suggest only calling apply after marking *all* episodes, and not only one season.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel2.kt
```diff
@@ -1640,17 +1677,28 @@ class ResultViewModel2 : ViewModel() {
...
                     setVideoWatchState(click.data.id, VideoWatchState.None)
                 } else {
                     setVideoWatchState(click.data.id, VideoWatchState.Watched)
                 }
-
                 // Kinda dirty to reload all episodes :(
                 reloadEpisodes()
             }
 
+            ACTION_MARK_WATCHED_UP_TO_THIS_EPISODE -> {
```
fire-light42: This can be placed in an ioSafe to prevent the screen from freezing

rockhero1234:
@fire-light42 Did above changes and its fluent now.

PR #1831, opened 2025-08-15
fix: Add workaround for 16KB page size alignment on Android 15

fam007e:
This PR addresses an issue where the app is not compatible with devices that have a 16KB page size, which will be a requirement for Android 15.
The issue is caused by several pre-built native libraries that are not aligned to 16KB boundaries. This PR adds a temporary workaround by using the `doNotStrip` packaging option in the `build.gradle.kts` file. This is a temporary measure to ensure compatibility until the upstream libraries are updated with a proper fix.


fire-light42:
This looks like ai hallucinations.

PR #1832, opened 2025-08-15
feat: Enhance security with obfuscation, encryption, and stricter network policies

fam007e:
This pull request introduces a series of crucial security enhancements to the Cloudstream application, addressing several critical and high-severity vulnerabilities identified during a recent security audit.

**Overall Assessment:**
Significant steps have been taken to improve the application's security posture by addressing critical vulnerabilities related to hardcoded secrets, code obfuscation, and data storage. However, several key areas still require attention to achieve a robust security profile.

---

#### **Implemented Fixes:**

1.  **Hardcoded Secrets (Critical)**
    *   **Status**: **ADDRESSED**. The hardcoded API key in `OpenSubtitlesApi.kt` has been removed.
    **Note**: This will disable OpenSubtitles functionality until you implement a secure key management solution (e.g., backend proxy, dynamic key distribution).

2.  **Lack of Code Obfuscation (High)**
    *   **Status**: **ADDRESSED**. `isMinifyEnabled = true` and `isShrinkResources = true` have been enabled for release builds in `app/build.gradle.kts`, which will obfuscate and minify your application's code.
    **Note**: Ensure your `proguard-rules.pro` file is correctly configured to prevent essential code from being removed or obfuscated.

3.  **Insecure Data Storage (Plaintext `lockPin`) (Critical)**
    *   **Status**: **PARTIALLY ADDRESSED**. `EncryptedSharedPreferences` has been implemented in `DataStore.kt` for *new* data storage. This secures sensitive information like `lockPin` going forward.
    *   **Remaining Action**: **Data Migration**: You will need to implement a separate migration strategy to move any existing plaintext sensitive data from regular `SharedPreferences` to the new encrypted storage. This is a complex task that was outside the scope of this session.

4.  **Insecure Communication (Medium)**
    *   **Status**: **ADDRESSED**. `android:usesCleartextTraffic="false"` is set in `AndroidManifest.xml`, with a `network_security_config.xml` created to allow `raw.githubusercontent.com` as an exception.

5.  **Insecure Data Backup (Medium)**
    *   **Status**: **ADDRESSED**. `backup_descriptor.xml` has been modified to explicitly exclude sensitive `SharedPreferences` files from backups.

6.  **Temporary Files (Low)**
    *   **Status**: **ADDRESSED**. `MainActivity.kt` now deletes the `last_error` crash log file immediately after reading it.

7.  **Lack of URL/URI Validation in Media Playback (Medium)**
[...]

PR #1833, opened 2025-08-15
feat: Groundwork for complex security fixes

fam007e:
This PR lays the groundwork for implementing complex security fixes, specifically by adding a placeholder for data migration to EncryptedSharedPreferences.
This is a preliminary step towards addressing the remaining security recommendations essentially about implementing robust code signing for plugins and stronger sandboxing. This is a significant undertaking that requires careful design and implementation which requires more expertise than what I can do. 😔


fire-light42:
See https://github.com/recloudstream/cloudstream/pull/1832#issuecomment-3192844338

PR #1849, opened 2025-08-19
Dual subtitles: long‑press selection

Ughur:
- Adds dual subtitles via long‑press (long‑press “No subtitles” to clear).
- Renders secondary at the top; primary remains at the bottom.
- Keeps secondary in sync with a lightweight periodic update.
- Fixes 403 on subtitle fetch by merging video headers (incl. referer) with subtitle headers.
- Stops the updater on release.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -16,6 +16,7 @@ import androidx.annotation.MainThread
 import androidx.annotation.OptIn
 import androidx.appcompat.app.AlertDialog
 import androidx.media3.common.C.TIME_UNSET
+import androidx.media3.common.C
 import androidx.media3.common.C.TRACK_TYPE_AUDIO
 import androidx.media3.common.C.TRACK_TYPE_TEXT
 import androidx.media3.common.C.TRACK_TYPE_VIDEO
@@ -31,6 +32,7 @@ import androidx.media3.common.VideoSize
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.database.StandaloneDatabaseProvider
 import androidx.media3.datasource.DataSource
+import androidx.media3.datasource.DataSpec
 import androidx.media3.datasource.DefaultDataSource
 import androidx.media3.datasource.DefaultHttpDataSource
 import androidx.media3.datasource.HttpDataSource
@@ -79,6 +81,7 @@ import com.lagradost.cloudstream3.ui.settings.Globals.TV
 import com.lagradost.cloudstream3.ui.settings.Globals.isLayout
 import com.lagradost.cloudstream3.ui.subtitles.SaveCaptionStyle
 import com.lagradost.cloudstream3.ui.subtitles.SubtitlesFragment.Companion.applyStyle
+import com.lagradost.cloudstream3.ui.subtitles.SubtitlesFragment
 import com.lagradost.cloudstream3.utils.AppContextUtils.isUsingMobileData
 import com.lagradost.cloudstream3.utils.AppContextUtils.setDefaultFocus
 import com.lagradost.cloudstream3.utils.Coroutines.ioSafe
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -239,6 +247,27 @@ class CS3IPlayer : IPlayer {
...
+    private var secondaryTicker: Runnable? = null
+
+    private fun startSecondaryTicker() {
+        if (secondaryTicker != null) return
+        secondaryTicker = Runnable {
+            try {
+                exoPlayer?.currentPosition?.let { renderSecondaryAt(it) }
+            } catch (_: Throwable) { }
+            // Re-schedule while active
+            secondaryHandler.postDelayed(secondaryTicker!!, 250L)
```
fire-light42, line 261: This !! can probably crash the app if `secondaryTicker?.let { secondaryHandler.removeCallbacks(it) }` get ran when this code is running.

fire-light42, line 261: Loop every 250ms is an unreliable, and imprecise update strategy. It will completely miss short duration cues, and be misaligned by a few ms for every cue.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1250,6 +1250,30 @@ class GeneratorPlayer : FullScreenPlayer() {
                     }
                 }
 
+                // Long-press on group: quick set secondary. On "No subtitles" clears secondary
+                subtitleList.setOnItemLongClickListener { _, _, which, _ ->
```
fire-light42, line 1254: There is no visual indication of what the secondary subtitles are, and these are applied instantly and not on apply?

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -781,6 +860,94 @@ class CS3IPlayer : IPlayer {
...
+                        if (read == C.RESULT_END_OF_INPUT) break
+                        if (read > 0) out.write(buffer, 0, read)
+                    }
+                    ds.close()
+                    out.toByteArray()
+                }
+
+                SubtitleOrigin.DOWNLOADED_FILE -> {
+                    // Treat url as a file path if possible
+                    val file = File(Uri.parse(sub.url).path ?: sub.url)
```
fire-light42, line 900: We cant use the File class, as it will fail on every device that uses scoped storage, which is every device running on Android 11+.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -781,6 +860,94 @@ class CS3IPlayer : IPlayer {
...
+                    if (file.exists()) file.readBytes() else null
+                }
+
+                SubtitleOrigin.EMBEDDED_IN_VIDEO -> {
+                    // Not supported yet for secondary; requires demuxing from current media source
+                    null
```
fire-light42, line 906: No visual indication that this is not possible.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -781,6 +860,94 @@ class CS3IPlayer : IPlayer {
...
+                }
+            }
+        } catch (t: Throwable) {
+            logError(t)
```
fire-light42, line 910: No visual error message.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1262,6 +1286,23 @@ class GeneratorPlayer : FullScreenPlayer() {
...
 
+                subtitleOptionList.setOnItemLongClickListener { _, _, which, _ ->
+                    val secondary = subtitlesGroupedList
+                        .getOrNull(subtitleGroupIndex - 1)?.value
+                        ?.getOrNull(which)
+                    if (secondary != null) {
+                        player.setSecondarySubtitles(secondary)
+                        showToast(
+                            ctx.getString(R.string.player_loaded_subtitles, secondary.name) +
+                                    " (secondary)",
```
fire-light42, line 1297: No translation? Use strings.xml

fire-light42 (changes requested):
After testing this pull request, I found several issues that would make this feature confusing and inaccessible for everyday users. The main problem is that nothing has any kind of visual identification making it hard to know that this features is even enabled: no selection icon, no error toast, no subtitle sync, no apply. 

How these subtitles are rendered is also non-standard, looping every 250ms very bad, and creates a lot of issues. 

I would much rather that you find some way of merging the cues at parsetime, meaning that media3 takes care of the rest. However there is no clean way of doing this as far as I know.

fire-light42:
😢

PR #1856, opened 2025-08-21
Minor Fix for StreamWish

phisher98:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/StreamWishExtractor.kt
```diff
@@ -162,7 +162,7 @@ open class StreamWishExtractor : ExtractorApi() {
         val pageResponse = app.get(resolveEmbedUrl(url), referer = referer)
 
         val playerScriptData = when {
-            !getPacked(pageResponse.text).isNullOrEmpty() -> getAndUnpack(pageResponse.text)
+            !getPacked(pageResponse.body.toString()).isNullOrEmpty() -> getAndUnpack(pageResponse.body.toString())
```
LagradOst, line 165: Hold your horses, is this really the correct fix? The issue I found with altsaltk thrown an exception to prevent OOM, are you sure this pageResponse is not the movie file, because if that is the case then this will just OOM.

phisher98: It’s working for me without any changes needed but when I checked the logs, I noticed a message suggesting to use Body instead of Text since the content is quite large.

phisher98: image

LagradOst: Yes that is my img, the problem may not be with the .text but that you are accidentally downloading a video file. In this case doing body will cause a crash in the app instead of throwing an exception. So can you double check that requestbody is not the body of a video file.

phisher98: Could you provide the url ?

LagradOst: > Could you provide the url ?

ask altsaltk

phisher98: He was not getting issue with Streamwish TorraStream doens't use any extractor it's A Torrentio Extension

phisher98: Closing this

PR #1860, opened 2025-08-24
feat:shareable movie/tvshow links

rockhero1234:
I had used external js script in html to redirect to custom schema https://github.com/rockhero1234/csredirect cs should deploy of their own.
Thank You!

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -361,11 +361,15 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
                             )
                         }
                     } else if(str.startsWith(APP_STRING_SHARE)){
-                        val data = str.substringAfter("$APP_STRING_SHARE:")
-                        val parts = data.split(":",limit=2)
-                        if(parts.size<2) return false
-                        loadResult(base64Decode(parts[1]),parts[0].replace("%20"," "),"")
-                        return true
+                        try{
+                            val data = String(base64DecodeArray(str.substringAfter("$APP_STRING_SHARE:")),Charsets.UTF_8)
+                            val parts = data.split("?",limit=2)
+                            loadResult(parts[1],parts[0],"")
+                            return true
+                        }catch (e: Exception) {
+                            showToast("Invalid Uri",Toast.LENGTH_SHORT)
+                            return false
+                        }
                     }else if (!isWebview) {
                         if (str.startsWith(DOWNLOAD_NAVIGATE_TO)) {
                             this.navigate(R.id.navigation_downloads)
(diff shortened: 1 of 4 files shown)


phisher98:
Interesting 🤔

fire-light42:
I am unsure about this pull request, as it ties core functionality to an external site. It could be done with a fork like you suggest, and I assume that sending the csshare:// link directly is worse because text renderers will not mark that a real link.

rockhero1234:
@fire-light42 We can't do this without external site.

fire-light42:
The GitHub page build is not working, https://github.com/recloudstream/csredirect/actions/runs/17246643925 I am unable to fully review, and merge it before then.

rockhero1234:
@fire-light42 I tried cloning and building page from another account it worked for me. Maybe you should reclone and build, not able to figure out actual issue

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -358,7 +360,13 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
                                 START_ACTION_RESUME_LATEST
                             )
                         }
-                    } else if (!isWebview) {
+                    } else if(str.startsWith(APP_STRING_SHARE)){
+                        val data = str.substringAfter("$APP_STRING_SHARE:")
+                        val parts = data.split(":",limit=2)
+                        if(parts.size<2) return false
+                        loadResult(base64Decode(parts[1]),parts[0].replace("%20"," "),"")
```
fire-light42: This does not accurately deserialize non ascii text.

Use `String(base64DecodeArray(parts[1]), Charsets.UTF_8)` instead.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
```diff
@@ -810,15 +815,17 @@ open class ResultFragmentPhone : FullScreenPlayer() {
                         resultShare.setOnClickListener {
                             try {
                                 val i = Intent(Intent.ACTION_SEND)
+                                val base64Url = base64Encode(d.url.toByteArray(Charsets.UTF_8))
+                                val encodedUri = URLEncoder.encode("$APP_STRING_SHARE:${d.apiName}:${base64Url}","UTF-8").replace("+","%20")
```
fire-light42: Is there any reason why "apiName" is not encoded with base64? apiName could be any arbitrary string (that might also include ":") so it might not get accurately serialized and deserialized using this string concatenation method.

rockhero1234:
@fire-light42 did the above changes and changed separator from  ":" to "?"

fire-light42:
> @fire-light42 did the above changes and changed separator from ":" to "?"

While this is more correct, I would still like to base64 name and url separately. Currently it is impossible to open links from apis with "?" in the name, and I do not want to break old links when someone makes a bug report about this. 

This preserves the name and url no matter the content. 
```kotlin
val nameBase64 = base64Encode(d.apiName.toString().toByteArray(Charsets.UTF_8))
val urlBase64 = base64Encode(d.url.toByteArray(Charsets.UTF_8))
val encodedUri = URLEncoder.encode("$APP_STRING_SHARE:$nameBase64?$urlBase64","UTF-8")
```

rockhero1234:
@fire-light42 That's good  solution. Done

fire-light42 (approved):
Very good ❤️

PR #1863, opened 2025-08-24
SubtitleHelper: add IETF BCP 47 and OpenSubtitles

diogob003:
* Add IETF BCP 48 tag

* Unify generation of emoji flag + language name

* A lot of manual work to:
  - include OpenSubtitles inconsistent codes
  - remove extinct and with < 1 mi speakers languages
  - include flag emoji to ~100 languages

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
@@ -1,9 +1,19 @@
 package com.lagradost.cloudstream3.utils
 
-import com.lagradost.cloudstream3.mvvm.logError
+import com.lagradost.cloudstream3.Prerelease
 import java.util.Locale
+import kotlin.text.RegexOption.IGNORE_CASE
+
+// If you find a way to use SettingsGeneral getCurrentLocale()
+// instead of this function do it.
+fun getCurrentLocale(): String {
+    return Locale.getDefault().toLanguageTag()
+}
 
 object SubtitleHelper {
+    @Deprecated(
+        "Default language code changed to IETF BCP 47 tag",
+        ReplaceWith("LanguageMetadata(languageName, nativeName, ISO_639_1.ifBlank { ISO_639_2_B }), ISO_639_1, ISO_639_2_B, ISO_639_3, ISO_639_1"))
     data class Language639(
         val languageName: String,
         val nativeName: String,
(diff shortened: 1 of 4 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -1,519 +1,648 @@
...
-        Language639("Yiddish", "ייִדיש", "yi", "yid", "yid", "yid", ""),
-        Language639("Yoruba", "Yorùbá", "yo", "yor", "yor", "yor", ""),
-        Language639("Zhuang", "Saɯ cueŋƅ, Saw cuengh", "za", "zha", "zha", "zha", ""),
-        Language639("Zulu", "isiZulu", "zu", "zul", "zul", "zul", ""),
-    )
-}
+package com.lagradost.cloudstream3.utils
+
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
```
fire-light42: This will fail the autobuild, as "library/src/commonMain" is used for the crossplatform jar build

fire-light42 (commented):
Looks good, might be good to use a hashmap for better performance but I have not profiled it yet. Other than that, I will merge this after testing, looking at extensions and a more in-depth profiling.

diogob003:
I did not change the data structures, but that's a valid concern. I can choose a better structure if you find it necessary (maybe a multiple keys map for languages), but it's not like the lists will keep growing ad infinitum; lang2country has about 200 items, and languages list < 200 items.

fire-light42:
> I did not change the data structures, but that's a valid concern. I can choose a better structure if you find it necessary (maybe a multiple keys map for languages), but it's not like the lists will keep growing ad infinitum; lang2country has about 200 items, and languages list < 200 items.

You are correct that more languages might be added, however that is not my main concern. 

The problem is that there is 5 string checks, on 200 items ran for each subtitle file. Some sources might have 100 subtitles, resulting in up to 100,000 string comparisons. Therefore even if I do not have an exact measurement, this computational complexity does not sound good and is why I want to profile this code.

fire-light42:
> > I did not change the data structures, but that's a valid concern. I can choose a better structure if you find it necessary (maybe a multiple keys map for languages), but it's not like the lists will keep growing ad infinitum; lang2country has about 200 items, and languages list < 200 items.
> 
> You are correct that more languages might be added, however that is not my main concern.
> 
> The problem is that there is 5 string checks, on 200 items ran for each subtitle file. Some sources might have 100 subtitles, resulting in up to 100,000 string comparisons. Therefore even if I do not have an exact measurement, this computational complexity does not sound good and is why I want to profile this code.

I just profiled it, and adding 100 subtitles takes ~100ms so there should not be a big concern, even if it could be optimized to ~20ms with a hashmap.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -1,519 +1,645 @@
...
+                return lang
+        }
+        return null
+    }
+
+    @Deprecated(
+        "Default language code changed to IETF BCP 47 tag",
+        ReplaceWith("fromTagToLanguageName(languageCode)"))
+    fun fromTwoLettersToLanguage(languageCode: String?): String? {
+        return getLanguageDataFromCode(languageCode)?.localizedName()
```
fire-light42: When testing I found that this localizedName messes things up. For example auto selecting subtitle track, or for Addic7ed. The old behavior was english only, so this breaks existing behavior in multiple places.

fire-light42 (changes requested):

app/src/main/AndroidManifest.xml
```diff
@@ -136,7 +137,7 @@
 
                 <data android:scheme="cloudstreamapp" />
             </intent-filter>
-            <intent-filter>
+            <intent-filter android:autoVerify="true">
```
fire-light42: What? Why? This is not relevant at all, considering that the scheme is not `http or https`

app/src/main/AndroidManifest.xml
```diff
@@ -175,7 +176,7 @@
                 <data android:scheme="cloudstreamcontinuewatching" />
             </intent-filter>
 
-            <intent-filter>
+            <intent-filter android:autoVerify="true">
```
fire-light42: We do not own the domain cs.repo, so this is irrelevant and wrong.

fire-light42 (changes requested):

app/src/main/AndroidManifest.xml
diogob003: Fixed.
You're 100% right. Sorry. This shouldn't be here, I accidentally accepted the auto-suggest.

diogob003:
> I would like you to mark all the functions that have been added or changed in signature with `@Prerelease`. This is to inform extension developers, as we want to avoid breaking the stable version of the app.

Good point.

> I am unsure if we need to have deprecation warnings yet, as it should only be applicable after the next stable version.

I'm also unsure about mark as deprecated now. How can i proceed now? 
I think I will revert signature changes
 - fromLanguageToTwoLetters()
 - fromLanguageToThreeLetters()
 - fromTwoLettersToLanguage()
 - fromThreeLettersToLanguage()
And remove `@Deprecated`

fire-light42:
> > I would like you to mark all the functions that have been added or changed in signature with `@Prerelease`. This is to inform extension developers, as we want to avoid breaking the stable version of the app.
> 
> Good point.
> 
> > I am unsure if we need to have deprecation warnings yet, as it should only be applicable after the next stable version.
> 
> I'm also unsure about mark as deprecated now. How can i proceed now? I think I will revert signature changes
> 
>     * fromLanguageToTwoLetters()
> 
>     * fromLanguageToThreeLetters()
> 
>     * fromTwoLettersToLanguage()
> 
>     * fromThreeLettersToLanguage()
>       And remove `@Deprecated`

Please make the api backwards compatible, otherwise we might get some strange crashes. However for the new apis we can just move the Deprecated into a comment so it is easy to reapply when the stable is pushed.

fire-light42:
Making something backwards compatible means making the signature match *exactly*, so fun f(x : Int) and f(x:Int, y:Int = 0) will not be "backwards compatible" even if the IDE marks them as the same.

diogob003:
> Making something backwards compatible means making the signature match _exactly_, so fun f(x : Int) and f(x:Int, y:Int = 0) will not be "backwards compatible" even if the IDE marks them as the same.

Thanks, I was searching about it now. Changing parameters naming or signature can break binary compatibility.

fire-light42 (approved):
Looks good ❤️

PR #1883, opened 2025-09-03
ui:improve 2x overlay btn

rockhero1234:


Dexbuddy:
https://github.com/recloudstream/cloudstream/issues/1884

PR #1894, opened 2025-09-07
Settings: translate languages and use IETF BCP 47 tag

diogob003:
* Rename locales to conform standard
* Use getNameNextToFlagEmoji() to get flag + translated language names
* Use nameNextToFlagEmoji() to unify flag + native name
* Move "All languages" to 1st position while selecting providers language
* Fix language sorting while selecting plugins language filter
* Other small fix related to language setting

Old vs new behavior
Screenshot_20250914-015054


app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsGeneral.kt
```diff
@@ -122,9 +129,25 @@ val appLanguages = arrayListOf(
     Triple("", "اردو", "ur"),
     Triple("", "Tiếng Việt", "vi"),
     Triple("", "中文", "zh"),
-    Triple("\uD83C\uDDF9\uD83C\uDDFC", "正體中文(臺灣)", "zh-rTW"),
+    Triple("", "正體中文(臺灣)", "zh-TW"),
```
fire-light42: If you are removing all override emojis, you might as well make it a pair, or in this case only make it IETF tags as only `third` is used.

diogob003: I was trying to split #1815 into smaller pull requests. And changing this breaks update locales workflow and a lot of places.
But changes applied (￣^￣ )ゞ

fire-light42 (approved):
Looks very good, I will merge after testing.

PR #1895, opened 2025-09-07
Addic7ed, Subdl: use IETF BCP 47 language tag

diogob003:


diogob003:
> Looks good, but will test some before merging.

On my tests Addic7ed was working fine for movies, but not for series. Now it works almost fine with the last changes

fire-light42:
> Now it works almost fine with the last changes

When testing right now it works fine for series? At least for `2nd case: found several or no results. Redirected to $HOST/search.php?search=title`

diogob003:
In some cases, language names were found within others `<tr>`, leading to creation of invalid `SubtitleEntry` and therefore invalid results.

The main problem was here
```kotlin
        // get subtitles links from download page
        return doc.select(".tabel95 .tabel95 tr:has(.language):contains($langName)").mapNotNull { node ->
```
I solved it by adding `:has(.language)`

PR #1896, opened 2025-09-07
help wanted: fix subsource and IETF tag to it

diogob003:


fire-light42:
Anything specific you want help with? I checked myself, but could not find any documentation on this subtitle provider.

diogob003:
There's a notice on their site about a big update that was released in July. I couldn't find any documentation either. So I made this update inspecting their site source and network traffic.

> Anything specific you want help with?

How to test just this provider? I couldn't test it properly.
Also, another opinion and more tests to ensure that I did not make any stupid mistakes :)

It would also be nice if you included in the wiki tips on testing and what tests you do. I'm not used to Android dev and `ProviderTests.kt` is commented.

fire-light42:
> It would also be nice if you included in the wiki tips on testing and what tests you do. I'm not used to Android dev and ProviderTests.kt is commented.

Testing varies between internal feature testing and external extension testing. I do not remember why ProviderTests.kt is commented, however it is now largely redundant because we have no built-in providers. Since extensions load dynamically they are tested dynamically in app, by normal use or by using the testing feature in Settings → Providers → Test all extensions. We should write more on the wiki about it.

> How to test just this provider? I couldn't test it properly.

I find the easiest way of testing app features is including a dummy TmdbProvider. This is done by:

1. Add the Tmdb provider to the provider [list](https://github.com/recloudstream/cloudstream/blob/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt#L61): `threadSafeListOf<MainAPI>(TmdbProvider())`
2. Setting meta load response to [true](https://github.com/recloudstream/cloudstream/blob/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/TmdbProvider.kt#L63)
3. Optionally add a fake video link in the TmdbProvider:
```kt
    override suspend fun loadLinks(
        data: String,
        isCasting: Boolean,
        subtitleCallback: (SubtitleFile) -> Unit,
        callback: (ExtractorLink) -> Unit
    ): Boolean {
        return loadExtractor("https://youtu.be/olV-7gqfrx8", subtitleCallback, callback)
    }
``` 
The TmdbProvider is now usable as a normal provider in app. This makes it possible to debug any issues with breakpoints, while benefiting from realistic provider metadata (such as Imdb ID).  

If you are able to contribute effective automated tests that would be much appreciated. It would be somewhat easy to create a test for subtitle providers and it would be very useful.

PR #1899, opened 2025-09-09
feat(Emulator UI): Click anywhere to hide Episodes

KingLucius:


fire-light42 (approved):
Works fine when testing, might cause focus issues, but I have not found any.

KingLucius:
> Works fine when testing, might cause focus issues, but I have not found any. 

I will test deeply for this 👍

PR #1901, opened 2025-09-09
Update AniListApi.kt

GeorgCantor:
Single-pass processing - Eliminated redundant string iterations

Direct character checks - Replaced slow regex with efficient range comparisons

StringBuilder usage - Used buildString for optimal memory management

No intermediate strings - Avoids creating temporary string objects

O(n) time complexity - Improved from previous O(2n) approach

Reduced object allocation - No Regex object creation overhead

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/AniListApi.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/AniListApi.kt
@@ -192,10 +192,13 @@ class AniListApi : SyncAPI() {
 
         const val ANILIST_CACHED_LIST: String = "anilist_cached_list"
 
-        private fun fixName(name: String): String {
-            return name.lowercase(Locale.ROOT).replace(" ", "")
-                .replace("[^a-zA-Z0-9]".toRegex(), "")
-        }
+        private fun fixName(name: String): String = buildString {
+    		name.lowercase(Locale.ROOT).forEach { char ->
+        		if (char in 'a'..'z' || char in '0'..'9') {
+            		append(char)
+        		}
+    		}
+		}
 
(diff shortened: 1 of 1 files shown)


fire-light42:
While we appreciate contributions, this is not one. The complexity is redundant on such a small function. In this case the maintainability is more important. Adding more or statements for future additions would not be sustainable, making the tiny efficiency gains moot. Please make meaningful contributions in the future, and I would be happy to merge them. 

Yes, it would be possible to only instantiate the regex object once, but your contribution looks like contribution farming.

GeorgCantor:
Both approaches produce identical results for spaces and special characters, but the new version is more efficient with single-pass processing and no regex overhead.
Read the code more carefully genius haha

GeorgCantor:
Although what can you expect from a person who has only been programming for two years?

GeorgCantor:
This implementation has several performance and efficiency issues:

Double string processing - Two consecutive replace() calls create unnecessary intermediate string objects

Redundant space removal - The regex [^a-zA-Z0-9] already removes spaces, making the explicit replace(" ", "") redundant

Regex overhead - Regular expressions are computationally expensive for simple character filtering operations

Inefficient memory usage - Multiple temporary string allocations during processing

fire-light42:
> Read the code more carefully genius haha

Note the edit time before your comment. 

> This implementation has several performance and efficiency issues:
> Double string processing - Two consecutive replace() calls create unnecessary intermediate string objects
>Redundant space removal - The regex [^a-zA-Z0-9] already removes spaces, making the explicit replace(" ", "") redundant
> Regex overhead - Regular expressions are computationally expensive for simple character filtering operations
> Inefficient memory usage - Multiple temporary string allocations during processing

Addressed in: https://github.com/recloudstream/cloudstream/pull/1901#issuecomment-3272433718

PR #1907, opened 2025-09-12
feat(Player): Pinch-to-Zoom support

KingLucius:
- [x] Pinch-to-Zoom in/out.
- [x] Pan to Move while in Zoom mode.
~~- [ ] Double-tap to reset is not working~~


KingLucius:
Removed the Double-tap to reset

Dexbuddy:
> Removed the Double-tap to reset

This new pinch-to-zoom feature in your PR is awesome. One suggestion though, using double-tap to reset the zoom might confuse users since double-tap usually skips forward and backward 10 seconds. Maybe add a temporary reset button instead to keep the playback controls clear?

KingLucius:
> > Removed the Double-tap to reset
> 
> This new pinch-to-zoom feature in your PR is awesome. One suggestion though, using double-tap to reset the zoom might confuse users since double-tap usually skips forward and backward 10 seconds. Maybe add a temporary reset button instead to keep the playback controls clear?

I removed the Double-tap, pinch out to back to normal to avoid conflict with other controls

Dexbuddy:
Hey @KingLucius, you could add a temporary button like this for restore. I’ve attached an image below to show what it could look like. It appears when you zoom in, and clicking it to restore hides it again. You mentioned that pinch out to back would return to normal, so this could work alongside that too.

1757760hh66


Great work on your efforts so far! If this isn’t possible right now, we can consider it for the future.

KingLucius:
I think it will be OK for temporary zooming but if someone uses zoom all the time it will be so annoying, I think we can change the size mode button status to be "reset" in case of zooming

Dexbuddy:
Just a quick follow-up. Since pinch-to-zoom in (bigger) is already implemented, the ability to pinch-to-zoom out (smaller) should also be there. A reasonable minimum limit of exactly 25% would ensure users have full control and better accessibility. This feature can be very helpful.

KingLucius:
Any use case for zooming out?

Dexbuddy:
> Any use case for zooming out?

For zooming out to minimum 25%, it’d help with curved screen phones to avoid edge distortion, make multitasking easier on tablets, and fix aspect ratio issues for some videos. It’s a small tweak but adds flexibility without impacting performance. What do you think?

KingLucius:
I will add a callback for Zoom state, Reset zoom fun & Scale number to be used in the UI

KingLucius:
> > Any use case for zooming out?
> 
> For zooming out to minimum 25%, it’d help with curved screen phones to avoid edge distortion, make multitasking easier on tablets, and fix aspect ratio issues for some videos. It’s a small tweak but adds flexibility without impacting performance. What do you think?

@fire-light42 what do you think?

Dexbuddy:
> Any use case for zooming out?

For zooming out to 25%, it’d help with curved screen phones to avoid edge distortion, make multitasking easier on tablets, and fix aspect ratio issues for some videos. It’s a small tweak but adds flexibility without impacting performance.

Dexbuddy:
I saw this in another video player app (PLAYit), and thought it would be helpful.

When the user is pinch-to-zooming in/out or panning in a video, a subtle overlay should appear in the corner showing the current zoom percentage (e.g., “188%”).

https://github.com/user-attachments/assets/ff21ecd0-c5e2-442f-80a3-75fab1020f06

This overlay should be visible only during active interaction and fade out smoothly after about 1–2 seconds of inactivity.  
If the user single-taps to bring up player controls before the fade finishes, the overlay should hide immediately so the regular controls (title, play/pause, 10s back/forward, etc.) can appear.

fire-light42:
The zoom indicator does not fit with the look of the app at all, nor a zoomout button. Moreover that zooming out more than 100% should reset it to 100% as settings already has a overscan option to zoom out the app more than 100%. For reference, YouTube does not allow zooming out more than 100%, so it would be strange to allow it, and make the UI less intuitive.

Dexbuddy:
Understood, overscan adjusts video beyond 100%, which helps with fit. Still, zooming less than 100% could be quick for aspect ratio tweaks on tall/curved screens.

KingLucius:
> The zoom indicator does not fit with the look of the app at all, nor a zoomout button. Moreover that zooming out more than 100% should reset it to 100% as settings already has a overscan option to zoom out the app more than 100%. For reference, YouTube does not allow zooming out more than 100%, so it would be strange to allow it, and make the UI less intuitive. 

So no need for the callback?
I was thinking about changing icon for "screen modes" to be changed to "reset zoom" in case of scale >1

fire-light42:
> > The zoom indicator does not fit with the look of the app at all, nor a zoomout button. Moreover that zooming out more than 100% should reset it to 100% as settings already has a overscan option to zoom out the app more than 100%. For reference, YouTube does not allow zooming out more than 100%, so it would be strange to allow it, and make the UI less intuitive.
> 
> So no need for the callback?

I do not think so. However testing your pull request I found some issues with how it interacts with the current gestures like 2x, hold to seek and such. It also zooms weird, unlike YouTube.

KingLucius:
What is the issue you have found?
I tried YT zoom now and it's the same?

You mean that sometimes after finish pan it seeks some seconds?

I have another idea to implement this, let me try it

KingLucius:
Please guys check the new way of zooming & panning

Only one issue I have now, after zoom or pan the UI appears
Tomorrow I will check it

rockhero1234:
We should implement like mx player it should not cut during zoom, instead whole screen. 

https://github.com/user-attachments/assets/60267a86-bd48-4ae9-91ae-cc36ca18e55b

KingLucius:
> We should implement like mx player it should not cut during zoom, instead whole screen. 
> 
> https://github.com/user-attachments/assets/60267a86-bd48-4ae9-91ae-cc36ca18e55b
> 
> 
> 

What do you mean by cut during zoom?

rockhero1234:
Currently it only zoom in this space it should be whole screen
markup_1000042261

KingLucius:
It depends on the video itself, the zoom feature is between 1x to 4x
1x is the original aspects whatever it is

fire-light42:
> It depends on the video itself, the zoom feature is between 1x to 4x 1x is the original aspects whatever it is

The problem is that the video will not fill the entire screen even if zoomed in. I assume that is because the exo_content_frame does not fill the screen in "fit to screen" mode, meaning that any zoom will be done inside the content area.

KingLucius:
> > It depends on the video itself, the zoom feature is between 1x to 4x 1x is the original aspects whatever it is
> 
> The problem is that the video will not fill the entire screen even if zoomed in. I assume that is because the exo_content_frame does not fill the screen in "fit to screen" mode, meaning that any zoom will be done inside the content area. 

OH you are right, I didn't noticed that
I will see what I can do about it

KingLucius:
OK I depend on video area for calculations, I should depend on screen area and apply on video for zoom
It's doable

HeroJack99:
l'm excited to see this feature in CloudStream.

HeroJack99:
Screenshot_2025-10-31-12-11-18-71

PR #1920, opened 2025-09-19
VidStack Improvement

phisher98:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/VidStack.kt
```diff
@@ -52,6 +53,20 @@ open class VidStack : ExtractorApi() {
...
+        val subtitleSection = Regex("\"subtitle\":\\{(.*?)\\}").find(decryptedText)?.groupValues?.get(1)
+
+        subtitleSection?.let { section ->
+            subtitlePattern.findAll(section).forEach { match ->
+                val lang = match.groupValues[1]
+                val rawPath = match.groupValues[2].split("#")[0]
+                if (rawPath.isNotEmpty()) {
+                    val path = rawPath.replace("\\/", "/")
+                    val subUrl = "$mainUrl$path"
+                    subtitleCallback(SubtitleFile(lang, fixUrl(subUrl)))
```
fire-light42, line 66: You concat "$mainUrl$path" then call fixUrl? Should it not be fixUrl(path)?

phisher98: I think I tried it but it didn't worked

Issue #1854, opened 2025-08-21
"Play Mirror" button


krrishkap:
### Describe your suggested feature

Just like the "Download Mirror" option or the "Play in vlc" thing it would be nice to get an option where we could select the exact stream . 

We already have "chromecast mirror" so it shouldn't be a completely new thing that the dev would've to implement from ground up.

Requesting this feature coz even after using the priority system CS plays dv/hdr content sometimes , since the player has some weird issues currently with dv/hdr content it would be easier to dodge that with this option .

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

LeopoldoDelgadillo:
I would like to add this, if someone can assign me.

diogob003:
> I would like to add this, if someone can assign me.

You don't need to wait assignment. You can create a fork of cloudstream and modify it. Once it's ready, create a pull request to CloudStream repository

LeopoldoDelgadillo:
PR is done, let me know if there's any issue.

PR #1926, opened 2025-09-21
"Play mirror" function on play context window.

LeopoldoDelgadillo:
Added "Play mirror" function on context window when hold pressing play on any media (play mirror allows you to select a source before loading the media).

Function designed to address issue #1854

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
@@ -38,7 +38,7 @@ import java.util.Locale
  * @see VideoClickActionHolder
  */
 const val ACTION_PLAY_EPISODE_IN_PLAYER = 1
-
+const val ACTION_PLAY_MIRROR = 2
 const val ACTION_CHROME_CAST_EPISODE = 4
 const val ACTION_CHROME_CAST_MIRROR = 5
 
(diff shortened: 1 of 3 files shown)


LeopoldoDelgadillo:
I moved the implementation into a VideoClickAction and removed it from ResultViewModel2.

rockhero1234:
Just one suggestion it would be nice to have a setting which says always ask for specific stream when clicked played. It will be helpful.

fire-light42:
> Just one suggestion it would be nice to have a setting which says always ask for specific stream when clicked played. It will be helpful.

We already have that?

app/src/main/java/com/lagradost/cloudstream3/actions/temp/PlayMirrorAction.kt
```diff
@@ -0,0 +1,86 @@
...
+                override suspend fun generateLinks(
+                    clearCache: Boolean,
+                    sourceTypes: Set<ExtractorLinkType>,
+                    callback: (Pair<ExtractorLink?, ExtractorUri?>) -> Unit,
+                    subtitleCallback: (SubtitleData) -> Unit,
+                    offset: Int,
+                    isCasting: Boolean
+                ): Boolean {
+                    index?.let { callback(result.links[it] to null) }
+                    result.subs.forEach { subtitle -> subtitleCallback(SubtitleData(subtitle.originalName,subtitle.nameSuffix,subtitle.url,subtitle.origin,subtitle.mimeType,subtitle.headers,subtitle.languageCode))
```
fire-light42: Why are you reconstructing the SubtitleData?

rockhero1234:
> We already have that?

In Prefered video player there can be option for extract stream internal player

fire-light42:
> > We already have that?
> 
> In Prefered video player there can be option for extract stream internal player

I do not entirely understand, adding this as a videoclickaction in allVideoClickActions automatically makes it an option in "Prefered video player". Is that what you mean?

rockhero1234:
> I do not entirely understand, adding this as a videoclickaction in allVideoClickActions automatically makes it an option in "Prefered video player". Is that what you mean?

Ohh! I didn't see that

app/src/main/java/com/lagradost/cloudstream3/actions/temp/PlayMirrorAction.kt
LeopoldoDelgadillo: I think it was around 4am and wasn't thinking straight, forgot I was taking already the subtitleData element complete. I went ahead and changed it to just put the SubtitleData element from the list.

PR #1934, opened 2025-09-26
Feature: support android 13+ language per-app

diogob003:
It needs more testing. Feel free to continue the work if you want to.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
@@ -21,8 +21,10 @@ import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.MainThread
 import androidx.annotation.StringRes
 import androidx.appcompat.app.AppCompatActivity
+import androidx.appcompat.app.AppCompatDelegate
 import androidx.appcompat.widget.SearchView
 import androidx.core.content.ContextCompat
+import androidx.core.os.LocaleListCompat
 import androidx.core.view.children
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
fire-light42, line 231:
```suggestion
    fun Context.updateLocale() {
        val settingsManager = PreferenceManager.getDefaultSharedPreferences(this)

        val settingsLocale = settingsManager.getString(getString(R.string.locale_key), null)

        val localeCode = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU)
            AppCompatDelegate.getApplicationLocales().get(0)?.toLanguageTag() ?: settingsLocale
        else
            settingsLocale

        setLocale(this, localeCode)
    }
```

fire-light42: I am thinking we use getApplicationLocales as the primary source of truth as to remove inconsistencies when setting it in android settings first and then app settings.

fire-light42 (commented):
Good pull request. 
We just need to figure out how to handle language regions such as English (New Zealand) which gets the language code en-NZ.

SittingDuck999:
@fire-light42 sorry for the ping but will this be merged anytime soon?

CranberrySoup:
> @fire-light42 sorry for the ping but will this be merged anytime soon?

Looks like there is still some work to do before this can be merged.

PR #1938, opened 2025-09-28
feat(library):scroll top on sort

rockhero1234:
scroll top on contentupdate


rockhero1234:
When we change the sorting method(it was intended to see something at top) it just scroll up.

fire-light42:
> When we change the sorting method(it was intended to see something at top) it just scroll up.

Makes perfect sense

PR #1942, opened 2025-10-02
Feat: New TV UI for Homepage

fire-light42:
Updated the TV UI on the homepage for better UX. Removed the play and info button in favor of focusing the entire container, making it both easier to navigate left and right, but also making it easier to visually identify the focus. The container now acts as the new info button.


KingLucius:
Nice ❤️
I loved it so much 🙏🙏

PR #1949, opened 2025-10-03
Feat: Infinite scrolling on quicksearch and trakt

fire-light42:
+ Implements the newish page search for quick-search
+ Updated TraktProvider, mainly for testing
+ Fixed distinct on expand
+ Fixed SearchAdaptor issue, might have caused bugs to appear in other places where it is used.

Any testing would be appreciated ❤️

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/quicksearch/QuickSearchFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/quicksearch/QuickSearchFragment.kt
@@ -16,6 +16,7 @@ import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.ViewModelProvider
 import androidx.recyclerview.widget.GridLayoutManager
+import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.bottomsheet.BottomSheetDialog
 import com.lagradost.cloudstream3.APIHolder.getApiFromNameNull
 import com.lagradost.cloudstream3.CommonActivity.activity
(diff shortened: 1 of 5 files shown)


KingLucius:
I tried quick search but it doesn't infinite scrolling. 
I use my own test plugin that's only inherit TraktProvider

fire-light42:
> I tried quick search but it doesn't infinite scrolling. I use my own test plugin that's only inherit TraktProvider

Remember that it will not be able to expand quickSearch results, only after you have pressed search does it work.

fire-light42:
> I tried quick search but it doesn't infinite scrolling. I use my own test plugin that's only inherit TraktProvider

For quick and dirty testing you can simply do `val allProviders = threadSafeListOf<MainAPI>(TraktProvider())`

phisher98:
Previous PR for infinite scroll for page didn't work on Extension Search and only works in Global Search it's a bug that's not fixed yet

Is that fixed in this ?

KingLucius:
> > I tried quick search but it doesn't infinite scrolling. I use my own test plugin that's only inherit TraktProvider
> 
> Remember that it will not be able to expand quickSearch results, only after you have pressed search does it work. 

It seems that I used a bad query word that only has 1 page of results 🤣😂

Works 👍

KingLucius:
> Previous PR for infinite scroll for page didn't work on Extension Search and only works in Global Search it's a bug that's not fixed yet
> 
> Is that fixed in this ?

This PR adds quickSearch support

fire-light42:
This might be a bit confusing but, there is two things in CloudStream named "quickSearch", this is about the fragment called quickSearch and not the function in mainapi called quickSearch. 

To specify what this pull request adds, this is about the search button in the homepage that allows you to search a single provider. If the provider implements pagination in the search, you can scroll down and view more results. However it will not work if you are in the middle typing the query (has the keyboard up when scrolling down), or if the provider has no more search results.

Turhvjbufv:
Does it show text or something that the provider doesn't have more search results? 
If not could it?

fire-light42:
> Does it show text or something that the provider doesn't have more search results? If not could it?

No, that would be a separate issue/pull request.

PR #1952, opened 2025-10-04
feat(TV UI): Crop the Main banner from the Top

KingLucius:
- Change Main banner crop mode to Top instead on Center crop
- Fix Focus up of main buttons cause scrolling of main banner


fire-light42:
❤️

Issue #1892, opened 2025-09-06
Hide controls/title when adjusting volume or brightness


Dexbuddy:
### Describe your suggested feature

Just like the recent change with the 2x indicator #1885 , it would be better if the **title and controls disappear right away** when adjusting volume or brightness using the side sliders. At the moment, if the controls are visible, they stay on screen while the slider is being used, which feels a bit distracting. But when the controls are already hidden, everything looks fine as it is.

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

KingLucius:
Totally agree, it will be niche

Dev-Aditya-More:
Hey. I want to work on this one, pls assign this to me

Dexbuddy:
> Hey. I want to work on this one, pls assign this to me

You don't have to wait for an assignment. Feel free to fork CloudStream, make your modifications, and then submit a pull request to the main CS repository when it's complete.

Sovan22:
@Dexbuddy, you’ll see that I’ve made a pull request for this. I’ve also made the controls reappear after the gesture finishes if the player is paused.

PR #1954, opened 2025-10-07
Enhancement : hide controls during volume/brightness gestures and restore on pause

Sovan22:
This change updates the player UI so that when a user adjusts volume or brightness through gestures, the on-screen controls are hidden during the interaction. Once the gesture completes, controls are restored if the player is in a paused state.

Testing:
Verified that controls hide and reappear correctly for both volume and brightness gestures.
Checked consistency on pause, play, and resume transitions.

References:
Issue: #1892

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -110,7 +110,7 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
 
     // state of player UI
     protected var isShowing = false
-    private var showUiAfterOverlayHide = false
+    private var uiShowingBeforeGesture = false
     protected var isLocked = false
 
     protected var hasEpisodes = false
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1202,6 +1223,14 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
                         currentClickCount = 0
                     }
 
+                    // If we hid the UI for a gesture and playback is paused, show it again
+                    if (!player.getIsPlaying()) {
+                        val didGesture = currentTouchAction != null || currentLastTouchAction != null
+                        if (didGesture) {
+                            showUiAfterOverlayHide = true
```
fire-light42: I do not understand the use of showUiAfterOverlayHide at all. The current behavior is identical if this is removed? 

I assume it is to prevent it from showing the UI mid swipe if the finger is very still for some amount of time. However if that is the case then the current code has a bug, as if you do any other gesture like swiping for time, it will also have this issue.

Sovan22: Yes, you're right, the variable is not needed. It was to ensure that I remembered to show the UI after the 1.5s fade-out delay, but I also called the showPlayerUI function in the progressBarHideRunnables, which already ensures what I intended. However, as you mentioned, since the UI should show instantly, none of this is needed. I will make the necessary changes in the meantime

PR #1958, opened 2025-10-07
Minor UI changes in dark and amoled themes

doGior:
I just noticed my ide formatted the whole file, sorry for the useless changes

Btw these are the changes:

Screenshot_20251008-010135_CloudStream Debug
Dark mode

Screenshot_20251008-010202_CloudStream Debug
Amoled

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
@@ -6,6 +6,7 @@ import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
+import androidx.core.view.setPadding
 import androidx.preference.PreferenceManager
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
```diff
@@ -180,6 +182,20 @@ class EpisodeAdapter(
...
+            if (isLayout(PHONE or EMULATOR)) {
+                val context = binding.root.context
+                val settingsManager = PreferenceManager.getDefaultSharedPreferences(context)
+                val theme =
+                    settingsManager.getString(
+                        context.getString(R.string.app_theme_key),
+                        "AmoledLight"
+                    )
+                if (theme == "Amoled") {
+                    binding.episodeHolderLarge.getChildAt(0).setPadding(5)
```
fire-light42: I do not quite understand this like, would it not make more sense to do:
```kotlin
binding.episodeHolderLarge.radius = 0.0f
binding.episodeHolder.setPadding(0)
```

doGior: I've only worked with Jetpack compose, so I kinda guessed what to do to change the ui. I'll try that later

app/src/main/java/com/lagradost/cloudstream3/ui/result/EpisodeAdapter.kt
```diff
@@ -180,6 +182,20 @@ class EpisodeAdapter(
             binding.episodeHolder.layoutParams.width = setWidth
 
 
+            if (isLayout(PHONE or EMULATOR)) {
```
fire-light42, line 185: Would it be possible to move the logic out every bind, and only calculate it once?

doGior: I thought of that, but again I have no idea how all this works. I've only made ui with compose

doGior:
Now I feel like the episodes are too much on the left. Or the season dropdown is too much on the right. Either way it's still better than before

ByAyzen:
Absolutely. I also noticed that long time ago, its looks bad.

PR #1960, opened 2025-10-08
Add pagination support to TMDB metaprovider

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/CrossTmdbProvider.kt
```diff
@@ -62,8 +63,12 @@ class CrossTmdbProvider : TmdbProvider() {
...
     }
 
-    override suspend fun search(query: String): List<SearchResponse>? {
-        return super.search(query)?.filterIsInstance<MovieSearchResponse>() // TODO REMOVE
+    override suspend fun search(query: String, page: Int): SearchResponseList? {
+        // TODO REMOVE
+        return super.search(query, page)
+            ?.items
+            ?.filterIsInstance<MovieSearchResponse>()
+            ?.toNewSearchResponseList()
```
Luna712, line 71: Just for the record I did it like this, using items then filtering only movie results then back to SearchResponseList just to keep the exact current behavior. Not sure if this should be done a different way.

PR #1962, opened 2025-10-08
Deprecate public constructor for ExtractorLink

Luna712:


Luna712:
> It is good to phase out the old constructors, however, ExtractorLink is such a fundamental part of extensions that it is likely to be used many times in every single repository. This means that the error is likely to force many changes for a lot of extension developers.
> 
> Because of this it would be better to start with a DeprecationLevel.WARNING to make the transition more gradual for extension developers. In a couple of months we can make it DeprecationLevel.ERROR.
> 
> Generally I want to make the extension developer experience as good as possible.

@fire-light42 

Good point. Thanks, and done.

PR #1964, opened 2025-10-08
Use the newSubtitleFile method

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1092,14 +1092,16 @@ fun TvType.isAnimeOp(): Boolean {
 /** Data class for the Subtitle file info.
  * @property lang Subtitle file language.
  * @property url Subtitle file url to download/load the file.
+ * @see newSubtitleFile
  * */
 @ConsistentCopyVisibility
-data class SubtitleFile private constructor(
+data class SubtitleFile internal constructor(
```
Luna712: Changed to internal so that newSubtitleFile can properly access it.

fire-light42: I do not understand this change. Making it internal is more public than private, and the constructor newSubtitleFile is using is not the "private" one.

Luna712: The comment for the other constructor says it will be deprecated so newSubtitleFile has been changed to use the "non deprecated" version (with headers) that means the constructor it uses is the private one, but its in a different class scope so in order to use that constructor there we do indeed make it more public then private is, but using internal still limits to at least the same gradle project and so that newSubtitleFile can at least use that constructor.

fire-light42: We regularly use the deprecated constructor in our new* functions. The current "private constructor" fits best for this case.

Luna712: We dont usually use backwards compatible deprecated constructors that I thought would eventually be removed do we? I thought we usually use the deprecated primary constructors which will never be dropped but discouraged whereas the other backwards compatibility deprecated constructors could eventually be removed down the line, od am I totally misunderstanding that?

fire-light42: "backwards compatibility deprecated constructors" will not be removed anytime soon as old extensions still rely on them. However, at the meantime we still use them to not expose any other functions to anyone, including the library itself. This is important, as it forces pull requests to also use the correct API. 

When, and if they do get removed we will change the new methods to use the correct system. Deprecating methods simply means that we can make breaking constructors sooner, but as the MainAPI is in the same file as the new constructors we can always switch over both the new constructors to use the correct api at the same time.

Luna712: Fair enough, done.

PR #1966, opened 2025-10-09
feat(TV UI): Add Spaces for cards to match the main banner

KingLucius:
- Add margin to Continue, Bookmark & main recyclers to match Hero Banner.
- Make the scrolled (Hidden) card partially visible for better UX (so the user can differentiate if the recycler is already scrolled)

Test:
- Emulator & Real Android TV
- Right-To-Left Languages UI

### DIFF
--- a/app/src/main/res/layout/homepage_parent_tv.xml
+++ b/app/src/main/res/layout/homepage_parent_tv.xml
@@ -13,7 +13,7 @@
         android:layout_height="wrap_content"
         android:layout_marginStart="10dp"
         android:layout_marginEnd="0dp"
-        android:padding="12dp"
+        android:padding="10dp"
         tools:text="Trending" />
 
     <androidx.recyclerview.widget.RecyclerView


app/src/main/res/layout/fragment_home_head_tv.xml
```diff
@@ -253,7 +254,7 @@
             android:layout_marginStart="@dimen/navbar_width"
             android:layout_marginEnd="0dp"
             android:background="?android:attr/selectableItemBackground"
-            android:padding="12dp"
+            android:padding="10dp"
```
fire-light42, line 257: I do not understand this change. Did you forget to also change it in the nested recycleview? Otherwise it is mismatched with the other WatchHeaderText.

KingLucius: Wait a min, I thought it is a mistake
Will review

KingLucius: Changed the child padding to 10 also to match  the texts on banner and continue watching

fire-light42:
1. Why did you add margins instead of padding? It just causes weird cutoffs on both the continue watching and bookmarked content. 

image

2. Why did you change the overscroll behavior? If you meant to do this, then change it for all horizontal scrollviews including "continue watching and bookmarked". I would do this in code to specifically target TV using `isLayout(TV)` instead of embedding it in the xml that is shared with emulator.

3. Can you explain this statement "Make the scrolled (Hidden) card partially visible for better UX (so the user can differentiate if the recycler is already scrolled)"

PR #1969, opened 2025-10-09
Make constructor for SearchResponseList internal

Luna712:
This was never not deprecated and doesn't exist in current stable ar all. Doesn't need to be used outside library module.

Would supersede #1961 also.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -1154,11 +1154,10 @@ data class HomePageList(
 /** Data class for the Search results.
  * @property items list of [SearchResponse] items that will be added to the search row.
  * @property hasNext if there is a next page or not.
+ * @see newSearchResponseList
  * */
 @Prerelease
-data class SearchResponseList
-@Deprecated("Use newSearchResponseList method", level = DeprecationLevel.WARNING)
-constructor(
+data class SearchResponseList internal constructor(
     val items: List<SearchResponse>,
     val hasNext: Boolean = false
 )


Luna712:
> I would not make the method invisible by making it `internal constructor`. Instead if it is marked as error, then people who might not be familiar with the API can still be pushed into using the correct API by the error message.

@fire-light42 

Something ive been thinking about doing is adding an annotation that requires optin to use by using something like `@DisallowedApiMethod` or something (and that could give a certain message in IDE etc... with `@RequiresOptIn(level = RequiresOptIn.Level.ERROR, message = "Some message I haven't figured out yet")` in the annotation)  and then if you ever use it add like `@OptIn(DisallowedApiMethod::class)`. That would allow better documentation as well IMO. Not sure on exact annotation name yet but just what I was thinking today and was wondering your thoughts on it as well?

fire-light42:
> > I would not make the method invisible by making it `internal constructor`. Instead if it is marked as error, then people who might not be familiar with the API can still be pushed into using the correct API by the error message.
> 
> @fire-light42
> 
> Something ive been thinking about doing is adding an annotation that requires optin to use by using something like `@DisallowedApiMethod` or something (and that could give a certain message in IDE etc... with `@RequiresOptIn(level = RequiresOptIn.Level.ERROR, message = "Some message I haven't figured out yet")` in the annotation) and then if you ever use it add like `@OptIn(DisallowedApiMethod::class)`. That would allow better documentation as well IMO. Not sure on exact annotation name yet but just what I was thinking today and was wondering your thoughts on it as well?

That would be useful for the Prerelease tag, but I would not put it on these methods.

Luna712:
#1961 now makes this just use ERROR deprecation level for now.

PR #1978, opened 2025-10-10
Require optin for prerelease

Luna712:
This should help extensions to maintain compatibility with stable as well. We also globally opt-in within the app itself to prevent needing so many `@OptIn` that would be removed every stable and serves no usage within the app itself anyway.


fire-light42:
Very good pull request! 

We sorely need more stable compatibility, but an instant `RequiresOptIn.Level.ERROR` may be too harsh on extension developers. They may simply annotate everything with OptIn without consideration for compatibility just to make it build. I think it would be better to start with WARNING and keep that for a month or two.

fire-light42:
On the other hand, we may merge this just after the next stable release when every @Prerelease annotation is removed and there will be no breakage. Or switch to Error then.

Luna712:
> On the other hand, we may merge this just after the next stable release when every @prerelease annotation is removed and there will be no breakage.

Thats true as well. Maybe switch to ERROR when @prerelease is removed for next stable?

PR #1983, opened 2025-10-11
Cache dependencies in builds

Luna712:
When testing this with Artifact Build it improved build speed from ~6 minutes to <2 minutes


fire-light42:
Cool, this will significantly improve the ci.

PR #1986, opened 2025-10-11
No gradle cache on prerelease for now

Luna712:
I have a way to fix it but for now let's just remove it.


Luna712:
@fire-light42

PR #1987, opened 2025-10-11
Bump gradle and AGP

Luna712:
This is in preparation for more upgrades including kotlin 2.2.20 which is required by many other package upgrades as well, and preparing for targetSdk 36 (A16) testing and upgrade.

PR #1995, opened 2025-10-11
Bump coil to 3.3.0

Luna712:
Can not be merged until both #1987 and #1989 are merged


fire-light42:
https://coil-kt.github.io/coil/faq/#how-do-i-target-java-8-or-java-11 ?

fire-light42:
I guess we do not use compose, so we should be fine.

Luna712:
> https://coil-kt.github.io/coil/faq/#how-do-i-target-java-8-or-java-11 ?

We dont use coil-compose, I dont think? it says 8 works for most of coil. But on that subject I do think it *might* be time for us to target Java 11, but I don't think it is needed here.

PR #1996, opened 2025-10-11
Readd gradle cache on prerelease

Luna712:
Now that all packages are up-to-date we should be able to readd this without any problem.


Luna712:
@fire-light42 this should be okay now

PR #2000, opened 2025-10-12
Improve support for building with Java 25

Luna712:
I've tested that building works with Java 17, Java 21, and Java 25 with this.


Luna712:
I've changed this PR to only add `--enable-native-access=ALL-UNNAMED`, and not bump anything in workflows because I think going to Java 25 was to much and we should only go to Java 21 in #2148 instead. So this was basically split into this PR, #2146, and #2148.

PR #2001, opened 2025-10-12
Bump newpipeextractor to v0.24.8

Luna712:


Luna712:
I accidentally forgot this one when doing other bumps... whoops...

PR #2014, opened 2025-10-17
Enable auto rotate preference by default

Luna712:
It makes sense to adjust the screen orientation based on the orientation of the video by default. Personally, I don't even see the need for this preference to exist but if it must exist it being default enabled just makes more sense to me.


Luna712:
> Please also change the call site where it is used
> 
> https://github.com/recloudstream/cloudstream/blob/d84a241026fbd16e66b575611cfedddf3f79cd93/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt#L1813

Whoops. Done, thanks!

PR #2022, opened 2025-10-18
circular resume touch fixes

rockhero1234:


app/src/main/res/layout/fragment_result.xml
```diff
@@ -603,38 +603,34 @@
...
-                                        style="@style/Widget.MaterialComponents.Button.TextButton"
                                         android:layout_width="wrap_content"
-                                        android:layout_height="match_parent"
-                                        android:gravity="center_vertical"
-                                        android:textColor="?attr/iconGrayBackground"
+                                        android:layout_height="wrap_content"
+                                        android:layout_gravity="center"
+                                        android:layout_marginStart="10dp"
+                                        android:gravity="center"
+                                        android:letterSpacing="0.09"
```
fire-light42: Why?

app/src/main/res/layout/fragment_result.xml
```diff
@@ -603,38 +603,34 @@
...
+                                        android:textColor="?attr/black"
```
fire-light42: Why? This does not match the other buttons color at all.

app/src/main/res/layout/fragment_result.xml
```diff
@@ -603,38 +603,34 @@
...
+                                        android:layout_marginStart="10dp"
                                         android:gravity="center"
-                                        android:paddingStart="5dp"
-                                        android:textColor="?attr/iconGrayBackground"
-                                        tools:text="69m remaining"
-                                        android:textStyle="bold"
-                                        tools:ignore="RtlSymmetry"
+                                        android:letterSpacing="0.09"
+                                        android:textAllCaps="false"
+                                        android:textColor="?attr/black"
```
fire-light42: Same issue.

app/src/main/res/layout/fragment_result.xml
```diff
@@ -603,38 +603,34 @@
...
+                                        android:layout_height="wrap_content"
+                                        android:layout_gravity="center"
+                                        android:layout_marginStart="10dp"
                                         android:gravity="center"
-                                        android:paddingStart="5dp"
-                                        android:textColor="?attr/iconGrayBackground"
-                                        tools:text="69m remaining"
-                                        android:textStyle="bold"
-                                        tools:ignore="RtlSymmetry"
+                                        android:letterSpacing="0.09"
```
fire-light42: Same issue.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
```diff
@@ -688,12 +688,12 @@ open class ResultFragmentPhone : FullScreenPlayer() {
 
         observeNullable(viewModel.movie) { data ->
             resultBinding?.apply {
-                resultPlayMovie.isVisible = data is Resource.Success
+                resultPlayMovie.isVisible = data is Resource.Success && (viewModel.resumeWatching.value == null)
```
fire-light42: This looks like it could be a race condition. Please do not do logic that depends on two different values.

fire-light42 (changes requested):

rockhero1234:
@fire-light42 I think previously fixes are not that good. Making clickable and focusable to false solved click issue i didn't noticed that before, and moving watching observable below movie solved play movie visibility issue.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
```diff
@@ -693,7 +639,7 @@ open class ResultFragmentPhone : FullScreenPlayer() {
                     data is Resource.Success && viewModel.currentRepo?.api?.hasDownloadSupport == true
 
                 (data as? Resource.Success)?.value?.let { (text, ep) ->
-                    resultPlayMovie.setText(text)
+//                    resultPlayMovie.setText(text)
```
fire-light42: I know that setText may change visibility here, but it is required that the text of resultPlayMovie is set as otherwise it will say "Play Movie" on both torrents and livestreams.

fire-light42: You could manually do the .toStringNull here if visibility is not what you want.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
```diff
@@ -705,6 +651,12 @@ open class ResultFragmentPhone : FullScreenPlayer() {
                         )
                         return@setOnLongClickListener true
                     }
+                    resultResumeSeriesButton.setOnLongClickListener {
```
fire-light42, line 708: I am unable to trigger this?

fire-light42: Did you mean to place this somewhere else given that no series will be a valid movie?

rockhero1234: Yeah it's only for movie

fire-light42:
> moving watching observable below movie solved play movie visibility issue.

That screams race condition to me, do not rely upon the execution order of observe. Observe may trigger in *any* order, so place `resultPlayMovie` in a dummy header view if you want to change the visibility from two different observables.

rockhero1234:
@fire-light42 did changes

PR #2023, opened 2025-10-18
feat:check icon on watched episode and auto mark watched

rockhero1234:


app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1647,6 +1656,13 @@ class GeneratorPlayer : FullScreenPlayer() {
                     )
                 }
             }
+            if(nextEp){
+                when (val meta = currentMeta) {
+                    is ResultEpisode -> {
+                        setVideoWatchState(meta.id, VideoWatchState.Watched)
```
fire-light42: What issue is this supposed to solve? setVideoWatchState is supposed to be for manual progress manipulation to override "DurPos". But if it is not set then DurPos should be used. Make 95% the same as "watched" if you want the ic_baseline_check_24 behavior instead.

rockhero1234: Got it

fire-light42: That is a good question, and yes that would make sense with our current model. So no auto set, but auto remove is fine.

rockhero1234: Ok I will set check above 95 and auto remove watched .

fire-light42 (approved):

rockhero1234:
@fire-light42 something went wrong

fire-light42:
> @fire-light42 something went wrong

What is wrong? I changed your code a bit after, but it should still work.

rockhero1234:
Showing check on unwatched episode

fire-light42:
> Showing check on unwatched episode

Willfix

PR #2036, opened 2025-10-19
Enable deprecations in SubtitleHelper

Luna712:
I'm not sure if we want to do this yet or not but I was just assuming this was supposed to be done after the latest stable. If this isn't ready to be done this can be closed.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SubSource.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SubSource.kt
@@ -29,7 +29,7 @@ class SubSourceApi : SubtitleAPI() {
 
         //Only supports Imdb Id search for now
         if (query.imdbId == null) return null
-        val queryLang = SubtitleHelper.fromTagToLanguageName(query.lang)
+        val queryLang = SubtitleHelper.fromTagToEnglishLanguageName(query.lang)
         val type = if ((query.seasonNumber ?: 0) > 0) TvType.TvSeries else TvType.Movie
 
         val searchRes = app.post(
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
@@ -159,7 +159,7 @@ object SubtitleHelper {
 
     @Deprecated(
         "Default language code changed to IETF BCP 47 tag",
-        replaceWith = ReplaceWith("fromTagToLanguageName(input)"),
+        replaceWith = ReplaceWith("fromTagToEnglishLanguageName(input)"),
         level = DeprecationLevel.WARNING
     )
     /**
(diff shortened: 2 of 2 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -154,19 +157,23 @@ object SubtitleHelper {
         return languages.getOrNull(index)
     }
 
-    // @Deprecated(
-    //     "Default language code changed to IETF BCP 47 tag",
-    //     replaceWith = ReplaceWith("fromTagToLanguageName(input)"))
+    @Deprecated(
+        "Default language code changed to IETF BCP 47 tag",
+        replaceWith = ReplaceWith("fromTagToLanguageName(input)"),
```
diogob003: Sorry, I think the correct replacement is `fromTagToEnglishLanguageName` instead of `fromTagToLanguageName` (which translates the name)
```suggestion
        replaceWith = ReplaceWith("fromTagToEnglishLanguageName(input)"),
```

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -154,19 +157,23 @@ object SubtitleHelper {
...
     fun fromTwoLettersToLanguage(input: String): String? {
         return getLanguageDataFromCode(input)?.languageName
     }
 
-    // @Deprecated(
-    //     "Default language code changed to IETF BCP 47 tag",
-    //     replaceWith = ReplaceWith("fromTagToLanguageName(input)"))
+    @Deprecated(
+        "Default language code changed to IETF BCP 47 tag",
+        replaceWith = ReplaceWith("fromTagToLanguageName(input)"),
```
diogob003:
```suggestion
        replaceWith = ReplaceWith("fromTagToEnglishLanguageName(input)"),
```

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SubSource.kt
```diff
@@ -29,7 +29,7 @@ class SubSourceApi : SubtitleAPI() {
 
         //Only supports Imdb Id search for now
         if (query.imdbId == null) return null
-        val queryLang = SubtitleHelper.fromTwoLettersToLanguage(query.lang!!)
+        val queryLang = SubtitleHelper.fromTagToLanguageName(query.lang)
```
diogob003:
```suggestion
        val queryLang = SubtitleHelper.fromTagToEnglishLanguageName(query.lang)
```

Luna712: According to doc comments, fromTagToLanguageName localizes it in the app language, is that what we want or not? I might be misunderstanding also.

diogob003: The docs are right.
We want to localize just when showing it to users. In this case we want it in English to compare with the API response (which is in English) 😉.
I think SubSource is not working anyway due to the API changes. I began to fix #1896, but the per-app limit was too low. Looks like they changed it again to a per user account limit. I think now it's worth fixing it.

fire-light42 (approved):

Issue #1351, opened 2024-09-25
Download progress and download itself not working


vdbhb59:
### Steps to reproduce

Even though a movie/episode (any video) is able to play, irrespective of the provider or source stream, the download never happens for the video file. Associated subtitles get downloaded, but not the movie itself. It was working before the last 4 commits (2 releases prior). The progress has vanished as well.

### Expected behavior

Video file should download, and the download progress should be visible as well.

### Actual behavior

Subtitles get downloaded, but not the video file itself.

### Cloudstream version and commit hash

pre-release-d899ecb & -3 pre-releases

### Android version

14 & 12

### Logcat

```java
https://pb.flossboxin.org.in/?5e0a080a6d32ca16#HKkBgoX8dUgMbgbTEhKsNKMKT7H1MkeX9kX78LDC8pK1
```

### Acknowledgements

- [X] I am sure my issue is related to the app and **NOT some extension**.
- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [X] I will fill out all of the requested information in this form.

fire-light42:
Nothing in the recent updates changed anything about downloads. The last update to the downloader was 25 days ago, and by looking at the logs it looks like it gets an invalid m3u8 file. M3u8 parsing is different for the downloader and player, so this might be the issue, and the site simply updated, but I am unable to verify this without the m3u8 file itself.

fire-light42:
By the logs it is a parsing error in m3u8. It might be that an extractor used by all sites changed/updated, and gives invalid links. However if you can just give me a short snippit of the m3u8 text I can say if this is the case or not.

vdbhb59:
> By the logs it is a parsing error in m3u8. It might be that an extractor used by all sites changed/updated, and gives invalid links. However if you can just give me a short snippit of the m3u8 text I can say if this is the case or not.

How can I give that? I am not sure what to do to provide you that.

dcb774:
So it could be fixed or it working only on website allow

Issue #1118, opened 2024-06-04
Downloads cancel in unstable network conditions 


trustosas:
### Steps to reproduce

1. Start a download
2. Turn off Internet (data/wifi)
3. Download fails (cancels)

### Expected behavior

I expected it to smartly pause the download, conserving the download progress that has been made for a later time when Internet is good.

### Actual behavior

The download fails entirely. All progress made is lost and you have to start from scratch.

### Cloudstream version and commit hash

4.3.2 b3e3dad 

### Android version

Android 11 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [X] If related to a provider, I have checked the site and it works, but not the app.
- [X] I will fill out all of the requested information in this form.

trustosas:
Can't the link be refreshed, like the way 1DM does it?

Issue #1089, opened 2024-05-13
Video Download problem 


EHMilon:
### Steps to reproduce

Video download is not working until it play. When I'm trying to download a movie by holding the download button or choosing download mirror option, it's just loading until I end. But when I play the video and all the links successfully load, then I try to download again now it's work fine. All the links shows. 

### Expected behavior

Holding the download button or choosing download mirror option should show the all download mirror links.
There is no option to select maximum download at a time. 

### Actual behavior

It's just loading until the video play and load all the links. 

### Cloudstream version and commit hash

4.3.2 ee4d1de May 9, 2024 7:48:07 PM 

### Android version

Android 12

### Logcat

```java
--------- beginning of crash
05-13 02:55:17.684 23959 23959 E AndroidRuntime: FATAL EXCEPTION: main
05-13 02:55:17.684 23959 23959 E AndroidRuntime: Process: com.lagradost.cloudstream3.prerelease, PID: 23959
05-13 02:55:17.684 23959 23959 E AndroidRuntime: android.view.InflateException: Binary XML file line #59 in com.lagradost.cloudstream3.prerelease:layout/home_result_grid_expanded: Binary XML file line #59 in com.lagradost.cloudstream3.prerelease:layout/home_result_grid_expanded: Error inflating class <unknown>
05-13 02:55:17.684 23959 23959 E AndroidRuntime: Caused by: android.view.InflateException: Binary XML file line #59 in com.lagradost.cloudstream3.prerelease:layout/home_result_grid_expanded: Error inflating class <unknown>
05-13 02:55:17.684 23959 23959 E AndroidRuntime: Caused by: java.lang.reflect.InvocationTargetException
05-13 02:55:17.684 23959 23959 E AndroidRuntime: 	at java.lang.reflect.Constructor.newInstance0(Native Method)
05-13 02:55:17.684 23959 23959 E AndroidRuntime: 	at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
05-13 02:55:17.684 23959 23959 E AndroidRuntime: 	at android.view.LayoutInflater.createView(LayoutInflater.java:858)
05-13 02:55:17.684 23959 23959 E AndroidRuntime: 	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:1010)
05-13 02:55:17.684 23959 23959 E AndroidRuntime: 	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:965)
[...]

PR #2037, opened 2025-10-19
Download rework

CranberrySoup:
Changes the whole download system to:

1. Add a user visible download queue
2. Migrate away from WorkManager (since it does not work for everyone)
3. Add instantaneous download button feedback
4. Add batch downloading

Remaining issues to fix:
- [x] Batch download button (easy)
- [x] Fix the resume download button (easy)
- [x] Add link loading notification (easy)
- [x] Good download queue UI (hard)
- [ ] Quality assurance

I plan to complete this pull request this week.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
@@ -49,6 +49,7 @@ class DownloadQueueService : Service() {
         const val DOWNLOAD_QUEUE_CHANNEL_NAME = "Download queue service"
         const val DOWNLOAD_QUEUE_CHANNEL_DESCRIPTION = "App download queue notification."
         const val DOWNLOAD_QUEUE_NOTIFICATION_ID = 917194232 // Random unique
+        @Volatile
         var isRunning = false
 
         fun getIntent(
@@ -176,7 +177,7 @@ class DownloadQueueService : Service() {
             debugWarning({ timeTaken == null || timeTaken > 3_000 }, {
                 "Abnormally long downloader startup time of: ${timeTaken ?: timeout.inWholeMilliseconds}ms"
             })
-            debugAssert({ timeTaken != null }, { "Downloader startup should not time out" })
+            debugAssert({ timeTaken == null }, { "Downloader startup should not time out" })
 
             totalDownloadFlow
                 .takeWhile { (instances, queue) ->
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueAdapter.kt
@@ -42,7 +42,7 @@ class DownloadQueueAdapter(val fragment: Fragment) : BaseAdapter<DownloadAdapter
     diffCallback = BaseDiffCallback(
         itemSame = { a, b -> a.item?.id == b.item?.id },
         contentSame = { a, b ->
-            a.item?.id == b.item?.id
+            a.item == b.item
         })
 ) {
     var currentDownloads = 0
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueFragment.kt
@@ -25,12 +25,14 @@ class DownloadQueueFragment :
 
     override fun onBindingCreated(binding: FragmentDownloadQueueBinding) {
         val adapter = DownloadQueueAdapter(this@DownloadQueueFragment)
+        val clearQueueItem = binding.downloadQueueToolbar.menu?.findItem(R.id.cancel_all)
 
         observe(queueViewModel.childCards) { cards ->
             val size = cards.queue.size + cards.currentDownloads.size
             val isEmptyQueue = size == 0
             binding.downloadQueueList.isGone = isEmptyQueue
             binding.textNoQueue.isGone = !isEmptyQueue
+            clearQueueItem?.isVisible = !isEmptyQueue
 
             adapter.submitQueue(cards)
         }
@@ -45,7 +47,7 @@ class DownloadQueueFragment :
                     }
                 }
                 setAppBarNoScrollFlagsOnTV()
-                menu?.findItem(R.id.cancel_all)?.setOnMenuItemClickListener {
+                clearQueueItem?.setOnMenuItemClickListener {
                         AlertDialog.Builder(context, R.style.AlertDialogCustom)
                             .setTitle(R.string.cancel_all)
                             .setMessage(R.string.cancel_queue_message)
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/downloader/DownloadUtils.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/downloader/DownloadUtils.kt
@@ -10,6 +10,7 @@ import coil3.request.ImageRequest
 import coil3.request.SuccessResult
 import com.lagradost.cloudstream3.R
 import com.lagradost.cloudstream3.mvvm.logError
+import com.lagradost.cloudstream3.mvvm.safe
 import com.lagradost.cloudstream3.ui.player.SubtitleData
 import com.lagradost.cloudstream3.ui.result.ExtractorSubtitleLink
 import com.lagradost.cloudstream3.utils.Coroutines.ioSafe
@@ -26,38 +27,33 @@ object DownloadUtils {
     internal fun Context.getImageBitmapFromUrl(
         url: String,
         headers: Map<String, String>? = null
-    ): Bitmap? {
-        try {
-            if (cachedBitmaps.containsKey(url)) {
-                return cachedBitmaps[url]
-            }
-
-            val imageLoader = SingletonImageLoader.get(this)
-
-            val request = ImageRequest.Builder(this)
-                .data(url)
-                .apply {
-                    headers?.forEach { (key, value) ->
-                        extras[Extras.Key<String>(key)] = value
-                    }
-                }
-                .build()
-
-            val bitmap = runBlocking {
-                val result = imageLoader.execute(request)
-                (result as? SuccessResult)?.image?.asDrawable(applicationContext.resources)
-                    ?.toBitmap()
-            }
-
-            bitmap?.let {
-                cachedBitmaps[url] = it
-            }
-
-            return bitmap
-        } catch (e: Exception) {
-            logError(e)
-            return null
+    ): Bitmap? = safe {
+        if (cachedBitmaps.containsKey(url)) {
+            return@safe cachedBitmaps[url]
         }
+
+        val imageLoader = SingletonImageLoader.get(this)
+
+        val request = ImageRequest.Builder(this)
+            .data(url)
+            .apply {
+                headers?.forEach { (key, value) ->
+                    extras[Extras.Key<String>(key)] = value
+                }
+            }
+            .build()
+
+        val bitmap = runBlocking {
+            val result = imageLoader.execute(request)
+            (result as? SuccessResult)?.image?.asDrawable(applicationContext.resources)
+                ?.toBitmap()
+        }
+
+        bitmap?.let {
+            cachedBitmaps[url] = it
+        }
+
+        return@safe bitmap
     }
 
     //calculate the time


app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
```diff
@@ -0,0 +1,158 @@
...
+                        VideoDownloadManager.currentDownloads.contains(it.downloadQueueWrapper.id)
+                            .not()
+                    }
+                // Just the queue
+                val currentVisualQueue = DownloadQueueManager.queue.size
+
+                updateNotification(context, currentVisualDownloads, currentVisualQueue)
+
+                // Arbitrary delay to prevent hogging the CPU, decrease to make the queue feel slightly more responsive
+                delay(500)
```
fire-light42: This should be a condition variable. Unfortunately the built in synchronization primatives are built for Java threading, and not the kotlin coroutines. 

The fix for this would be something like:
```kotlin
val conditionSignal = CompletableDeferred<Unit>()

// ...

select {
    onTimeout(500) {
        println("Waited")
    }
    conditionSignal.onAwait {
        println("Skipped waiting")
    }
}
```

Select simply awaits the fastest execution. So you can `conditionSignal.complete(Unit)` to skip the 500ms delay. If you implement this correctly you can do a timeout of 10s or something similar.

app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
```diff
@@ -0,0 +1,158 @@
...
+            startForeground(
+                DOWNLOAD_QUEUE_NOTIFICATION_ID,
+                baseNotification.build(),
+                FOREGROUND_SERVICE_TYPE_DATA_SYNC
+            )
+        } else {
+            startForeground(DOWNLOAD_QUEUE_NOTIFICATION_ID, baseNotification.build())
+        }
+
+        val context = this.applicationContext
```
fire-light42: `this` is a context, I do not understand why it needs to be applicationContext

app/src/main/java/com/lagradost/cloudstream3/utils/downloader/DownloadUtils.kt
```diff
@@ -0,0 +1,168 @@
...
+                return cachedBitmaps[url]
+            }
+
+            val imageLoader = SingletonImageLoader.get(this)
+
+            val request = ImageRequest.Builder(this)
+                .data(url)
+                .apply {
+                    headers?.forEach { (key, value) ->
+                        extras[Extras.Key<String>(key)] = value
```
fire-light42: headerBuilder["User-Agent"] = USER_AGENT

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadFragment.kt
```diff
@@ -225,6 +226,10 @@ class DownloadFragment : BaseFragment<FragmentDownloadsBinding>(
                 setOnClickListener { showStreamInputDialog(it.context) }
             }
 
+            downloadQueueButton.setOnClickListener {
+                findNavController().navigate(R.id.action_navigation_global_to_navigation_download_queue)
```
Luna712: Why not use `activity?.navigate` like elsewhere in this class?

app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueFragment.kt
```diff
@@ -0,0 +1,60 @@
...
+            adapter.submitList(cards)
+        }
+
+        binding.apply {
+            downloadQueueToolbar.apply {
+                title = txt(R.string.title_queue).asString(context)
+                if (isLayout(PHONE or EMULATOR)) {
+                    setNavigationIcon(R.drawable.ic_baseline_arrow_back_24)
+                    setNavigationOnClickListener {
+                        activity?.onBackPressedDispatcher?.onBackPressed()
```
Luna712: I wonder if this one should use `dispatchBackPressed()` from BaseFragment here?

app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueAdapter.kt
```diff
@@ -0,0 +1,292 @@
...
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.core.content.ContextCompat
+import androidx.core.view.isGone
+import androidx.core.view.isVisible
+import androidx.core.widget.doOnTextChanged
+import androidx.fragment.app.Fragment
+import androidx.recyclerview.widget.ItemTouchHelper
+import androidx.recyclerview.widget.RecyclerView
+import com.lagradost.cloudstream3.AcraApplication.Companion.getKey
```
Luna712: Can use CloudStreamApp (it's new and drop in replacement) instead. Only doing one comment about it but there are other places in this PR in other classes that could also.

CranberrySoup:
The queue should now work regardless of how you try to interrupt it. Please test it @fire-light42 @Luna712

Luna712 (commented):
Screenshot_20251206_140424_CloudStream Debug

Haven't extensively tested much yet, but did find a UI bug with the download button in results.

CranberrySoup:
Thank you. 
I did a lot of testing but only in series. I forgot the movies.

CranberrySoup:
Fixed :+1:

Luna712:
Here is a bit more comprehensive review with more issues I've found. Though still not done testing completely.
1. Managed to hit a crash:
Screenshot_20251206_182203_CloudStream Debug

2. Cancel all queued downloads doesn't cancel in progress or paused downloads (not sure if intentional or not)
3. Not sure if this is intentional either, but 1) you can't drag one from below the separator to above the separated to start that download immediately and requeue one above it. Additionally if you pause the 3 parallel downloads others don't start while those are paused. Though this may be intentional behavior.
4. When I click the delete file on an in progress downloads while in the queue, it doesn't remove them from the adapter (or more likely view model because even if I close and reopen the fragment they still remain there).
5. A bit inconsistent/odd padding for the download all button: 
Screenshot_20251206_182838_CloudStream Debug

Perhaps the UI could be simplified by condensing into a download all icon? Not 100% sure though.

6. When I click download all, the icons immediately change, but clicking them doesn't give an option to cancel them until it finishes finding the links for them. Also maybe a cancel all button should be added if all currently queued to reverse that download all action?
7.  Some animations randomly stop spinning. Not sure if that is just due to recyclerviews though: <video src="https://github.com/user-attachments/assets/a69befce-16c5-46a1-ba92-c49839059b21" />

CranberrySoup:
> 1. Managed to hit a crash

It is such a weird issue because the mutex is surrounded by a try catch. I will try to fix it.

> 2. Cancel all queued downloads doesn't cancel in progress or paused downloads (not sure if intentional or not)

This is intentional and I tried to clarify that using the wording in the popup. The queued downloads are instantaneous and easy to readd, making them of lower importance. While active downloads, which require link loading, are presumed to be more important, requiring explicit intent to cancel or pause. 

> 3. Not sure if this is intentional either, but 1) you can't drag one from below the separator to above the separated to start that download immediately and requeue one above it. Additionally if you pause the 3 parallel downloads others don't start while those are paused. Though this may be intentional behavior.

All of this behavior is intentional, although some of it may change in a future update. While it would be nice to move the separator or active downloads it would complicate the system, pull request and UX too much. Future pull requests may try to improve this, this pull request is big enough already.

> 4. When I click the delete file on an in progress downloads while in the queue, it doesn't remove them from the adapter (or more likely view model because even if I close and reopen the fragment they still remain there).

I thought I fixed that issue, but I suppose not. Thank you for reporting it.

> 5. A bit inconsistent/odd padding for the download all button. Perhaps the UI could be simplified by condensing into a download all icon? Not 100% sure though.

Will fix :+1: 
Netflix designs the button with an additional line in the download icon, maybe that would be nice to have. 

> 6. When I click download all, the icons immediately change, but clicking them doesn't give an option to cancel them until it finishes finding the links for them. Also maybe a cancel all button should be added if all currently queued to reverse that download all action?

I will add a click to cancel popup to each. However, I do not want to add more buttons to the result fragment. It already lacks space. A cancel all button in the queue menu is enough in my opinion.

> Some animations randomly stop spinning. Not sure if that is just due to recyclerviews though. 

I think the state is not properly set. I will try to fix.

Thank you for your review :heart:

Luna712:
> This is intentional and I tried to clarify that using the wording in the popup. The queued downloads are instantaneous and easy to readd, making them of lower importance. While active downloads, which require link loading, are presumed to be more important, requiring explicit intent to cancel or pause.

That makes sense and yeah I wasn't sure.

> All of this behavior is intentional, although some of it may change in a future update. While it would be nice to move the separator or active downloads it would complicate the system, pull request and UX too much. Future pull requests may try to improve this, this pull request is big enough already.

Same as above

> I will add a click to cancel popup to each. However, I do not want to add more buttons to the result fragment. It already lacks space. A cancel all button in the queue menu is enough in my opinion.

I wasn't necessarily thinking another button but rather replace the download all with cancel all if all in progress or delete all if all downloaded in that available batch, but the other button works too. Was just an idea, but isn't actually necessary for this PR. I think the first part of this I mentioned where you can't cancel at all until it finishes finding links is probably a bug though that could be fixed but not certain if that was intentional as well or not.

> Thank you for your review ❤️

Of course, this PR really works amazing. Downloads no longer lag the entire app when in progress and the UI especially queue and instant download feedback is amazing.

Luna712:
Just for the record to keep track of some issues this should at least in theory improve or fix:

* #372 (fixes)
* #399 (fixes)
* #621 (fixes)
* #1089 (I think this fixes this one, though kinda confused what that one is asking)
* #1118 (at least improves not sure if it resolves completely)
* #1166 (fixes)
* #1351 (at least improves not sure if it resolves completely)
* #2182 (at least improves not sure if it resolves completely)
* #2224 (at least improves not sure if it resolves completely)
* #2268 (fixes)
* #2269 (fixes)

Potentially more also as I didn't go through every one.

CranberrySoup:
Changes:

- Replaced the download all button with an icon.
- Added a text to view the queue size at a glance from the download fragment
- Likely fixed the crash
- Likely fixed the issue of deleted downloads remaining in the queue
- Added redundant key removal (will make the logcat much cleaner)

I was not able to fix the animations by changing the button code, even when I changed setStatusInternal to always start the animation and never cancel the animation it still stopped. I suppose we must use ItemAnimator to manage the animations.

app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+    init {
+        viewModelScope.launch {
+            totalDownloadFlow.collect { queue ->
+                updateChildList(queue)
+            }
+        }
+    }
+
+    fun updateChildList(downloads: DownloadAdapterQueue) =
+        viewModelScope.launchSafe {
```
fire-light42: Why does this need to be launched?

app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
```diff
@@ -0,0 +1,185 @@
...
+                        val newDownloads = minOf(
+                            // Cannot exceed the max downloads
+                            maxOf(0, maxDownloads - currentInstanceCount),
+                            // Cannot start more downloads than the queue size
+                            queue.size
+                        )
+
+                        // Cant start multiple downloads at once. If this is rerun it may start too many downloads.
+                        if (newDownloads > 0) {
+                            val downloadInstance = DownloadQueueManager.popQueue(context)
```
fire-light42, line 210: This looks like an impure function being called from a updateAndGet? You do know that updateAndGet may run this code many times, and will only use the latest value if there is no contention. So your lambda must be 100% pure.

fire-light42 (changes requested):
Very cool pull request, the code looks very clean but I have not yet tested it.

app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueViewModel.kt
CranberrySoup: Old code, fixed now.

app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
CranberrySoup: Fixed

CranberrySoup:
Rebased onto master and fixed various small issues, including the animations. 
Please test and review @fire-light42 @Luna712

Luna712:
One major issue I found with this is that removing keys should probably be done using batch like how search history and backups are done. When I did this and restored from backup of pre I have it logged in logcat about **18,000** times removing keys,  which froze the app and it never finished leading to a crash (app not responding) loop.

CranberrySoup:
> One major issue I found with this is that removing keys should probably be done using batch like how search history and backups are done.

Excellent feedback! I pushed a fix which should resolve this issue, please test to see if it works.

Luna712:
Awesome thanks, I'll do another in depth test on this tomorrow.

Luna712:
The key deletion issue seems fixed. But maybe add the backup keys to `nonTransferableKeys`? Backup and restore and backup immediately again caused backup file to grow a bit.

The issues/suggestions I've found so far now is:

Major issue is that backup while there is in progress/queued downloads, then clear app data and restore tries to re-download. While it sometimes works, it won't always work due to storage permissions and should be excluded from backup. It seems to only happen with the 3 in progress downloads but I'm not certain if there is some case it could try all queued also.

While links are loading there is absolutely no way to cancel a download. Not from the queue and not from the series page. Long clicking the loading animation download icon just opens sources again (probably fine for that one) and in the queue clicking it does nothing. While this was probably the case before, the instant feedback makes it more obvious and not clear that you cant cancel it. Probably allowing to  cancel a download in the queue before link loading is finished would be good.

When you delete an in progress download, it does now remove from the adapter but it takes about 15 seconds sometimes but other times it was very quick so may not be an issue and just my device.

Perhaps there should be some UI indicator for no downloads in queue other then just a blank page? Not sure.

I'm not certain if it does, but perhaps preload links for next item in queue when its close to being active (like when current active are almost done) just a suggestion perhaps a later follow up.

Maybe not for this PR not sure, but better handling for failed downloads would probably be good. I tested this by batch downloading something with 26 very short and small episodes, then put my phone in airplane mode. The UI flickers for a few seconds while all queued downloads are instantly failed and removed from the queue while the 3 that were actually running are also removed from the queue, but still appear as half downloaded in the actual child list. Also when the downloads failed all downloads in queue created 0 byte files that remained on the device.

Again, probably not for this PR, but downloads should probably honor the source priority system or something because that is made worse with batch downloads, whereas it ends up downloading like 4K first which takes like 30GB sometimes.

Maybe this (and the subtitle if it exists) should auto scroll to eventually show the full title? Also when there is no subtitle the title doesnt quite align with the icon.
Screenshot_20251212_121644_CloudStream Debug

CranberrySoup:
> The key deletion issue seems fixed. But maybe add the backup keys to nonTransferableKeys? Backup and restore and backup immediately again caused backup file to grow a bit.

Good idea. I will do that :+1: 

> Major issue is that backup while there is in progress/queued downloads, then clear app data and restore tries to re-download. While it sometimes works, it won't always work due to storage permissions and should be excluded from backup. It seems to only happen with the 3 in progress downloads but I'm not certain if there is some case it could try all queued also.

Downloads should not start from backups. I will also add relevant keys to `nonTransferableKeys` to ensure it does not happen.

>  While links are loading there is absolutely no way to cancel a download. Not from the queue and not from the series page. Long clicking the loading animation download icon just opens sources again (probably fine for that one) and in the queue clicking it does nothing. While this was probably the case before, the instant feedback makes it more obvious and not clear that you cant cancel it. Probably allowing to cancel a download in the queue before link loading is finished would be good.

I initially considered this out of scope for the pull request. However, it would be easy to fix. I will add it. 

> When you delete an in progress download, it does now remove from the adapter but it takes about 15 seconds sometimes but other times it was very quick so may not be an issue and just my device.

The queue directly reflects the internal state of the download manager. If the download stays for 15 seconds it may reflect a mistake in the download management. Debugging this may fix some hidden issue. However, I have not managed to recreate the issue. Are you able to recreate it?

> Perhaps there should be some UI indicator for no downloads in queue other then just a blank page? Not sure.

Good idea. I will add it.

> I'm not certain if it does, but perhaps preload links for next item in queue when its close to being active (like when current active are almost done) just a suggestion perhaps a later follow up.

An excellent idea, but better to have as a separate pull request after this one. Any additional features or changes to the internals may create new bugs which would be hard to catch in an already huge pull request. I also want to prevent scope creep to ensure this pull request gets merged at all.

> Maybe not for this PR not sure, but better handling for failed downloads would probably be good. I tested this by batch downloading something with 26 very short and small episodes, then put my phone in airplane mode. The UI flickers for a few seconds while all queued downloads are instantly failed and removed from the queue while the 3 that were actually running are also removed from the queue, but still appear as half downloaded in the actual child list. Also when the downloads failed all downloads in queue created 0 byte files that remained on the device.

Same reasoning as above. Good to have as a separate pull request.

> Again, probably not for this PR, but downloads should probably honor the source priority system or something because that is made worse with batch downloads, whereas it ends up downloading like 4K first which takes like 30GB sometimes.

This is something I want myself. However, I want to keep the core download logic as similar as possible to before. It would be one of my first pull requests after this one.

> Maybe this (and the subtitle if it exists) should auto scroll to eventually show the full title? 

I had that before but it was too laggy to keep. Scrolling with 20 animated texts was a terrible experience. 

> Also when there is no subtitle the title doesnt quite align with the icon.

Will fix.



Thank you a lot for another well thought out review :smile:

Luna712:
> The queue directly reflects the internal state of the download manager. If the download stays for 15 seconds it may reflect a mistake in the download management. Debugging this may fix some hidden issue. However, I have not managed to recreate the issue. Are you able to recreate it?

I can only reproduce it if I immediately cancel almost as soon as it starts and when there are 3 in progress at once, and I try to cancel two or all three in quick succession. It is also much more obvious if I delete all 3 at once by deleting the entire series using multi-delete, while 3 are in progress then go back into the queue and they remain there for 15 or so seconds longer.

> I had that before but it was too laggy to keep. Scrolling with 20 animated texts was a terrible experience.

Oh yeah that makes sense in that case totally not worth it.

Luna712:
I found another bug as well. I enabled 10 parallel downloads and the app eventually crashed with an OOM (expected to be honest) but the issue is when it restarted into safe mode and downloads resumed, some downloads that were previously active are now queued, the queue order is changed to some random order and only a few downloads start simultaneously while others in queue are paused and those above the parallel downloads won't download and does like one at time. It's honestly hard to explain the behavior but it acts quite odd. But this is also such an edge case, and I think it always happened it just wasn't visible before so isn't necessarily something to fix with this PR.

The final issue I've found is when downloads are completed they aren't *always* removed from the queue sometimes they stay in the queue while the icon is the completed icon. There is a potential this ties into the above issue as well, since I noticed this issue at the same time as well.

**EDIT** For that second issue its like its supposed to be downloading but isn't, it doesn't appear at all in main downloads, the queue says "8/8", and clicking the download completed icon does nothing. Clear queue does nothing either. They are still there after restarting the app also. The notification also still says "8 active downloads". There seems to be no way to remove them at all, making that potential issue a larger issue then I first thought.

**EDIT 2** This also seems to be the actual reason the first issue happens. Because there was these 8, they take up the 8 threads somehow so only 2 other downloads were actually working in parallel.

**EDIT 3** The episodes did fully download, the full files exist, but it's like they weren't moved from the queue to completed states when they should have.

---
On another note, amazing work with this PR downloads work so smoothly without lagging the entire app now also, before the OOM there was zero lag with 10 ongoing downloads, before it barely worked with 2 without lagging much.

CranberrySoup:
Fixes:

- No automatic queue in safe mode
- More reliable queue management
- Now possible to cancel loading links
- Empty queue message
- Centered text for queue items when no subtitle
- No download data in backups

Luna712:
New bugs I've found now:

* Download something, then while links are still load go to queue and cancel it, then return to the results fragment for the download you canceled, now a download failed and clicking it still shows cancel option. Additionally, I tried this a second time and when canceling in queue for an episode, it showed the failed icon and remained in the queue. Queue remained showing 1/1, and exiting and reentering does not clear it.

Screenshot_20251216_124447_CloudStream Debug
Screenshot_20251216_124425_CloudStream Debug
Screenshot_20251216_124354_CloudStream Debug



* The safe mode bug is mostly fixed, however when I triggered it again, then exited safe mode by restarting apps, the downloads that were running already were automatically resumed but ever other queued download was removed from queue and only active ones resumed again. Additionally while in safe mode the downloads notifications just appear frozen, downloads appear paused but the notifications remain as if they are frozen rather than any indication of being paused or anything and the cancel button doesn't work.
* Some notifications seem to actually get stuck (might be related to the safe mode bug) I have multiple download episodes, that is still a notification but no longer in the queue at all. Most sitting at 99%, but a couple at like 40 or 50% even though rhey are complete, in some cases or perhaps canceled in other cases? A couple of them ended up in main downloads with just a download icon:
Screenshot_20251216_123917_CloudStream Debug
Screenshot_20251216_124009_One UI Home
Screenshot_20251216_124027_One UI Home

----

**EDIT** one more thing, subtitles are not deleted when download is canceled after it actually started, which while an existing issue also from what I can understand from the code this does have some logic to delete subtitles when canceled also, so thought I'd mention it.

CranberrySoup:
This is not abandoned. I will ensure it gets possible to merge soon.

CranberrySoup:
Additions:
- It should now be completely safe to crash (good luck trying to make it fail unsafely)
- Cancelling links does not register as a failure
- TV layout is now usable
- Subtitles now get deleted if cancelled at any time

fire-light42:
Please test @Luna712 
I am going to test too. I want to get this merged soon!

Luna712:
Will test again tomorrow

app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
```diff
@@ -0,0 +1,261 @@
...
+                    while (!(PluginManager.loadedOnlinePlugins && PluginManager.loadedLocalPlugins)) {
+                        delay(100.milliseconds)
+                    }
+                }
+            }
+
+            debugWarning({ timeTaken == null || timeTaken > 3_000 }, {
+                "Abnormally long downloader startup time of: ${timeTaken ?: timeout.inWholeMilliseconds}ms"
+            })
+            debugAssert({ timeTaken != null }, { "Downloader startup should not time out" })
```
fire-light42: Unfortunately this is wrong, as the assert is a bit counterintuitive and is the inverse of what you expect.

fire-light42: We should update the API to be correct.

app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueAdapter.kt
```diff
@@ -0,0 +1,274 @@
...
+class DownloadAdapterItem(val item: DownloadQueueWrapper?) {
+    val isSeparator = item == null
+}
+
+
+class DownloadQueueAdapter(val fragment: Fragment) : BaseAdapter<DownloadAdapterItem, Unit>(
+    diffCallback = BaseDiffCallback(
+        itemSame = { a, b -> a.item?.id == b.item?.id },
+        contentSame = { a, b ->
+            a.item?.id == b.item?.id
```
fire-light42: This is is not entirely correct, it should be `a.item == b.item` to avoid any sort of view invalidation.

app/src/main/java/com/lagradost/cloudstream3/ui/download/queue/DownloadQueueFragment.kt
```diff
@@ -0,0 +1,77 @@
...
+                    setNavigationIcon(R.drawable.ic_baseline_arrow_back_24)
+                    setNavigationOnClickListener {
+                        dispatchBackPressed()
+                    }
+                }
+                setAppBarNoScrollFlagsOnTV()
+                menu?.findItem(R.id.cancel_all)?.setOnMenuItemClickListener {
+                        AlertDialog.Builder(context, R.style.AlertDialogCustom)
+                            .setTitle(R.string.cancel_all)
+                            .setMessage(R.string.cancel_queue_message)
```
fire-light42, line 53: If you have no items in the queue this cancel message is a bit confusing. It should contain how many get will removed from the queue, or simply set the visibility = queue > 0.

app/src/main/java/com/lagradost/cloudstream3/utils/downloader/DownloadUtils.kt
```diff
@@ -0,0 +1,168 @@
...
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.runBlocking
+
+/** Separate object with helper functions for the downloader */
+object DownloadUtils {
+    private val cachedBitmaps = hashMapOf<String, Bitmap>()
+    internal fun Context.getImageBitmapFromUrl(
+        url: String,
+        headers: Map<String, String>? = null
+    ): Bitmap? {
```
fire-light42: = safe {

app/src/main/java/com/lagradost/cloudstream3/services/DownloadQueueService.kt
```diff
@@ -0,0 +1,261 @@
...
+import kotlin.time.Duration.Companion.seconds
+
+class DownloadQueueService : Service() {
+    companion object {
+        const val TAG = "DownloadQueueService"
+        const val DOWNLOAD_QUEUE_CHANNEL_ID = "cloudstream3.download.queue"
+        const val DOWNLOAD_QUEUE_CHANNEL_NAME = "Download queue service"
+        const val DOWNLOAD_QUEUE_CHANNEL_DESCRIPTION = "App download queue notification."
+        const val DOWNLOAD_QUEUE_NOTIFICATION_ID = 917194232 // Random unique
+        var isRunning = false
```
fire-light42, line 53: `@Volatile`

voidforcex9:
Screenshot_2026-02-11-12-04-27-71_bba4627ecec1346e77f282610a5b88e8

Can you please add the feature to choose quality in which we want to batch download by long press the batch download icon. It automatically makes the highest quality link in queue.

CranberrySoup:
> Can you please add the feature to choose quality in which we want to batch download by long press the batch download icon. It automatically makes the highest quality link in queue.

That is a separate issue. Once this is merged I will improve the link selection for normal downloads and batch downloads.

PR #2043, opened 2025-10-20
Upgrade to targetSdk 36

Luna712:
Now with #2002 done, this should be possible to do.

This removed `android:enableOnBackInvokedCallback` because it's enabled by default when targeting Android 16 now, so we don't need to have it enabled explicitly anymore.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -1409,7 +1409,8 @@ class CS3IPlayer : IPlayer {
                     }
                 }
 
-                //fixme: Use onPlaybackStateChanged(int) and onPlayWhenReadyChanged(boolean, int) instead.
+                // fixme: Use onPlaybackStateChanged(int) and onPlayWhenReadyChanged(boolean, int) instead.
+                @Suppress("OVERRIDE_DEPRECATION")
                 override fun onPlayerStateChanged(playWhenReady: Boolean, playbackState: Int) {
                     exoPlayer?.let { exo ->
                         event(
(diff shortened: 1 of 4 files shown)


Luna712:
@fire-light42  Just an FYI, the largest behavioral change this makes is removing the ability to lock orientation on screen sizes with widths >= 600dp, I tested this on my phone by changing minimum width in developer settings to 600dp, and it does indeed no longer lock orientation in videos, there is an opt out for this behavior on 36 only but mentions it goes away in 37 so I decided not to opt out here because there is no alternative, either we do it now or we eventually do it when we target 37, there is no way around it unfortunately. If we want me to opt out, I can but I didn't see the usefulness of doing it because we will have to eventually anyway (unless we change app category to a game as those are the only thing exempt lol)

Luna712:
> @fire-light42 Just an FYI, the largest behavioral change this makes is removing the ability to lock orientation on screen sizes with widths >= 600dp, I tested this on my phone by changing minimum width in developer settings to 600dp, and it does indeed no longer lock orientation in videos, there is an opt out for this behavior on 36 only but mentions it goes away in 37 so I decided not to opt out here because there is no alternative, either we do it now or we eventually do it when we target 37, there is no way around it unfortunately. If we want me to opt out, I can but I didn't see the usefulness of doing it because we will have to eventually anyway (unless we change app category to a game as those are the only thing exempt lol)

Okay so actually I just tested on pre. Orientation lock still doesn't work but its behavior is much odder, targeting 36 actually improved this behavior.

Luna712:
@fire-light42 I think this could probably move forward? Edge to Edge has been there for a couple weeks now with not really any reported issues, so we could probably move forward with this for those that use pre release and work out any potential issues (though hopefully there isn't any). If you don't want to move forward with it yet though that's fine too, just thought I'd bring it up.

PR #2062, opened 2025-10-24
Unify clear search icons logic

Luna712:
Reduces code duplication, and also adds support for them to show on repo and extensions search.

### DIFF
--- a/app/src/main/res/drawable/ic_baseline_close_24.xml
+++ b/app/src/main/res/drawable/ic_baseline_close_24.xml
@@ -1,4 +1,4 @@
-<vector android:height="24dp" android:tint="#FFFFFF"
+<vector android:height="24dp" android:tint="?attr/white"
     android:viewportHeight="24" android:viewportWidth="24"
     android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
     <path android:fillColor="@android:color/white" android:pathData="M19,6.41L17.59,5 12,10.59 6.41,5 5,6.41 10.59,12 5,17.59 6.41,19 12,13.41 17.59,19 19,17.59 13.41,12z"/>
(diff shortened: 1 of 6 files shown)


app/src/main/res/drawable/ic_baseline_close_24.xml
```diff
@@ -1,4 +1,4 @@
-<vector android:height="24dp" android:tint="#FFFFFF"
+<vector android:height="24dp" android:tint="?attr/white"
```
fire-light42, line 1: This is different from ?attr/textColor, is this intended?

Luna712: Yes, because the current android:tint is `#FFFFFF` (white), so to just keep that and switch between black and white I decided to just use `?attr/white` here. If you want me to use textColor still I can though.

fire-light42: The question is because the old behavior was to set the color programmatically to textcolor (and not #fff)
```kotlin
activity?.theme?.resolveAttribute(R.attr.textColor, searchExitIconColor, true)
searchExitIcon?.setColorFilter(searchExitIconColor.data)
```

I am only asking because it is an behavioral change without any reasoning behind it, not because it is wrong.

Luna712: I did forget to mention the change, but it was intentional, just because when I originally added this, it was one of my first PRs and I totally did it wrong, I apparently didn't know I could do it in the drawable, so it was just intended to make it work on dark and light themes which changing just to the white attr here keeps the original intended behavior if the drawable while maintaining that dark and light theme compatibility.

fire-light42: ?attr/textColor and ?attr/white is not the same color? I am only asking about the motivation behind the color change.

Luna712: I know they aren't. I can use textColor here if you'd prefer. The reason I used white is even though it was using textColor programmatically the drawable was using white there which seemed to be the intended behavior of the drawable. It only used textColor because I originally added it like that to maintain dark and light themes, attr/white does that too except keeps the original drawable intended behavior.

fire-light42: Understood, thanks for the response. ❤️

Issue #2045, opened 2025-10-21
Focus issues on homepage on TV


Luna712:
### Steps to reproduce

1. Have a home page with the preview scroll header
2. Scroll down a bit
3. Scroll back up trying to get to the API selection

### Expected behavior

Scrolling beyond the home page preview section, should go up to API provider and the other buttons there.

### Actual behavior

It gets stuck on the preview header until you try and scroll to the side then back and up then it works to reach it.

### Cloudstream version and commit hash

4.6.0 4270f15

### Android version

Android 14

### Logcat

```java

```

### Acknowledgements

- [x] I am sure my issue is related to the app and **NOT some extension**.
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [x] I will fill out all of the requested information in this form.

fire-light42:
Unable to replicate.

Luna712:
> Unable to replicate.

@fire-light42 

So I did some more debugging, this happens like if your a few rows below the large header, while holding the button down still (so fast scroll) then the up button starts scrolling the header items horizontally until you let go of the key for a couple seconds then go down and back up slowly.

ByAyzen:
Yeah it still happens.

VickyGittub:
Same issue. After going down, can't directly go back up to provider selection. If I keep the up button held, it will start scrolling on the horizontal trending movies/shows part. Then I have to stop and scroll down, then go up 1-by-1 click to reach the provider button.

Similar navigation issue on search page. If I scroll down to a search result, go inside it and then come back. it focuses on the search text box or type of content (movie/series/anime) and if I click on DOWN, it goes to the menu icons (HOME, SEARCH, FAV...) and now there is no way to focus on the search results again. Sometimes clicking RIGHT 2-3 times helps focus on the search results again.

PR #2063, opened 2025-10-25
Fix: homepage preview scroll stuck issue (#2045)

2400032210:
Fixed the issue where the homepage preview scroll section on TV would get stuck and not scroll to the API provider section.

Changes

Adjusted scroll logic in HomeFragment.kt

Added smooth scroll handler for top navigation

Testing

Tested on Android 14, Cloudstream 4.6.0 (commit 4270f15).
Scrolling now works smoothly.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
@@ -699,33 +699,28 @@ class HomeFragment : Fragment() {
                             }
                         }
                     } else {
-                        // Header scrolling is only relevant to TV/Emulator
+    val scrollParent = binding?.homeApiHolder
+    val firstVisibleView = recyclerView.findViewHolderForAdapterPosition(0)?.itemView
 
-                        val view = recyclerView.findViewHolderForAdapterPosition(0)?.itemView
-                        val scrollParent = binding?.homeApiHolder
+    if (firstVisibleView == null) {
+        // Scrolled past header – hide the API provider row
+        scrollParent?.isVisible = false
+    } else {
+        // Smoothly re-show API provider row when scrolled back up
+        val rect = IntArray(2)
+        firstVisibleView.getLocationInWindow(rect)
 
-                        if (view == null) {
-                            // The first view is not visible, so we can assume we have scrolled past it
-                            scrollParent?.isVisible = false
-                        } else {
-                            // A bit weird, but this is a major limitation we are working around here
-                            // 1. We cant have a real parent to the recyclerview as android cant layout that without lagging
-                            // 2. We cant put the view in the recyclerview, as it should always be shown
-                            // 3. We cant mirror the view in the recyclerview as then it causes focus issues when swaping out the mirror view
-                            //
-                            // This means that if we want to have a parent view to the recyclerview we are out of luck
-                            // Instead this uses getLocationInWindow to calculate how much the view should be scrolled
-                            // as recyclerView has no scrollY (always 0)
-                            //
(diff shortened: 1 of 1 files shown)


2400032210:
i have changed code in else block
Screenshot 2025-10-25 215212

fire-light42:
> i have changed code in else block Screenshot 2025-10-25 215212

Yes, I see that. But the "fix" does not make any sense.

fire-light42:
Can you please provide real motivation for how and why this fixes anything.

2400032210:
Hi @fire-light42, thanks for reviewing!

The issue occurs only on Android TV layout, where the homeApiHolder (top API provider section) sometimes gets stuck and doesn’t reappear after scrolling the homepage preview.

Root cause:
The RecyclerView wasn’t restoring the visibility or translation of the homeApiHolder when focus or scroll returned upward. As a result, after the header scrolled off-screen once, it stayed hidden.

Fix:
In the else block of the scroll listener, I added logic to:

Detect the first visible item (findViewHolderForAdapterPosition(0)).

If the first item is visible → gradually reset translationY of homeApiHolder to bring it back into view.

If the first item is gone → hide the holder to avoid empty space.

Ensure translationY never goes positive (prevents it from getting stuck below visible range).

Result:
The homepage preview section now scrolls smoothly, and the provider section correctly hides/shows during scroll. Tested on Android TV (Cloudstream 4.6.0, Android 14).

In my written there comments and more spaces So thats why it showing +21 -26

fire-light42:
> Ensure translationY never goes positive (prevents it from getting stuck below visible range).

How exactly does this occur? It is not possible to scroll so the translationY becomes so positive it gets hidden.

fire-light42:
Please add a `Log.i("HomeScroll", "translationY = $translationY")` and to confirm this is really the issue.

fire-light42:
@Luna712 Can you please test this pull request, because I am unable to see exactly how this would fix our current issue.

Luna712:
> @Luna712 Can you please test this pull request, because I am unable to see exactly how this would fix our current issue.

It looks like the only change this does different then current is `if (translationY > 0f) translationY = 0f` which I guess in theory could fix it, no? (I haven't tested yet and can't right now but can in a bit). However, regarding this PR since thats the only change wouldn't it be best to keep everything else the same and just change that one line to `translationY = (rect[1].toFloat() - 60.toPx).coerceAtMost(0f)`? If that even affects it that is... I'm assuming the idea is to only let it affect going down and not up which is why I said in theory it could affect it but I haven't confirmed and I'm not certain either.

fire-light42:
> > @Luna712 Can you please test this pull request, because I am unable to see exactly how this would fix our current issue.
> 
> It looks like the only change this does different then current is `if (translationY > 0f) translationY = 0f` which I guess in theory could fix it, no? (I haven't tested yet and can't right now but can in a bit). However, regarding this PR since thats the only change wouldn't it be best to keep everything else the same and just change that one line to `translationY = (rect[1].toFloat() - 60.toPx).coerceAtMost(0f)`? If that even affects it that is... I'm assuming the idea is to only let it affect going down and not up which is why I said in theory it could affect it but I haven't confirmed and I'm not certain either.

Ye, but the problem is that the logic does not make any sense as the translation will never be positive. When you scroll down it becomes negative, and when you scroll up it caps at like 0.0f or 1.0f as you cant scroll up more than there exits content.

Luna712:
> Ye, but the problem is that the logic does not make any sense as the translation will never be positive. When you scroll down it becomes negative, and when you scroll up it caps at like 0.0f or 1.0f as you cant scroll up more than there exits content.

Fair, I'll test later when I can and confirm.

PR #2068, opened 2025-10-25
Bump version to 4.6.1

Luna712:
Since we've had a major change with edge to edge, bumping the version could help know if it's related if an issue is created.


fire-light42:
Makes perfect sense. Good call.

PR #2078, opened 2025-10-26
Upgrade to minSdk 23

Luna712:
Most major packages, including all androidx packages are now being upgraded to minSdk 23. This means we also need to upgrade now otherwise we will be stuck unable to continue upgrading at all. Additionally, there really isnt a need to keep supporting API 21 and API 22 (Both are only Android 5).

I understand there is hesitation, this is not immediately urgent anyway, but I am just putting out there. For the ease of confirming that is actually necessary, I have attached references here for releases and other things about the change. So far only `work` has a stable release with the change, however others will follow soon. This is also not an exhaustive list.

References:
* https://issuetracker.google.com/issues/380448311
* https://developer.android.com/jetpack/androidx/versions#version-table
* https://developer.android.com/jetpack/androidx/releases/activity#1.12.0-alpha06
* https://developer.android.com/jetpack/androidx/releases/lifecycle#2.10.0-alpha02
* https://developer.android.com/jetpack/androidx/releases/media3#1.9.0-alpha01
* https://developer.android.com/jetpack/androidx/releases/work#2.11.0
* https://developer.android.com/jetpack/androidx/releases/biometric#1.4.0-alpha05

I have also only bumped the minSdk and set `useLegacyPackaging` here, and not removed any backwards compatible code for pre 23, that is because it will be done in a follow-up after this is determined whether we will do it and that doesn't any issues.

`useLegacyPackaging` was set because starting when targeting minSdk 23, it doesn't compress things by default causing the total base APK size to nearly double. The change was to improve app loading speed because when an app is loaded, it decompresses, but the trade was larger app size in return for faster start time. To maintain the same behavior I set that here. If we ever decide we want to utilize the new behavior we can remove it.

In my personal opinion (perhaps a bit less popular of an opinion), CloudStream3 should follow to use the same minSdk as androidx, if another package doesn't support it, we just don't upgrade unless we absolutely have to then we can figure out an alternative, but androidx packages (in particular media3 comes to mind which has already bumped to minSdk 23) often have very good bug fixes and are also needed for future targetSdk, which we can't just simply never bump.

I'm mostly creating this just to document a lot of this, its still up to whether or not this is merged and we go that route. But it has been asked in the past for a very good reason to bump minSdk and I believe being required to for core dependent packages is a fairly good reason to do so, which is why I am creating with documentation and references, etc...

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -181,9 +181,8 @@ dependencies {
     implementation(libs.core.ktx)
     implementation(libs.activity.ktx)
     implementation(libs.appcompat)
-    implementation(libs.bundles.navigationKtx)
-    implementation(libs.lifecycle.livedata.ktx)
-    implementation(libs.lifecycle.viewmodel.ktx)
+    implementation(libs.bundles.lifecycle)
+    implementation(libs.bundles.navigation)
 
     // Design & UI
     implementation(libs.preference.ktx)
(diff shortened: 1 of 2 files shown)


Luna712:
NOTE: This is not necessarily meant to be merged unless you decide to, but more of a proof of concept.

Luna712:
Just to note, lifecycle and activity are now stable and require minSdk 23 as well. savedstate which we don't define explicitly but is included implicitly in something else seems to as well but doesn't show it does in release notes.
```
Manifest merger failed : uses-sdk:minSdkVersion 21 cannot be smaller than version 23 declared in library [androidx.savedstate:savedstate-ktx:1.4.0] 
```

Luna712:
@fire-light42 just FYI, media3 1.9.0 stable is *scheduled* for release tomorrow, which fixes many bugs also, but requires minSdk 23 so I think it might be time to consider moving forward with this also.

Luna712:
@fire-light42 any chance we can move forward with this so that #2342 can be merged?

PR #2080, opened 2025-10-26
Add new BaseFragment

Luna712:
This is designed to:

* Reduce code duplication
* Improve a bit of null safety around binding
* Fix an issue with edge to edge on orientation change when changed outside the current fragment (the main purpose of this actually to fix this without a ton of code duplication)

For now, this only converts HomeFragment, LibraryFragment, and QuickSearchFragment. I will gradually do PRs to move other fragments over after this is merged, I don't want to make this one PR overly large and do it all at once.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/BaseFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/BaseFragment.kt
@@ -15,6 +15,21 @@ import com.lagradost.cloudstream3.mvvm.logError
 import com.lagradost.cloudstream3.utils.txt
 import com.lagradost.cloudstream3.utils.UIHelper.fixSystemBarsPadding
 
+/**
+ * A base Fragment class that simplifies ViewBinding usage and handles view inflation safely.
+ *
+ * This class allows two modes of creating ViewBinding:
+ * 1. Inflate: Using the standard `inflate()` method provided by generated ViewBinding classes.
+ * 2. Bind: Using `bind()` on an existing root view.
+ *
+ * It also provides hooks for:
+ * - Safe initialization of the binding (`onBindingCreated`)
+ * - Automatic padding adjustment for system bars (`fixPadding`)
+ * - Optional layout resource selection via `pickLayout()`
+ *
+ * @param T The type of ViewBinding for this Fragment.
+ * @param bindingCreator The strategy used to create the binding instance.
+ */
 abstract class BaseFragment<T : ViewBinding>(
     private val bindingCreator: BindingCreator<T>
 ) : Fragment() {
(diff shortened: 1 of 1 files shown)


Luna712:
@fire-light42 just wandering your thoughts on this before I do other fragments, otherwise I'll just fix the edge to edge issue with orientation another way without duplication. The idea behind this was to reduce duplication beyond just that issue though, and to improve null safety with view binding etc... but was just wondering your opinion on going this route before I do others as well.

CranberrySoup:
I really like this idea!

app/src/main/java/com/lagradost/cloudstream3/ui/BaseFragment.kt
```diff
@@ -0,0 +1,58 @@
...
+        fixPadding(binding?.root)
+        super.onConfigurationChanged(newConfig)
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        _binding = null
+    }
+
+    private fun fixPadding(view: View?) {
```
fire-light42: companion object?

Luna712: I thought about it and might when this is done, I didn't because I was thinking maybe make it an open method that can be overriden if we need fragments to be able to differentiate this, which I havent moved many over to it yet so havent decided on that yet.

I also am still working on changing this a bit more to expand `onCreateView` more and handle multiple layout versions (TV and Phone differentiating) using unified logic so that it handles that and we can even further reduce code duplication.

Luna712: @fire-light42 I made it more expandable and easy to use now, which also allows overriding fixPadding so this can't really be done. I'm also done with this PR unless you want me to change it. I'll gradually migrate other fragments over so that we don't have one massive PR with this doing all at once. Also easier to review and debug if there are issues that way.

fire-light42: Yep, I see that, and I have to say, this is a very cool solution to the TV/Phone layout problem. I thought about it too but could not design a clean API for it, but your sealed classes are a very good fit for our case.

Luna712:
> I have not tested, but the code looks good and should work fine. However it might be good to add some documentation about the motivation and use cases to the common functions and BaseFragment itself as it will be used for every fragment. This is mainly to keep temporal consistency, and aim of the codebase more clear for future developers.

@fire-light42 

That was a good idea, added documentation to BaseFragment.

fire-light42:
While this pull request is mainly for clean code, I believe it should be possible to recycle bindings like we do for a recycleview, and if we are bold even do a companion object sharedPool to improve performance. 

Right now the lag spike when navigation is mainly due to inflate, and if the onViewCreated is properly coded it should be fine to reattach an old binding from a different instance or old instance. So while it might not be your goal I would suggest looking into recycling the entire fragment binding.

Luna712:
That's a great idea. I'll look into it.

PR #2090, opened 2025-10-29
Fix next episode autoplay issue

### LINKED ISSUE #2084: Videos can't be restarted or rewinded if reached the very end and autoplay next episode is enabled
### Steps to reproduce

1. Make sure autoplay next episode is enabled (issue doesn't happen without it)
2. Play a local video file, the last episode of a series, or any movie, anything where it can't "autoplay next episode"
4. Fast-forward it to say 2 seconds from the end and wait for it to reach the end
5. Once it has reached the end, try using any buttons or seekbar to rewind the video

### Expected behavior

The video should properly rewind.

### Actual behavior

The seekbar can be moved but the play button won't play it to resume at the new point, also the time doesn't update, the rewind button does nothing at all.

4.6.1 a1a5c36

### Android version

Android 16

### Logcat

```java

```

### Other details

I've noticed this issue for a very long time, however I could never figure out what triggered it until I finally decided to investigate more now, and realized its the autoplay next episode being enabled.

### Acknowledgements

Luna712:
Fixes #2084

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
@@ -513,6 +513,11 @@ abstract class AbstractPlayerFragment(
             }
 
             is PositionEvent -> {
+                val player = player
+                if (player is CS3IPlayer) {
+                    player.hasNextEpisode = hasNextEpisode
+                }
+
                 playerPositionChanged(position = event.toMs, duration = event.durationMs)
             }
 
@@ -560,7 +565,6 @@ abstract class AbstractPlayerFragment(
 
         val player = player
         if (player is CS3IPlayer) {
-            player.hasNextEpisode = hasNextEpisode
             // preview bar
             val progressBar: PreviewTimeBar? = playerView?.findViewById(R.id.exo_progress)
             val previewImageView: ImageView? = playerView?.findViewById(R.id.previewImageView)


app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -558,6 +560,7 @@ abstract class AbstractPlayerFragment(
 
         val player = player
         if (player is CS3IPlayer) {
+            player.hasNextEpisode = hasNextEpisode
```
fire-light42: I do not understand this logic, `player.hasNextEpisode` is only modified once in onViewCreated? Does that not imply that the bug still occurs when you have more than one episode?

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -519,6 +520,7 @@ abstract class AbstractPlayerFragment(
                 context?.let { ctx ->
                     // Resets subtitle delay on ended video
                     player.setSubtitleOffset(0)
+                    if (!hasNextEpisode) return@let
```
fire-light42, line 528: Why did you place this under setSubtitleOffset, is it not supposed to be above if you replay?

Luna712: Maybe I was just trying to fix the rewind issue which this does so not sure if we also want to stop setSubtitleOffset if no next episode as well?

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: It does not. I was looking at that myself and didn't fully understand why it worked but it did. I tested with all forms of player and it worked, next episode can autoplay if it has a next episode but it doesn't if it doesn't have one. Even if you start on a previous episode and get there with the next episode button or autoplay.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: I think we probably do want it to reset always because its not wrapped in tbe next episode autoplay which means even if not a next episode we probably want it to reset, no?

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: Well I have no idea why it worked but I moved it to the callback because I feel it would be more reliable there.

fire-light42:
I wrote a proper fix for this as the root cause was not because of the caller, but because the calle did not properly boundscheck before releasing the player. Futhermore this also fixes next episode shortcuts causing the same issue.

Luna712:
> I wrote a proper fix for this as the root cause was not because of the caller, but because the calle did not properly boundscheck before releasing the player. Futhermore this also fixes next episode shortcuts causing the same issue.

Awesome thanks for that. And makes a lot of sense too.

PR #2093, opened 2025-10-29
feat:separate episode text in continue watching

rockhero1234:


app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
```diff
@@ -263,10 +265,16 @@ object SearchResultBuilder {
                 }
 
                 playImg?.visibility = View.VISIBLE
+                if (card.type?.isMovieType() == false && showEpisodeLabel) {
+                    val context = cardText?.context
+                    val text = when {
+                        showTitle -> context?.getNameFull(card.name, card.episode, card.season)
```
fire-light42: I do not understand this line? If you disable title but enable episode label it looks very weird with no labels on movies but labels on tv-series. Futhermore this breaks the UI as it will push the UI up and down moving everything, if you have mixed set of both movies and tv-series.

rockhero1234: I tested working fine for me

fire-light42: In your continue watching have a row like this where you only have 3~4 items visible at the same time. 

"MMMMMEEEEEEMEMEMEMEE" where M = Movie and E = Episode. If you only have movies then it is hidden, but as soon as you scroll to episodes it pops in and moves the entire layout down due to wrap_content. Moving layouts after inflating it is a no-go for UX, and is only done in Cloudstream when it is really needed.

rockhero1234:
@fire-light42  what do you think about this

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
```diff
@@ -256,10 +260,10 @@ object SearchResultBuilder {
                 }
 
                 playImg?.visibility = View.VISIBLE
-
-                if (card.type?.isMovieType() == false) {
-                    cardText?.text =
-                        cardText.context?.getNameFull(card.name, card.episode, card.season)
+                cardText?.text = card.name
```
fire-light42: This line is unnecessary, see L132.

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
```diff
@@ -256,10 +260,10 @@ object SearchResultBuilder {
...
+                if (card.type?.isMovieType() == false && showEpisodeText) {
+                    episodeText?.text = episodeText.context?.getShortSeasonText(card.episode, card.season)
+                    episodeText?.isVisible = true
```
fire-light42: This is always visible, even if "getShortSeasonText" returns null/empty string. I would sanity check the return value here to prevent a blobs of "" showing up in the UI.

app/src/main/res/values/strings.xml
```diff
@@ -901,4 +903,5 @@
     <string name="top_left">Top left</string>
     <string name="top_center">Top center</string>
     <string name="top_right">Top right</string>
+    <string name="episode_text">S1E1</string>
```
fire-light42: Why is this translatable? Did you mean to write translatable=false?

rockhero1234:
@fire-light42 fixed

app/src/main/res/values/strings.xml
rockhero1234: I forgot to write that my bad.

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchResultBuilder.kt
```diff
@@ -258,12 +258,12 @@ object SearchResultBuilder {
                     bar?.progress = (pos.position / 1000).toInt()
                     bar?.visibility = View.VISIBLE
                 }
-
                 playImg?.visibility = View.VISIBLE
-                cardText?.text = card.name
                 if (card.type?.isMovieType() == false && showEpisodeText) {
-                    episodeText?.text = episodeText.context?.getShortSeasonText(card.episode, card.season)
-                    episodeText?.isVisible = true
+                    episodeText?.context?.getShortSeasonText(card.episode, card.season).let {text->
```
fire-light42: Did you mean `?.let`?

rockhero1234: Yeah!

PR #2097, opened 2025-10-29
Add BaseFragment compat for other Fragment subclasses

Luna712:
Adds compatability for `DialogFragment` (and `BottomSheetDialogFragment`) and `PreferenceFragmentCompat` to `BaseFragment` to unify logic between all the different fragment types we have (unless I missed others but I wrote it so its easy to expand later on if we need to). `BasePreferenceFragmentCompat` is pretty much a stub right now, only handling padding but we can further expand it later also. Unifying all the fragments into `BaseFragment` will also make it easier to do performance improvement on view inflation later on if we do manage to do that.

This will also, in theory, make it possible for extension developers who use settings fragments that use `BottomSheetDialogFragment` or `DialogFragment` to benefit from the unified logic and use it as well which would also add security against random crashes through the error handling in `BaseFragmentHelper`, if extension fragments don't inflate correctly.


fire-light42:
🔥

PR #2103, opened 2025-10-30
Migrate WebviewFragment to use BaseFragment

Luna712:


Luna712:
After this we finally only have `AbstractPlayerFragment`, `ResultFragmentPhone` (requires `AbstractPlayerFragment` to be done at the same time as it uses it), and `ResultFragmentTv` left I think. There is no point doing anything with `MiniControllerFragment` as that is different and unified logic wouldnt do it any good since its not using view binding. I'll do the rest another day.

PR #2104, opened 2025-10-30
fix: remove listeners before releasing exoplayer

y-hbb:
this might fix the TV player issue

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -122,6 +122,7 @@ const val toleranceAfterUs = 300_000L
 
 @OptIn(UnstableApi::class)
 class CS3IPlayer : IPlayer {
+    private var playerListener: Player.Listener? = null
     private var isPlaying = false
     private var exoPlayer: ExoPlayer? = null
         set(value) {
@@ -612,7 +613,10 @@ class CS3IPlayer : IPlayer {
                 // No documented exception, but just to be extra safe
                 logError(t)
             }
-
+            playerListener?.let {
+                removeListener(it)
+                playerListener = null
+            }
             stop()
             release()
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -613,6 +614,7 @@ class CS3IPlayer : IPlayer {
                 logError(t)
             }
 
+            exoPlayer?.removeListener(playerListener!!)
```
fire-light42: Do not do !!, do `playerListener?.let { ... }` instead

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1363,173 +1532,7 @@ class CS3IPlayer : IPlayer {
...
-                    event(ResizedEvent(height = videoSize.height, width = videoSize.width))
-                }
-
-                override fun onRenderedFirstFrame() {
-                    super.onRenderedFirstFrame()
-                    onRenderFirst()
-                    updatedTime(source = PlayerEventSource.Player)
-                }
-            })
+            exoPlayer?.addListener(playerListener!!)
```
fire-light42: This creates a huge diff, and is not safe. Do `exoPlayer?.addListener((...).also { playerListener = it })`

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
fire-light42: Do not forget to release the object itself too, by doing `playerListener = null`

y-hbb: done

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
y-hbb: done

fire-light42:
Just an FYI, but exoPlayer.release() calls listeners.release(), which in turn releases all listeners. So I do not believe that this pull request does anything at all.

y-hbb:
> Just an FYI, but exoPlayer.release() calls listeners.release(), which in turn releases all listeners. So I do not believe that this pull request does anything at all.

I can confirm that I am not seeing the issue anymore in TV. I have prerelease and debug build in TV. not sure if it's a temporary thing.

PR #2109, opened 2025-10-30
Add SimpleBaseFragment for ResultFragmentPhone compat

Luna712:


Luna712:
Just to note @fire-light42 this is mostly just to be consistent with other fragments in how it handles padding and not duplicate padding logic on both methods in `ResultFragmentPhone` directly, but I did as a separate PR rather than also making `ResultFragmentPhone` implement `fixPadding` with it just to see if its an okay solution, if its not and you'd rather me just do it directly feel free to just close this. Eventually we should be able to migrate `AbstractPlayerFragment` to use `BaseFragment` directly but this is a solution for now anyway as it'd require more significant refactoring to `AbstractPlayerFragment` to migrate to it.

fire-light42:
This pull request is overengineering, and makes the inheratance even worse than it already is. I agree that it should eventually become an BaseFragment, but having this SimpleBaseFragment in the meantime does not really accomplish anything. 

I am unsure about how exactly we should migrate the player, but I think we should make it into com.lagradost.cloudstream3.ui.PlayerView.kt instead of a fragment so it is embeddable anywhere. The inheritance in the player was a mistake, and we should ideally migrate everything to only inherit from BaseFragment for a much flatter architecture.

Luna712:
> This pull request is overengineering, and makes the inheratance even worse than it already is. I agree that it should eventually become an BaseFragment, but having this SimpleBaseFragment in the meantime does not really accomplish anything.
> 
> I am unsure about how exactly we should migrate the player, but I think we should make it into com.lagradost.cloudstream3.ui.PlayerView.kt instead of a fragment so it is embeddable anywhere. The inheritance in the player was a mistake, and we should ideally migrate everything to only inherit from BaseFragment for a much flatter architecture.

@fire-light42 

I absolutely agree with you about that. So yeah I guess I'll just fix padding for ResultFragmentPhone manually for now.

PR #2113, opened 2025-10-30
Fix system bars padding on orientation change in ResultFragmentPhone

Luna712:


Luna712:
Disclaimer: this is not tested as in theory it should work and is fairly minor.

fire-light42:
> Disclaimer: this is not tested as in theory it should work and is fairly minor.

Do not worry, I test every pull request

PR #2118, opened 2025-10-30
Fix entering PiP mode even when we shouldn't

### LINKED ISSUE #1807: enterPictureInPictureMode called even if not enabled
### Steps to reproduce

Somehow on Android 15 PiP is enabled when exiting the player. This is the default behavior and is changeable with a setting, but somehow this setting is not respected. On Android 15 this setting does nothing?

What makes this weirder is that if you remove all the enterPictureInPictureMode code it still works. Android 15 or Media3 overrides the behavior making it inconsistent between android versions. 

### Expected behavior

The player should pause and the app should exit. 

### Actual behavior

It enters pip mode. 

17c35840045c4af30684c6af8e90ef023c477210

### Android version

Android 15

### Logcat

```java
N/A
```

### Other details

N/A

### Acknowledgements

Luna712:
Because of `setAutoEnterEnabled(isPlaying)` in `PlayerPipHelper` we need to only set `isPlaying` if we can enter PiP, otherwise it will enter regardless of if `enterPictureInPictureMode` is actually called.

Fixes two issues:
1. PiP mode is activated for trailers
2. The PiP mode setting within CS3 has no affect

From my testing both were fixed with this.

`setAutoEnterEnabled` is also [much smoother](https://developer.android.com/develop/ui/views/picture-in-picture#smoother-transition) for modern devices using gesture navigation then `onUserLeaveHint` with `enterPictureInPictureMode` is as well. So I also added a check so the `onUserLeaveHint` version isn't triggered if using `setAutoEnterEnabled` so we aren't triggering both versions as we are now.

Fixes #1807

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
@@ -313,6 +313,8 @@ object CommonActivity {
     }
 
     fun onUserLeaveHint(act: Activity?) {
+        // On Android 12 and later we use setAutoEnterEnabled() instead.
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) return
         if (canEnterPipMode && canShowPipMode) {
             act?.enterPIPMode()
         }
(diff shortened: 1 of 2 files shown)


fire-light42:
This makes too much sense. Lucky us that this is fixable instead of an upstream issue.

PR #2119, opened 2025-10-31
Remove onConfigurationChanged from ResultFragmentPhone

Luna712:
I just realized we don't need this one here. Because `fixSystemBarsPadding` is not orientation dependent here since there are no nav views. So rerunning on configuration change is unnecessary in this case. That was my bad here.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint
 import android.app.Dialog
 import android.content.Intent
 import android.content.res.ColorStateList
-import android.content.res.Configuration
 import android.graphics.Rect
 import android.os.Build
 import android.os.Bundle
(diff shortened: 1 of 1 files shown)


fire-light42:
Configuration is not only limited to orientation, so I would keep this to prevent any incredibly rare, or future issue.

Luna712:
> Configuration is not only limited to orientation, so I would keep this to prevent any incredibly rare, or future issue.

@fire-light42 

Makes sense I guess but also no future issue *shouldn't* be possible because fixSystemBarsPadding is fixed, its never changed, its never conditional so running on configuration changes just recreates the same listener which could have some very minor inefficiencies. But also it probably won't affect all the much.  Shouldn't be possible doesn't always mean its not impossible with Android anyway.

fire-light42:
> Makes sense I guess but also no future issue _shouldn't_ be possible because fixSystemBarsPadding is fixed, its never changed, its never conditional so running on configuration changes just recreates the same listener which could have some very minor inefficiencies. But also it probably won't affect all the much. Shouldn't be possible doesn't always mean its not impossible with Android anyway.

Remember that setOnApplyWindowInsetsListener is only applicable to Android 11+. The statusbar height can change during the runtime of the program by software emulation, and it would not surprise me that some devices with multiple screen calls onConfigurationChanged when it switches its active screen. Right now it calls onViewCreated when the statusbar is changed during the runtime of the program, but it could be that on some current OS used onConfigurationChanged instead. 

It is not an inefficiency right now because it is not called in a hot loop nor calculates something expensive. This is the JVM, everything is inefficient by default if we use those standards.

Luna712:
Thats fair, you make a good point.

PR #2125, opened 2025-11-01
Performance improvements to homepage recyclers and coil

Luna712:


app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
```diff
@@ -679,6 +679,7 @@ class HomeFragment : BaseFragment<FragmentHomeBinding>(
                 homeViewModel.queryTextSubmit("")
             }
 
+            homeMasterRecycler.setHasFixedSize(true)
```
fire-light42, line 671: Are you sure about this?

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
```diff
@@ -89,79 +90,51 @@ open class ParentItemAdapter(
...
-                            )
-                        }"
-                    )*/
-                    //!recyclerView.canScrollHorizontally(1)
-                    if (!recyclerView.isRecyclerScrollable() && hasNext && expandCount != count) {
-                        expandCount = count
-                        expandCallback?.invoke(name)
+        val currentAdapter = (binding.homeChildRecyclerview.adapter as? HomeChildItemAdapter)
+            ?.apply {
+                if (isHorizontal != info.isHorizontalImages) isHorizontal = info.isHorizontalImages
```
fire-light42, line 98: isHorizontal and hasNext is not an expensive operation? Why are you doing this check?

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
```diff
@@ -89,79 +90,51 @@ open class ParentItemAdapter(
...
-                                1
-                            )
-                        }"
-                    )*/
-                    //!recyclerView.canScrollHorizontally(1)
-                    if (!recyclerView.isRecyclerScrollable() && hasNext && expandCount != count) {
-                        expandCount = count
-                        expandCallback?.invoke(name)
+        val currentAdapter = (binding.homeChildRecyclerview.adapter as? HomeChildItemAdapter)
+            ?.apply {
```
fire-light42, line 97: You forgot to update the clickCallback, nextFocusUp and nextFocusDown. As `sharedPool` is shared ___between___ adapters this needs to always be set.

app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
```diff
@@ -85,25 +89,29 @@ object ImageLoader {
         headers: Map<String, String>? = null,
         builder: ImageRequest.Builder.() -> Unit = {} // for placeholder, error & transformations
     ) {
-        // clear image to avoid loading & flickering issue at fast scrolling (e.g, an image recycler)
-        this.dispose()
+        if (imageData == null) {
```
fire-light42, line 92: What? `dispose` stops the current request and clears the image. I do not see what advantage this brings. Moreover, will this not break if we do `loadImage(url); loadImage(null)` ?

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
```diff
@@ -89,79 +90,51 @@ open class ParentItemAdapter(
...
-            if (isLayout(PHONE)) {
-                homeChildMoreInfo.setOnClickListener {
-                    moreInfoClickCallback.invoke(item)
-                }
+                })
+                binding.homeChildRecyclerview.adapter = adapter
             }
-        }
+
+        binding.homeChildRecyclerview.setRecycledViewPool(HomeChildItemAdapter.sharedPool)
```
fire-light42, line 126: I believe that you should set the pool before attaching the adapter. I do not have a source for this, but I remember reading it.

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
```diff
@@ -89,79 +90,51 @@ open class ParentItemAdapter(
...
+        binding.homeChildRecyclerview.setHasFixedSize(true)
```
fire-light42, line 127: Can you please double check this. I am unsure if we can make such assumptions if we recycle the entire homeChildRecyclerview, but I have not checked.

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
```diff
@@ -48,7 +49,7 @@ open class ParentItemAdapter(
 ) {
     companion object {
         val sharedPool =
-            RecyclerView.RecycledViewPool().apply { this.setMaxRecycledViews(CONTENT, 4) }
+            RecyclerView.RecycledViewPool().apply { this.setMaxRecycledViews(CONTENT, 8) }
```
fire-light42, line 52: Any reason why this needs to be 8? This is the parent view for each row, not the items themself.

app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
```diff
@@ -27,22 +27,24 @@ import coil3.util.DebugLogger
...
+import okio.Path.Companion.toOkioPath
 
 object ImageLoader {
 
     private const val TAG = "CoilImgLoader"
 
     internal fun buildImageLoader(context: PlatformContext): ImageLoader = ImageLoader.Builder(context)
             .crossfade(200)
-            .allowHardware(SDK_INT >= 28) // SDK_INT >= 28, cant use hardware bitmaps for Palette Builder
+            .allowHardware(SDK_INT >= 28) // SDK_INT < 28, can't use hardware bitmaps for Palette Builder
```
fire-light42, line 42: I think you misinterpreted the comment. The comment is "SDK_INT >= 28" and "can't use hardware bitmaps for Palette Builder", they are unrelated, and it does not make any sense to switch >= to <. See https://github.com/recloudstream/cloudstream/pull/1396/commits/3636d8ecbfa0a1ac2bead962bfa572fb0ded93c2

app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
```diff
@@ -85,25 +89,29 @@ object ImageLoader {
...
     ) {
-        // clear image to avoid loading & flickering issue at fast scrolling (e.g, an image recycler)
-        this.dispose()
+        if (imageData == null) {
+            this.setImageDrawable(null)
+            return
+        }
 
-        if(imageData == null) return // Just in case
+        // Only dispose if a different image is requested to avoid unnecessary reloads
```
fire-light42, line 97: Is this really a thing that happens? Does dispose really show up in a flamegraph?

app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
fire-light42: Did you mean to write 

```kotlin
this.dispose()
this.setImageDrawable(null)
```

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
```diff
@@ -89,79 +90,51 @@ open class ParentItemAdapter(
...
+                id = id + position + 100,
+                clickCallback = clickCallback,
+                nextFocusUp = binding.homeChildRecyclerview.nextFocusUpId,
+                nextFocusDown = binding.homeChildRecyclerview.nextFocusDownId,
+            ).also { adapter ->
+                adapter.isHorizontal = info.isHorizontalImages
+                adapter.hasNext = item.hasNext
+                adapter.submitList(info.list)
+
+                binding.homeChildRecyclerview.addOnScrollListener(object : RecyclerView.OnScrollListener() {
```
fire-light42, line 112: Same problem as the `You forgot to update the clickCallback`

fire-light42 (changes requested):
After thoroughly reading your pull request, I several potential problems. Some motivation of how it improves performance, and correctness statements would be appreciated. ❤️

app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
Luna712: Oh that was my had yeah whoops.

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
Luna712: Because sometimes there are many rows. That was the intent anyway.

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapter.kt
Luna712: Because I was randomly trying things and it ended up in there when I opened the PR. That won't be in the final version.

Luna712:
> After thoroughly reading your pull request, I several potential problems. Some motivation of how it improves performance, and correctness statements would be appreciated. ❤️

Thanks for the review alot I hadn't yet considered, but for the record I only opened this PR so that I could just put it out here, it was far from finished which is why I marked as a draft. I was trying to reduce the lag on the home page on some devices so I was trying things. Coil seems to be a major part of the expensive operations also, and setHasFixedSize can help also, but you also make a point if that is okay to do. I'm almost certain it should be fine on the master recycler, less so on the child ones. (or maybe I'm in reverse here and its the other way around).

app/src/main/java/com/lagradost/cloudstream3/utils/ImageModuleCoil.kt
Luna712: Well the idea was to not clear and reload the image every time but only if the actual source changes but its also possible I misinterpreted the intent behind dispose here.

Luna712:
Whoops was going through deleting old branches and accidentally deleted this one but it also reminded me to go back and look at it again.

PR #2127, opened 2025-11-02
Rename fixPadding to fixLayout and expand usage

Luna712:
This renames it to `fixLayout()` so that it can be used and still make sense for things beyond just `fixSystemBarsPadding()`. It merges `fixGrid()` into it in fragments that have that running both in `onBindingCreated()` and `onConfigurationChanged()`, and removes any default handling for padding so that if we don't want padding we don't have an empty override, and always override the method when we do want padding. Eventually I was thinking of requiring it be overridden still with errors when its not, unless a simple var is overridden like `override val needsPadding = false` to prevent potential bugs where this is accidentally forgotten on some fragment that needs it as well.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/BaseFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/BaseFragment.kt
@@ -17,7 +17,6 @@ import com.lagradost.cloudstream3.R
 import com.lagradost.cloudstream3.mvvm.logError
 import com.lagradost.cloudstream3.ui.settings.SettingsFragment.Companion.setSystemBarsPadding
 import com.lagradost.cloudstream3.utils.txt
-import com.lagradost.cloudstream3.utils.UIHelper.fixSystemBarsPadding
 
 /**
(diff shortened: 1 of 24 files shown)


fire-light42 (approved):

Luna712:
Thanks for merging @fire-light42, I was wondering your thoughts on the last part I mentioned in the description about enforcing `fixLayout()` being called without some sort of variable telling it it doesn't need it. The idea is just to future proof it, reducing the chance we have random bugs and make sure we don't forget padding in some fragments (which I've noticed happens in other apps all to often, including Google's own apps like even Play Store and Chrome under certain locations and conditions)

fire-light42:
Well sure you can add it by simply making the super throw a NotImplementedException. I do not really understand the use for a variable given that the implementation can just be `override fun fixLayout(view: View) = Unit` if it is not needed.

Luna712:
> Well sure you can add it by simply making the super throw a NotImplementedException. I do not really understand the use for a variable given that the implementation can just be `override fun fixLayout(view: View) = Unit` if it is not needed.

Fair enough about the override

PR #2139, opened 2025-11-04
Fix updating UI when changing some settings

### LINKED ISSUE #2130: Changing poster size doesn't always apply to all rows until app is restarted
### Steps to reproduce

1. Open homepage and scroll down to the bottom so all rows load
2. Go to settings and change poster size (I change to 15 just to make it very very noticeable when it works)

### Expected behavior

All posters should be changed without restarting the app.

### Actual behavior

Only some rows are changed, I've noticed it most where the top two rows aren't (the resume watching and bookmarks ones)

4.6.1 acbedd1

### Android version

Android 16

### Logcat

```java

```

### Other details

It seems to happen when you scroll all the way to the bottom of home page and stay there when you go to settings to change the poster size. This also happens on search when you have an ongoing query that is actually searching when you make the change to the poster size. Also, not sure if intentional or not but poster size doesn't apply at all to library or anywhere else I've noticed except search and home page, while most other places probably shouldn't as much the library would make sense to apply to.

### Acknowledgements

### LINKED ISSUE #2131: Changing Poster title location setting does not apply to all items until app is restarted
### Steps to reproduce

1. Open homepage and scroll all the way down
2. Go to settings and toggle "Poster title location"

### Expected behavior

It should change throughout the app where it was supposed.

### Actual behavior

It changes in random locations, it isnt even consistent per row on the home page, one item changes and another doesn't, some full rows don't, others all except the first item in the row change, others only the first item changed.

4.6.1 acbedd1

### Android version

Android 16

### Logcat

```java

```

### Other details

_No response_

### Acknowledgements

Luna712:
This fixes a few issues that exist in pre release but not in stable and also some more longstanding issues that also exist in stable.

Fixes #2130
Fixes #2131

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
@@ -1,6 +1,5 @@
 package com.lagradost.cloudstream3.ui.settings
 
-import android.content.SharedPreferences
 import android.os.Build
 import android.os.Bundle
 import android.view.View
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -28,6 +32,18 @@ import com.lagradost.cloudstream3.utils.UIHelper.hideKeyboard
 import com.lagradost.cloudstream3.utils.UIHelper.toPx
 
 class SettingsUI : BasePreferenceFragmentCompat() {
+    // Use a listener to apply when preferences are actually committed in
+    // order to prevent any potential race conditions.
+    private val posterSizeListener =
```
fire-light42: Even with the comment, I do not understand why you did it this way. Why not:
```kotlin
(getPref(R.string.poster_size_key) as? SeekBarPreference?)?.setOnPreferenceChangeListener { perf, newValue
```

Luna712: Because that runs when it's changed not when its committed, and updatePosterSize relies on the value of it so it needs to run when it's committed or there is a race condition.

Luna712: I updated the comment to be more clear.

fire-light42: Then use the newValue from setOnPreferenceChangeListener and update the updatePosterSize to support a value instead. Doing OnSharedPreferenceChangeListener is very dirty.

Luna712: Done

Luna712: I honestly really should have thought about doing it like that... I have no idea why I didn't, I spent way to much time overthinking this on how to handle that issue...

Well thanks for pointing that out, that is a much much better way to do this...

fire-light42 (approved):

PR #2141, opened 2025-11-04
Fix setting summary

Luna712:
Currently it only works after it is clicked, this fixes it to show the initial summary for bytes before clearing. I guess due to how the context is being accessed from view doesn't work initially, until clicked? I'm actually not certain why this works but it does for me.


fire-light42 (approved):
Wow, good catch. 

I assume this is because someone (https://github.com/recloudstream/cloudstream/commit/cd579a2230d89b370b72408e1a37f8f51cc05d0c) just pasted it from the onViewCreate but forgot that onCreatePreferences function does not have a variable called view so it used the fragment member variable view instead. 

It does not create any warning or error, but the fragment view is always null in onCreatePreferences and non-null after it has been set up.

PR #2146, opened 2025-11-05
Add jdk toolchain

Luna712:
For better support for building with later java versions which have deprecated compiling with Java 8 so it doesn't always give deprecation warnings.

Starting with AGP 9, building with Java 17 is no longer supported, Java 8 still works for compiling, but building will require Java 21 or later so this is preparing for that. It might still be out a bit but it's probably a good thing to start preparing for it.

This PR should have no affect to the current build process other than making it easier and more clear to build with more modern Java versions without loosing any legacy support or being spammed with a bunch of warnings.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -169,7 +169,7 @@ dependencies {
 
     // Android Core & Lifecycle
     implementation(libs.core.ktx)
-    implementation(libs.activity)
+    implementation(libs.activity.ktx)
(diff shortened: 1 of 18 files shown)


Luna712:
@fire-light42 could this be merged by chance? At least in theory (though I'm pretty sure) it shouldn't affect existing behavior at all just stop warnings when building with higher Java versions then 17 ensuring we keep supporting Java 8 as well.

PR #2148, opened 2025-11-05
Use JDK 21 in workflows

Luna712:
Starting with AGP 9, building with Java 17 is no longer supported, Java 8 still works for compiling, but building will require Java 21 or later so this is preparing for that. It might still be out a bit but it's probably a good thing to start preparing for it and using it here in workflows, etc... would also help to encourage extension developers and other developers who contribute to the app to use it more for a more gradual migration to it, akin to when we deprecate something so it doesn't take developers by surprise for those that still use versions lower than Java 21 to build.


Luna712:
Just to note AGP 9 moved from alpha to beta recently, which means our timeline for this is moving up. Bumping AGP isn't something we just can't do either due to how Android Studio (*unfortunately*) works. I personally feel we should prepare for this ahead of time also to give extensions more time to adapt to it as well.

PR #2155, opened 2025-11-06
Require `fixLayout()` to be overridden

Luna712:


fire-light42:
I know that I wrote that fixLayout could throw a NotImplementedError, but have you also considered just not implementing it? It is an interface so you can have no implementation and force the user to implement it that way as well. It is not a needed change, just wanted to highlight an alternative.

Luna712:
> I know that I wrote that fixLayout could throw a NotImplementedError, but have you also considered just not implementing it? It is an interface so you can have no implementation and force the user to implement it that way as well. It is not a needed change, just wanted to highlight an alternative.

Good idea, done.

PR #2161, opened 2025-11-07
Adding HubCloud and PixelDrain Improvement

phisher98:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/HubCloud.kt
```diff
@@ -0,0 +1,223 @@
...
+                    if (dlink.isNotBlank()) {
+                        callback.invoke(
+                            newExtractorLink(
+                                "$referer [BuzzServer]",
+                                "$referer [BuzzServer] $labelExtras",
+                                dlink,
+                            ) { this.quality = quality }
+                        )
+                    } else {
+                        Log.w("HubCloud", "BuzzServer: No redirect")
```
fire-light42: I would recommend using tag constants as opposed to hard-coding "HubCloud". Not a required change for this pull request.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/HubCloud.kt
```diff
@@ -0,0 +1,223 @@
...
+                            link,
+                        ) { this.quality = quality }
+                    )
+                }
+
+                text.contains("10Gbps", ignoreCase = true) -> {
+                    var currentLink = link
+                    var redirectUrl: String?
+
+                    while (true) {
```
fire-light42: Please avoid while true loops with dubious exit conditions and repeated network calls. We should avoid infinite app.get in any circumstances.

You need to make a max limit for the amount of redirects.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -842,6 +847,37 @@ suspend fun loadExtractor(
...
+suspend fun loadSourceNameExtractor(
+    source: String,
+    url: String,
+    referer: String? = null,
+    quality: Int? = null,
+    subtitleCallback: (SubtitleFile) -> Unit,
+    callback: (ExtractorLink) -> Unit,
+) {
+    loadExtractor(url, referer, subtitleCallback) { link ->
+        CoroutineScope(Dispatchers.IO).launch {
```
fire-light42: Please remove the `CoroutineScope(Dispatchers.IO).launch` as it is redundant.

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
phisher98: It won't work if I remove CoroutineScope and will give me errors

phisher98: Removed loadSourceNameExtractor for now as it gives Suspension functions can only be called within coroutine body.

PR #2165, opened 2025-11-07
Fix a few warnings

Luna712:


fire-light42:
UpdatedDefaultExtractorsFactory and the UpdatedMatroskaExtractor should ideally just suppress all warnings as they are copy pasted from media3 and as such should be read-only.

Luna712:
> UpdatedDefaultExtractorsFactory and the UpdatedMatroskaExtractor should ideally just suppress all warnings as they are copy pasted from media3 and as such should be read-only.

@fire-light42 

I think we may want to fix at least the one on L290 but I can file;suppress deprecations, if you think we should suppress all, do you know how, or do you know what name it is for that one on L290 if you don't think we should even fix it?

Luna712:
@fire-light42 I suppressed most if not all in both those files now.

PR #2166, opened 2025-11-07
Replace deprecated Thread.id in ExceptionHandler

Luna712:


fire-light42:
Such a stupid warning. Cant they just make the getId final.

PR #2169, opened 2025-11-08
More response player quality profile items

Luna712:
More response for portrait mode as well, so they don't grow massive. Uses the same size as account select linear profile sizes.

### DIFF
--- a/app/src/main/res/layout/player_quality_profile_item.xml
+++ b/app/src/main/res/layout/player_quality_profile_item.xml
@@ -8,8 +8,8 @@
 
     <androidx.cardview.widget.CardView
         android:id="@+id/card_view"
-        android:layout_width="100dp"
-        android:layout_height="100dp"
+        android:layout_width="110dp"
+        android:layout_height="110dp"
         android:layout_marginStart="5dp"
         android:layout_marginEnd="5dp"
         android:animateLayoutChanges="true"


app/src/main/res/layout/player_quality_profile_item.xml
```diff
@@ -8,9 +8,10 @@
 
     <androidx.cardview.widget.CardView
         android:id="@+id/card_view"
-        android:layout_width="0dp"
-        android:layout_height="0dp"
-        android:layout_marginStart="10dp"
+        android:layout_width="100dp"
```
fire-light42: I do not understand why this is so small. Even the small account profiles are 110dp, and this takes up the entire screen on a single line. 110dp-150dp looks much better.

Luna712: > I do not understand why this is so small. Even the small account profiles are 110dp, and this takes up the entire screen on a single line. 110dp-150dp looks much better.

Its even bigger then it is right now. And actually account select linear use 100dp as seen in a dimen, 100dp makes them even larger than they are at current though in landscape. This just prevents them being so super massive on portrait which in a follow-up I planned to add some grid design for portrait to look better in that case.

Luna712: https://github.com/recloudstream/cloudstream/blob/ef114f9271aa647752afe769b4c9589b6118600b/app/src/main/res/values/dimens.xml#L27

Luna712: @fire-light42 if you want me to use 110dp, I can by the way, I only used 100dp to keep it like the account select linear version since unlike AccountSelectActivity which is a grid this is a linear scroll like those are also so thought being consistent with that one was better but if you'd like I can still increase it.

fire-light42: > account_select_linear_item_size

account_select_linear_item_size is different from the account_list_item which is what I base it on as account_recycler_view is a horizontally scrollable view with the same layout requirements.

fire-light42: > Its even bigger then it is right now. And actually account select linear use 100dp as seen in a dimen, 100dp makes them even larger than they are at current though in landscape. This just prevents them being so super massive on portrait which in a follow-up I planned to add some grid design for portrait to look better in that case.

I assume this is because of screen size differences, but 100dp is very tiny to me.

Luna712: switched to 110dp

fire-light42 (approved):

PR #2172, opened 2025-11-08
Lock orientation when using the rotate button

### LINKED ISSUE #2026: Auto rotate and rotate button conflict
### Steps to reproduce

1. Make sure both auto rate and manual rotate button preferences are enabled
2. Play any video
3. Click the rotate button
4. Go home on your device
5. Go back to cloudstream

### Expected behavior

The rotation orientation is lost and reset to the auto detected video orientation.

### Actual behavior

The rotation should be maintained until the button is clicked again or the video is closed.

4.6.0 5bcf457

### Android version

Android 16

### Logcat

```java

```

### Other details

My assumption is this has something to do with it not saving the manual rotation state. It should likely save it until the video is closed and handle it in `onResume` so the state doesn't get lost.

### Acknowledgements

Luna712:
Fixes #2026

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -417,56 +417,41 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         activity.requestedOrientation = orientation
     }
 
-    open fun lockOrientation(activity: Activity, allowReverse: Boolean = false) {
-        val display = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
-            @Suppress("DEPRECATION")
+    open fun lockOrientation(activity: Activity) {
+        @Suppress("DEPRECATION")
+        val display = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R)
             (activity.getSystemService(Context.WINDOW_SERVICE) as WindowManager).defaultDisplay
-        } else activity.display!!
-
+        else activity.display!!
         val rotation = display.rotation
         val currentOrientation = activity.resources.configuration.orientation
         val orientation: Int
-
         when (currentOrientation) {
-            Configuration.ORIENTATION_LANDSCAPE -> {
-                orientation = if (allowReverse) {
-                    // Allow both landscape and reverse landscape
-                    ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
-                } else {
-                    // Lock to current landscape direction only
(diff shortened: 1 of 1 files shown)


Luna712:
> I do not understand this change, it just looks to me that you made lockOrientation into restoreOrientationWithSensor. The logic is confusing and I cant follow where and when this is applicable.

@fire-light42 

I changed lock orientation to allow reversing the orientation via sensor. Thats because when you lock with the lock button it locks completely (not even allow reverse) but when we manually change the orientation we only care about portrait or landscape so we still lock the orientation with lockOrientation, but we accept reversing to reverse landscape or reverse portrait modes still, as we only care about it locking to the user defined orientation.

fire-light42:
> > I do not understand this change, it just looks to me that you made lockOrientation into restoreOrientationWithSensor. The logic is confusing and I cant follow where and when this is applicable.
> 
> @fire-light42
> 
> I changed lock orientation to allow reversing the orientation via sensor. Thats because when you lock with the lock button it locks completely (not even allow reverse) but when we manually change the orientation we only care about portrait or landscape so we still lock the orientation with lockOrientation, but we accept reversing to reverse landscape or reverse portrait modes still, as we only care about it locking to the user defined orientation.

But why do you use the lockOrientation function for that? restoreOrientationWithSensor does exactly that?

Luna712:
> > > I do not understand this change, it just looks to me that you made lockOrientation into restoreOrientationWithSensor. The logic is confusing and I cant follow where and when this is applicable.
> > 
> > 
> > @fire-light42
> > I changed lock orientation to allow reversing the orientation via sensor. Thats because when you lock with the lock button it locks completely (not even allow reverse) but when we manually change the orientation we only care about portrait or landscape so we still lock the orientation with lockOrientation, but we accept reversing to reverse landscape or reverse portrait modes still, as we only care about it locking to the user defined orientation.
> 
> But why do you use the lockOrientation function for that? restoreOrientationWithSensor does exactly that?

@fire-light42 

I tried using restoreOrientationWithSensor, but it caused a conflict with locking instead, so using the lock button then changing with the rotation button unlocked the rotation, keeping it in lockOrientation kept it to behave like an actual locking behavior instead. But I can try to change it some other way if you want.

fire-light42:
> I tried using restoreOrientationWithSensor, but it caused a conflict with locking instead, so using the lock button then changing with the rotation button unlocked the rotation, keeping it in lockOrientation kept it to behave like an actual locking behavior instead. But I can try to change it some other way if you want.

If I understand the code correctly then you should be able to do use the old code and just modify `ignoreDynamicOrientation` to `ignoreDynamicOrientation || rotatedManually` to achieve the exact same effect.

Luna712:
Wow why didn't I think of that... you're absolutely right...

PR #2174, opened 2025-11-09
Use internal modifier rather than method name

Luna712:
This means it can't be used in plugins, only the same gradle module. I'm unsure if this is wanted but thought I'd just put it out there. If you don't want it do feel free to just close this.

It doesn't seem like any plugin at all actually does use them anyway.

Also we shouldn't use `@Suppress("DEPRECATION_ERROR")` on entire large methods or we may suppress to much.


fire-light42:
I still want to have the DEPRECATION_ERROR as it clearly shows: 
1. Be very careful when you use it
2. What/why you should be careful about using those functions
3. What you should use instead
4. You should opt-in to using this method

Previously these where comments, but no one actually read them. 

I understand that it might be unwise to have a huge `@Suppress("DEPRECATION_ERROR")`, but these functions are only used in a few places internally. Moreover, it is still possible to use "internal" functions (as seen in e.g. UpdatedMatroskaExtractor).

PR #2184, opened 2025-11-10
Fix one more build issue on future version of Kotlin

Luna712:
It's a warning on Kotlin 2.3 and will be an error in 2.4:

`Type annotation class 'Nullable' of the inferred type is inaccessible. Check the module classpath for missing or conflicting dependencies. This will become an error in language version 2.4.`


fire-light42 (commented):
I do not understand what exactly is wrong with the current code. Moreover, are you sure that .data is semantically equivalent to .html?

Luna712:
> I do not understand what exactly is wrong with the current code. Moreover, are you sure that .data is semantically equivalent to .html?

@fire-light42 

According to docs in jsoup, data() is the correct usage to pull from a script tag. And its used elsewhere here also. As for what is wrong, it just couldn't understand that selectFirst is nullable with the Nullable annotation in jsoup without it being explicit using data inline which, in my opinion is actually better anyway.

Luna712:
https://github.com/jhy/jsoup/blob/f2128717794737bb644e6e186de50b109d2c7349/src/main/java/org/jsoup/nodes/Element.java#L1721

PR #2191, opened 2025-11-11
minor ui fixes

rockhero1234:
unsure of second commit


app/src/main/res/values/styles.xml
```diff
@@ -1052,6 +1052,8 @@
         <item name="trackThickness">@dimen/circular_progress_thickness</item>
         <item name="android:layout_gravity">center</item>
         <item name="indicatorColor">@color/white</item>
+        <item name="trackCornerRadius">0dp</item>
+        <item name="indicatorTrackGapSize">0dp</item>
```
fire-light42, line 1055: I agree with the indicatorTrackGapSize=0, but feel like trackCornerRadius=0 is a bit too much.

rockhero1234: done

PR #2202, opened 2025-11-13
Revert "Allow to enter PIP mode if buffering"

Luna712:
Reverts recloudstream/cloudstream#2199

I can't reproduce any issues with it but it seems to cause some issues where PIP is entered after exiting the player while its still loading then going to another app. Not sure why or why I can't reproduce but we can just revert this for now.


Luna712:
Let's leave it for now as it may be possible to fix instead.

fire-light42:
Do not worry, I have a fix for this.

Luna712:
> Do not worry, I have a fix for this.

Thanks so much

PR #2206, opened 2025-11-13
Use version catalog for plugins

Luna712:
Closes #1823 as it applies the last part of that as well.


Luna712:
Just for the record I tested building this with everything that prerelease builds. The only thing that wasn't tested was dokka but I see absolutely no way dokka would break when everything else works with this.

PR #2207, opened 2025-11-13
Remove acra and replace AcraApplication with CloudStreamApp

Luna712:
Maintains some backwards compatible wrappers. If extensions use others besides these ones, if necessary, we can add more backwards compatible wrappers as well.

Closes #1479


Luna712:
In order to not make this PR even more massive it also only removes the five strings from the English (source) language, as I figured weblate would automatically remove the rest. If that isn't the case we could remove them separately.

Luna712:
I decided to remove messages from all languages now since this was already reviewed so shouldn't make it harder anymore and them not being in source language but being in other languages seems to cause a bunch of warnings. However, if you'd rather, I can easily revert that.

PR #2220, opened 2025-11-16
full series btn

rockhero1234:
For the providers which sources complete series in one video


rockhero1234:
Many providers use newmovieresponse with tvtype.series that's why I made this pr.

fire-light42:
> Many providers use newmovieresponse with tvtype.series that's my I made this pr. But I know

😐

PR #2228, opened 2025-11-17
UI fixes for portrait subtitle offset dialog

### LINKED ISSUE #2221: Some minor UI improvements to portrait subtitle offset dialog

This issue is just created so I don't forget about it, but I plan to do this myself.

First part, increase size of offset field, reducing padding on the sides of the + and - buttons so the field is larger:

Image

That should be pretty easy and just base it more on how it works on landscape just in a different position.

And second part, add inset support so that when typing in the offset integer field when the keyboard is visible, the subtitle offset section moves up so the field is still visible. This could potentially be done but adding IME inset support to move it up when shown.

The alternative option is to flip the entire position of the offset controls to be on the top, but I personally didn't think it looked as good on the top and was a little more difficult to use, especially one handed so a bit worse accessibility.

### Other details

_No response_

### Acknowledgements

Luna712:
Fixes #2221


fire-light42:
❤️

PR #2229, opened 2025-11-17
Fix showing input when entering search fragment

### LINKED ISSUE #2227: Keyboard doesn't pop up automatically when entering search
### Steps to reproduce

I'm not certain if this is really a bug. It's happened for a long time but I only now realized that it *appears* its intended to automatically pop up the keyboard when opening the search fragment, based on:
https://github.com/recloudstream/cloudstream/blob/0579a15f9b940b0af310e8a328a671f62ccc03f9/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt#L130-L132

### Expected behavior

The keyboard should appear.

### Actual behavior

The keyboard does not automatically appear even though the cursor does automatically appear in the search box, you still have to click it again for the keyboard to appear.

4.6.1 0579a15

### Android version

Android 16

### Logcat

```java

```

### Other details

I also now realized that when you first open the search fragment, the keyboard does not appear, but if you close the app and reopen it while its already in the search fragment but the keyboard is hidden, it automatically shows itself. So it only works if you're already in the search fragment, which I'm not certain if that is the expected behavior or not. If that is the expected behavior I think it should probably only show if it is shown by user focus before leaving the app and restore the state when opening it, or always show it when entering the search fragment, either way this seems like a bug to me.

### Acknowledgements

Luna712:
Fixes #2227

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt
@@ -75,6 +75,7 @@ import com.lagradost.cloudstream3.utils.UIHelper.dismissSafe
 import com.lagradost.cloudstream3.utils.UIHelper.fixSystemBarsPadding
 import com.lagradost.cloudstream3.utils.UIHelper.getSpanCount
 import com.lagradost.cloudstream3.utils.UIHelper.hideKeyboard
+import com.lagradost.cloudstream3.utils.UIHelper.showInputMethod
 import java.util.Locale
 import java.util.concurrent.locks.ReentrantLock
 
@@ -221,6 +222,10 @@ class SearchFragment : BaseFragment<FragmentSearchBinding>(
         currentSpan = view.context.getSpanCount()
         binding?.searchAutofitResults?.spanCount = currentSpan
         HomeFragment.configEvent.invoke(currentSpan)
+
+        // Fix focus being lost
+        val searchText = view.findViewById<TextView>(androidx.appcompat.R.id.search_src_text)
+        searchText?.post { searchText?.requestFocus() }
     }
 
     override fun onBindingCreated(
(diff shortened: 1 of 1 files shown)


Luna712:
I'm honestly unsure if this is even wanted, but it seems to be the intended behavior, I'm just still on the fence if this is a good behavior or not.

LagradOst:
Wowow, I would be very careful with this. We did this in the past, but had sooo many issues with the virtual keyboard. People were unable to  close it, it reopend when going back and it did not close when going into an item. I did not have the time to comment on your issue, but please do not touch this as android behavior is super strange on this thing.

Luna712:
> Wowow, I would be very careful with this. We did this in the past, but had sooo many issues with the virtual keyboard. People were unable to close it, it reopend when going back and it did not close when going into an item. I did not have the time to comment on your issue, but please do not touch this as android behavior is super strange on this thing.

@LagradOst 

Some of those issues happen without the PR. And most of them I had when trying with `setOnQueryTextFocusChangeListener` but not with `setOnFocusChangeListener` and it happened without the focus fix here. But it seemed not go here but behavior may also be inconsistent so I'm not sure. If you feel it should be this can just be closed though.

LagradOst:
The difference is that no one has actually reported any of those as an issue.

Luna712:
Thats fair and just for the record I only did this PR because it seemed like intended behavior, not some behavior I actually wanted myself. If it really isnt intended behavior it can just be closed no worries.

LagradOst:
> Thats fair and just for the record I only did this PR because it seemed like intended behavior, not some behavior I actually wanted myself. If it really isnt intended behavior it can just be closed no worries.

Alright. However, if we get strange reports about this issue, then we might need to revisit this, as the current implementation is like this mainly because no one has complained yet. I do not want anyone to touch a working implementation when so many issues have been caused by this in both cs3 and qn.

PR #2230, opened 2025-11-18
Upgrade Gradle wrapper to 9.2.1

Luna712:


Luna712:
Just for the record adding that comment makes it have to be manually readied everytime the wrapper update command is ran. Only reason I didn't add something for that because the command removed it. It's no problem, just thought I'd mention that.

PR #2239, opened 2025-11-19
Close overlapping panels on back press

### LINKED ISSUE #2222: Add back press handling to close gesture panels on mobile

On mobile when you open either the recommendations or sync panels, and then use the device back button, it leaves the entire results page and goes back to the home page. I think clicking the back button should first close the panels if they are open instead.

### Other details

I could probably do this myself if no one else gets to it first, so mostly just creating the issue so I don't forget about it as well, but if someone else wanted to do it also, that'd be fine as well.

### Acknowledgements

Luna712:
Closes #2222

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt
@@ -248,7 +248,7 @@ open class ResultFragmentPhone : FullScreenPlayer() {
         resultBinding = null
         syncBinding = null
         recommendationBinding = null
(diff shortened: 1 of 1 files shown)


Luna712:
> This is bugged on nested viewing. E.g. If you open a result page press on of the recommendations, then go back, the initial "ResultFragmentPhone" will get removed. I suggest doing toString on `this` to make the id of the callback unique.
> 
> `val callbackId = "ResultFragmentPhone$this"`

@fire-light42 

I *think* I did it? I'm a little confused though because I couldn't reproduce any issue. But I just did it like the ResultFragmentTv callback is done.

fire-light42 (approved):

PR #2240, opened 2025-11-20
Rename ResultViewModel2 to ResultViewModel

Luna712:
At one point there were actually two classes. Now there is only one. Calling it ResultViewModel2 doesn't really make any sense anymore. Using ResultViewModel class name makes it consistent with every other view model class as well.

However, if this is not wanted feel free to close.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultViewModel.kt
@@ -309,7 +309,7 @@ data class ExtractorSubtitleLink(
 
 fun LoadResponse.getId(): Int {
     // this fixes an issue with outdated api as getLoadResponseIdFromUrl might be fucked
-    return (if (this is ResultViewModel2.LoadResponseFromSearch) this.id else null)
+    return (if (this is ResultViewModel.LoadResponseFromSearch) this.id else null)
         ?: getLoadResponseIdFromUrl(uniqueUrl, apiName)
 }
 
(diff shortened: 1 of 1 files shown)


fire-light42:
> At one point there were actually two classes. Now there is only one. Calling it ResultViewModel2 doesn't really make any sense anymore. Using ResultViewModel class name makes it consistent with every other view model class as well.
> 
> However, if this is not wanted feel free to close.

The reason for keeping the "2" is not only because there was two classes, but because the git history that becomes easier to manage. However given that this has been like this for some time I understand the change. So why is this a draft?

Luna712:
Its only a draft to prevent it conflicting with some of my other open PRs first.

Luna712:
Mostly just #2207 actually as I didn't want to have to rectify a class rename merge conflict in that PR.

Luna712:
@fire-light42 wondering what your thoughts are about refactoring the whole app to use the `src/main/kotlin` directory instead of java? That would also make this one in particular follow the history better. But thats not the only reason. Like the library uses kotlin so thought maybe it'd be more consistent that way, but im not sure really so was just wondering your thoughts on it.

fire-light42:
> `kotlin`

This would majorly fuck up all history, it is totally irrelevant to change it to kotlin. I understand that the 2 may be confusing, but not /java/.

Luna712:
Fair enough for sure.

Luna712:
I'll do this some other time if I decide to. Won't keep fixing these type of merge conflicts so will have to do it sometime when there arent as many PRs open.

PR #2243, opened 2025-11-20
Add helper for default back press callback handling

Luna712:


Luna712:
Draft for now because I wanted to do this as a separate PR from #2239, but this will be updated to apply to that as well when/if that one is merged.

app/src/main/java/com/lagradost/cloudstream3/utils/BackPressedCallbackHelper.kt
```diff
@@ -7,18 +7,35 @@ import java.util.WeakHashMap
...
+        id: String,
+        callback: CallbackHelper.() -> Unit
+    ) {
+        val callbackMap = backPressedCallbacks.getOrPut(this) { mutableMapOf() }
         if (callbackMap.containsKey(id)) return
 
         val newCallback = object : OnBackPressedCallback(true) {
             override fun handleOnBackPressed() {
-                callback.invoke()
+                CallbackHelper(this@attachBackPressedCallback, this).callback()
```
fire-light42: Will this not hold a strong reference to activity? I am not super well versed in anonymous objects, but right here it looks like the anonymous object will capture this@attachBackPressedCallback and hold it as a strong reference. I am unsure if this is wise, because it *might* create a cyclic reference or something similar.

Luna712: > Will this not hold a strong reference to activity? I am not super well versed in anonymous objects, but right here it looks like the anonymous object will capture this@attachBackPressedCallback and hold it as a strong reference. I am unsure if this is wise, because it _might_ create a cyclic reference or something similar.

@fire-light42 

It does hold a strong reference but from my understanding it isn’t a leak because although the anonymous `OnBackPressedCallback` does hold a strong reference to the `Activity`, that reference is owned and scoped by the `Activity` itself via `OnBackPressedDispatcher` and the lifecycle; when the `Activity` is destroyed, the dispatcher removes its callbacks, breaking all references at once, and the `WeakHashMap` does not keep the `Activity` alive either. The `CallbackHelper` instance is not stored anywhere and exists only for the duration of a single method call, so it cannot retain anything beyond the stack frame. Cyclic references would be irrelevant here because at least from my understanding the garbage collector only cares whether the `Activity` is reachable from a GC root, and after destruction it is not.

I could be wrong anyway (and please correct me if I am) but that was my understanding of it.

fire-light42: I agree, but garbage collectors are very dumb... that is to say err on the safe side. I actually tested a minimal version of this code myself to prove this, and you can check it out. I actually think that a lot of functions in cs3 has some issue like this as cs3 leaks some data right now. 

```kotlin
import java.util.WeakHashMap

open class Activity { open val data : Int = 0 }
abstract class Callback { abstract fun f() }

val mapping : WeakHashMap<Activity, Callback> = WeakHashMap()

fun Activity.attachBackPressedCallback() {
    val callback = object : Callback() {
        override fun f() {
            // Simulate using activity by printing an internal field
            println("${this@attachBackPressedCallback.data}") // replace with a constant like 0
        }
    }
    mapping[this] = callback
}

fun attachAndKill() {
    val act = Activity()
    act.attachBackPressedCallback()
    // Here "act" goes out of scope and becomes garbage
}

fun main() {
    attachAndKill()
    println("Before GC: $mapping")
    System.gc()
    println("After GC: $mapping")
}
```

If you run this, you will see that act will not get collected, but it will if we remove its usage in f.

Luna712: Oh, that makes sense, thanks for the explanation. If I just passed it as a WeakReference like `val activity = WeakReference(this)` would that fix it then?

fire-light42: Yes, that will work. These implicit references are really hard to spot, but WeakReference will solve most GC problems as then it will not be able to depend on itself. Lambdas and anonymous objects only capture what variables are needed, and nothing more.

Luna712: Thanks a lot for that. That is now done.

fire-light42 (approved):
Looks good, will merge after testing!

PR #2249, opened 2025-11-21
Improve caching system for actions

Luna712:
This requires adding a new secret in GitHub settings, `GRADLE_ENCRYPTION_KEY` which can be generated with `openssl rand -base64 16`.

This significantly improves caching performance, not only supporting build cache, but configuration cache as well. The encryption key is necessary to enable that for security. It won't work without it.

Using this action also provides more information on build status in workflow results page. It also adds validation for wrapper jar files.

Marked as draft for now because while I have tested this some, I need to test this more before also adding to prerelease and other actions.


Luna712:
When it ends up using configuration cache it builds in 17 seconds. Note that configuration cache does get invalidated when almost anything is changed (including `.git/HEAD` which means any git commit at all) for safety (because the file is checked in exec so that it updates properly with the git commit displayed in settings). But this does have some advantages when rerunning builds or something, and if they change what invalidates it or something in the future.

Screenshot_20251121_155547_Chrome

And the result summary that would now be shown:
Screenshot_20251121_155619_Chrome

Luna712:
Im working on a new truly configuration cache supported version of the git commit, once that is finished this will be much more useful, in the meantime, all my testing indicates this works very well and also helps speed up the build drastically even when things like build.gradle.kts is changed, and only purges the cache and rebuilds when actual packages are changed making the cache smarter and faster in more circumstances as well.

Luna712:
This will remain a draft until #2206 is merged, as with that, I think I found a way to make the git commit stuff fully configuration cache compatible (but couldn't figure out how to using the legacy buildscript/classpath method) so it doesn't prevent reuse when `.git/HEAD` is changed. And when configuration cache is fully used, I got prerelease to build in 40 seconds (down from 10-15 minutes that it used to be)

Luna712:
I still need to make the git stuff truly support configuration cache but what I thought was working was actually just changing what invalides it (exec output instead of file) the 40 seconds for pre release does work when we don't invalidate (like if the git code is commented out) so if we can get it to support it will be a massive benefit (I think its possible its just more difficult then I thought for me anyway).

Luna712:
Combined with #2285 (which I FINALLY figured out... took way to long) this has a massive boost in build time when using configuration cache.

.github/workflows/pull_request.yml
```diff
@@ -13,11 +13,16 @@ jobs:
...
         java-version: 17
-        cache: gradle
 
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
 
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@v5
+      with:
+        cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
```
fire-light42, line 23: This would potentially allow attackers to leak secrets by just creating a pull request and reading the cache. Should this be used for pull requests?

https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#saving-configuration-cache-data

Luna712: > This would potentially allow attackers to leak secrets by just creating a pull request and reading the cache. Should this be used for pull requests?
> 
> https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#saving-configuration-cache-data

@fire-light42 

It won't be accessible from pull requests made outside of the source repo, (github doesnt allow secrets like that, I think), also we dont use secrets in PRs I think, so this just lets it work, and only a PR from one of you as far as I know. It helps if someone wants to generate their own in a fork for testing without more changes to the action which is why I left it there.

fire-light42: > also we dont use secrets in PRs I think

We do not. However, the pull request may be able to read cache from a previous pre-release. A specifically crafted pull request to this repository may therefore be able to extract the signing keys for pre-release.

fire-light42: It looks like a huge security risk to have unauthorized actions reuse cache from secret actions.

Luna712: > > also we dont use secrets in PRs I think
> 
> We do not. However, the pull request may be able to read cache from a previous pre-release. A specifically crafted pull request to this repository may therefore be able to extract the signing keys for pre-release.

@fire-light42 

Im not exactly sure how since it'd need the same cache encryption key in order to do that and GitHub doesnt share secrets with forks (I am pretty sure anyway) so it couldn't access the encryption key from this repository.

fire-light42: This is not about forks, this is about pull requests sent to this repository. `.github/workflows/pull_request.yml` runs whenever someone creates a pull request to this repository, with our secrets as far as i know. Please correct me if I am wrong.

Luna712: > This is not about forks, this is about pull requests sent to this repository. `.github/workflows/pull_request.yml` runs whenever someone creates a pull request, with our secrets.

@fire-light42 

No one can create a pull request other than from a fork unless they have direct push access to this repo in which case there are other ways to access that already.

Luna712: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow

"With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository."

A PR that triggers this workflow can only be done from a forked repository or direct push access and will otherwise not have access to secrets.

fire-light42: Sorry, I was under the distinct impression that the workflow `[pull_request]` triggered on receiving pulls, not creating pulls. I now see its like you say. My mistake.

fire-light42:
Thank you for your pull requests. More to be merged tomorrow :+1:

PR #2252, opened 2025-11-22
Replace srcDirs with directories

Luna712:
This is an error in AGP 9, but works in past AGP releases also. Just to prepare for AGP 9 we can just do this now.


Luna712:
Just for the information, this, in theory, is the very last PR to support future AGP 9, making it fully drop in once dokka supports, which I've been following their work on supporting it. Although it still also relies on a bunch of my other open PRs (I think about 5 or 6 of them) to both here and recloudstream/gradle for extensions as well.

Luna712:
Just an FYI I did confirm that with this change the generated `app-sources.jar` is the exact same size in bytes as it was before meaning it still generates properly.

PR #2253, opened 2025-11-23
Refactor PluginManager: store plugins in private files dir

### LINKED ISSUE #2076: PluginManager writing on internal storage
### Steps to reproduce

Use app

### Expected behavior

PluginManager shouldn't write on internal storage, only on app data folder 📂

### Actual behavior

PluginManager is writing on internal storage
/storage/emulated/0/Cloudstream3/plugins

4.6.1 9295795

### Android version

*

### Logcat

```java
PluginManager: Files in '/storage/emulated/0/Cloudstream3/plugins' folder: 0
```

### Other details

_No response_

### Acknowledgements

itsyourap:
This pull request updates the plugin storage directory to the app's private files directory, ensuring a clean state before loading plugins, and improving error handling when the plugins folder does not exist.

**Plugin directory management improvements:**

* Changed the plugin storage directory from the external files directory to the app's private files directory (`context.filesDir`) to avoid `SecurityException` when loading dex files.
* Cleaned up all non-existing plugins in the plugins directory by comparing them with the local plugins directory before loading new plugins to prevent stale plugins from persisting.
* If local plugin folder (in Internal Storage) does not exist, then do not create it and remove all local plugins currently installed.

**Error handling and plugin loading logic:**

* Improved handling when the local plugins folder does not exist by logging a message and marking plugins as loaded, instead of attempting to create the directory and logging a warning.
* Ensured that all local plugins are fully refreshed by removing the local plugins key before loading.

**Related Issues**

- Fixes #2076

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
@@ -532,7 +532,7 @@ object PluginManager {
             // Clean up all existing plugins since the local plugins folder doesn't exist
             // This is required since the user might stop using local plugins but still have old
             // plugins lying around taking up space.
-            // These cannot be deleted without manually by the user without root access
+            // These cannot be deleted manually by the user without root access
             pluginDirectory.listFiles()?.forEach { it.delete() }
 
             // Since there are no local plugins, we can consider them loaded


Luna712:
This won't fix the issue mentioned. It still writes on internal storage in two directories which is what the issue mentioned. Also if I remember correctly there was some issue with storing here. As far as I know there is no good way for dex, the read only issue happens if you don't follow the instructions to enable all files access. In the original folder that is an issue regardless but even here under some circumstances and under some conditions (not all the time though) it will still give dex issues.

Luna712:
I will test this more tomorrow, looking at it more and maybe it will work better because I think I tried something a little bit different then this initially which didn't properly work but I'll test this one more.

Luna712:
Always deleting might be bad for performance though, checking certain conditionals before deleting, only deleting if they are truly stale or outdated may be more efficient though I'm not certain.

itsyourap:
> only deleting if they are truly stale or outdated may be more efficient

Yes that can be done. I am working on it.

Luna712:
Thanks a lot. I'll test this tomorrow. But I think it could probably work a lot better. Thanks for working on this. Local plugjns have proven to be a bit of a challenge due to Android security and dex handling now.

PR #2256, opened 2025-11-23
Replace QuickJS with Zipline

Luna712:
QuickJS was renamed to Zipline all the way back in 2021. Unlike old QuickJS, newer Zipline versions are 16kb aligned. Current Zipline is also compatible back to minSdk 21.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -173,9 +173,8 @@ dependencies {
     implementation(libs.core.ktx)
     implementation(libs.activity.ktx)
     implementation(libs.appcompat)
-    implementation(libs.bundles.navigationKtx)
-    implementation(libs.lifecycle.livedata.ktx)
-    implementation(libs.lifecycle.viewmodel.ktx)
+    implementation(libs.bundles.lifecycle)
+    implementation(libs.bundles.navigation)
 
     // Design & UI
     implementation(libs.preference.ktx)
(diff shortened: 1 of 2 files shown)


Luna712:
It actually looks like the only extension that even used this no longer exists anymore. @phisher98 is this even still necessary to exist?

phisher98:
It's good to have I think if needed to deof js like webcrack else it's upto u guys @Luna712

Luna712:
We can just keep this in that case unless @fire-light42 feels differently? But since it's not currently used right now also, this should be able to merged and is a no-op that shouldn't cause any issues at all with any extensions etc... (even though its mostly the same as QuickJS imports would need a different package so if ut was actually used this PR would be a bit more difficult, but since it isn't right now it should be easy to merge)

Luna712:
@fire-light42 wondering if this could be merged? Shouldn't affect any extensions now but this one should be used if any extensions decide to use it and the longer we wait the more likely some extensions ends up using QuickJS and this will break something. If you have an issue with it thats fine though it can still wait.

fire-light42:
I would still like to mark it as deprecated with an appropriate waiting period even if it is only a package change but that is not  an option with external libraries. I will merge it but perhaps we should create wrappers for these tools in the future, to facilitate easy replacement and usage in extensions.

PR #2261, opened 2025-11-25
Add explicit dependsOn for copyJar

Luna712:
So we don't care about tbe run order of the tasks, it can just do it properly itself no matter what order it's ran in.


fire-light42:
Great pull request!

PR #2273, opened 2025-11-28
Move untranslatable strings to seperate file

Luna712:
This could cause crashes or poisoned data on some languages as some untranslatable strings were being translated, including keys and format strings that shouldn't be translatable. Also when translating the episodes key on weblate it caused a conflict between the plural version (which weblate does support) and the actual episodes key, meaning the episodes key was translating as the singular version of the plural episodes version in some cases. Moving to a separate resource file should hopefully prevent these issues.


Luna712:
Just for the record I confirmed some language like Arabic has fully translated keys and thus when switching to Arabic the data is lost and not saved between languages. That happens with a few keys in a few languages. The auto rotate and rotate button keys are present in numerous languages most notably. Also I confirmed a few wrongly formatted translated keys that shouldn't be do in fact cause the app to fully crash in safe mode in some languages.

This should prevent the issues because by using `donottranslate-*.xml`, we don't even need to define `translatable="false"`, they always aren't translatable and helps to keep that known, and also weblate will only translate `strings.xml` anyway.

This PR keeps all keys that were present before, but in the future I will also do a PR to remove the unused ones.

This also removes `subtitles_encoding_values` from `array.xml` in languages other than the source language. It's not translated by weblate anyway but does have `translatable="false"` which indicates it shouldn't have been translated anyway, so this just removed but kept in the main `values/array.xml` untouched.

Luna712:
#2257 should probably be merged and this rebased (if necessary) and merged immediately after to prevent potential conflicts with weblate as well.

Luna712:
@fire-light42 could this perhaps be merged now if your able, to prevent further conflicts with weblate?

fire-light42:
Thank you for the contribution! I am sorry for the excessive delay.

Luna712:
> Thank you for the contribution! I am sorry for the excessive delay.

@fire-light42

No problem at all. Thanks for merging!

PR #2282, opened 2025-11-30
Add plot and score to `DownloadHeaderCached`

Luna712:
In preparation for adding more details to download items, and to be the same as what `DownloadEpisodeCached` has.


Luna712:
Marked as draft so this isnt merged so it doesn't further conflict with the other download rework PR

Luna712:
Will do this some other time.

PR #2284, opened 2025-12-01
Fix: match packed JavaScript with space

diogob003:
Match packed JavaScript with spaces as in:
- `function(p, a, c, k, e, d)`
- `function(p,a,c,   k,  e,  r)`
...

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
@@ -797,7 +797,7 @@ fun getQualityFromName(qualityName: String?): Int {
     }?.value ?: match.toIntOrNull() ?: Qualities.Unknown.value
 }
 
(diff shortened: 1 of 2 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -797,7 +797,7 @@ fun getQualityFromName(qualityName: String?): Int {
     }?.value ?: match.toIntOrNull() ?: Qualities.Unknown.value
 }
 
-private val packedRegex = Regex("""eval\(function\(p,a,c,k,e,.*\)\)""")
+private val packedRegex = Regex("""eval\(function\(p,\s*a,\s*c,\s*k,\s*e,.*\)""")
 fun getPacked(string: String): String? {
     return packedRegex.find(string)?.value
```
diogob003: Is this expected to return the entire packed JavaScript?

PR #2285, opened 2025-12-02
Add true configuration cache support for git commit hash

Luna712:
This prevents the configuration cache from rebuilding every single time there is a new commit, which can drastically improve build performance, especially when there are no Kotlin files changed. I tested this on pre release builds, once with no files changed that is in the build (just changing GitHub workflows) and pre release built in 38 seconds. The second time, I tried just changing an XML resource and it built in about 2 minutes. Third time changing a Kotlin file (which requires recompiling it all) and it ran in about 6 minutes, which is still drastically faster than without configuration cache at all.

There were other ways to do this, I tried numerous ways including:
1. Have it generate a strings resource file, but this method requires resources be rebuilt every single build, which from my testing takes about 45 seconds.
2. Have it generate a Kotlin file that can just access a field from it, but this requires Kotlin to rebuild every single time even if no Kotlin files are changed, which takes about 4 minutes of extra build time when no other Kotlin files are changed.

Using an asset only requires merging new assets every single time which only takes about 3 seconds which is very performant and allows other gradle features aimed at speeding up build to work and allow ensuring that files that don't need rebuilding aren't. This prevents the reclaculating of configuration cache every new commit allowing it to work properly. Runtime performance seems to not be affected at all when reading the asset rather than a `resValue`. Seems to have no affect at runtime.

This needs to be combined with #2249 to have much impact here on the GitHub builds though.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -73,10 +73,6 @@ android {
         includeInBundle = false
     }
 
-    viewBinding {
-        enable = true
-    }
-
     androidComponents {
         onVariants { variant ->
             variant.sources.assets?.addGeneratedSourceDirectory(
@@ -193,6 +189,7 @@ android {
 
     buildFeatures {
         buildConfig = true
+        viewBinding = true
     }
 
     packaging {
--- a/app/src/main/java/com/lagradost/cloudstream3/AcraApplication.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/AcraApplication.kt
@@ -1,103 +1,78 @@
 package com.lagradost.cloudstream3
 
-import android.content.Context
-import com.lagradost.api.setContext
-import com.lagradost.cloudstream3.utils.DataStore.getKey
-import com.lagradost.cloudstream3.utils.DataStore.removeKeys
-import com.lagradost.cloudstream3.utils.DataStore.setKey
-import java.lang.ref.WeakReference
-
 /**
  * Deprecated alias for CloudStreamApp for backwards compatibility with plugins.
  * Use CloudStreamApp instead.
  */
-// Deprecate after next stable
-/*@Deprecated(
+@Deprecated(
     message = "AcraApplication is deprecated, use CloudStreamApp instead",
     replaceWith = ReplaceWith("com.lagradost.cloudstream3.CloudStreamApp"),
     level = DeprecationLevel.WARNING
-)*/
+)
 class AcraApplication {
-	// All methods here can be changed to be a wrapper around CloudStream app
-	// without a seperate deprecation after next stable. All methods should
-	// also be deprecated at that time.
 	companion object {
 
-		// This can be removed without deprecation after next stable
-		private var _context: WeakReference<Context>? = null
-		/*@Deprecated(
+		@Deprecated(
 		    message = "AcraApplication is deprecated, use CloudStreamApp instead",
 		    replaceWith = ReplaceWith("com.lagradost.cloudstream3.CloudStreamApp.context"),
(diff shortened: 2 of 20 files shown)


Luna712:
@fire-light42 would it be possible to get this reviewed and (hopefully) merged? That way we can also hopefully move forward with #2249 at some point as well.

Luna712:
@fire-light42 sorry for the ping again on this but any chance to review and potentially merge?

app/build.gradle.kts
```diff
@@ -15,23 +16,44 @@ val javaTarget = JvmTarget.fromTarget(libs.versions.jvmTarget.get())
...
+        val gitHashDir = layout.buildDirectory.dir("generated/git")
+
+        doLast {
+            val assetsDir = outputs.files.singleFile
+            val gitHashFile = gitHashDir.get().file("git-hash.txt").asFile
+            val outFile = File(assetsDir, "git-hash.txt")
+            gitHashFile.copyTo(outFile, overwrite = true)
+        }
     }
 }
```
fire-light42:
```suggestion
abstract class GenerateGitHashTask : DefaultTask() {

    @get:InputFile
    @get:PathSensitive(PathSensitivity.RELATIVE)
    abstract val headFile: RegularFileProperty

    @get:InputDirectory
    @get:PathSensitive(PathSensitivity.RELATIVE)
    abstract val headsDir: DirectoryProperty

    @get:OutputDirectory
    abstract val outputDir: DirectoryProperty

    @TaskAction
    fun generate() {
        val head = headFile.get().asFile

        val hash = try {
            if (head.exists()) {
                // Read the commit hash from .git/HEAD
                val headContent = head.readText().trim()
                if (headContent.startsWith("ref:")) {
                    val refPath = headContent.substring(5) // e.g., refs/heads/main
                    val commitFile = File(head.parentFile, refPath)
                    if (commitFile.exists()) commitFile.readText().trim() else ""
                } else headContent // If it's a detached HEAD (commit hash directly)
            } else ""  // If .git/HEAD doesn't exist
        } catch (_: Throwable) {
            "" // Just set to an empty string if any exception occurs
        }.take(7) // Get the short commit hash

        val outFile = outputDir.file("git-hash.txt").get().asFile
        outFile.parentFile.mkdirs()
        outFile.writeText(hash)
    }
}

val generateGitHash = tasks.register<GenerateGitHashTask>("generateGitHash") {
    val gitDir = layout.projectDirectory.dir("../.git")

    headFile.set(gitDir.file("HEAD"))
    headsDir.set(gitDir.dir("refs/heads"))

    outputDir.set(layout.buildDirectory.dir("generated/git"))
}
```

app/build.gradle.kts
fire-light42:
```suggestion
    androidComponents {
        onVariants { variant ->
            variant.sources.assets?.addGeneratedSourceDirectory(
                generateGitHash,
                GenerateGitHashTask::outputDir
            )
        }
    }
```

app/build.gradle.kts
Luna712: This seems like a great solution, I will do some testing with it myself in the next day or two to ensure it works fully with configuration cache without issues. But your solution seems like it will to me.

Luna712: Tested and it did seem to work. Thanks for that suggestion!

Luna712:
@fire-light42 I *think* this should be okay now?

Luna712:
@fire-light42 sorry for the ping on this again, and I know there is probably more important things, but any chance this can be merged soon? I am kinda just wanting to clear out my local branches before working on to much more.

PR #2291, opened 2025-12-02
Move app version to BuildConfig

Luna712:
Also, the intent seems to be to be to set the version to `-PRE` when in pre release, which doesn't currently work, but this fixes that. If that isn't the desired behavior that part can be removed though.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -65,7 +65,6 @@ android {
         versionCode = 67
         versionName = "4.6.1"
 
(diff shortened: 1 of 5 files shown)


Luna712:
Actually, I think we could rework this, removing the need for either resValues or BuildConfig here by using the same method to get current version from package that the update dialog uses. Not 100% certain on other affects that it'd have, but avoiding more work in Gradle would probably be good also.

PR #2292, opened 2025-12-02
Fix compiling library with debug flag

Luna712:
I'm not even sure if this is necessary for the library to even have a debug flag as no one noticed anything different, but at current it never works to compile as debug and always uses release.


Luna712:
Note: im not sure but this could also end up conflicting with build cache causing some issues as well.

Luna712:
Did a different way with #2330 instead.

PR #2296, opened 2025-12-03
Remove unused swiperefreshlayout dependency

Luna712:


gradle/libs.versions.toml
```diff
@@ -39,7 +39,7 @@ qrcodeKotlin = "4.5.0"
 rhino = "1.8.0"
 safefile = "0.0.8"
 shimmer = "0.5.0"
-swiperefreshlayout = "1.1.0"
+swiperefreshlayout = "1.2.0"
```
fire-light42: This is currently not even used. It is better if we remove it.

Luna712: Good idea

Luna712: Done

fire-light42 (approved):

PR #2300, opened 2025-12-05
Add lint.xml and add ignores to it

Luna712:
It's more expandable later on this way. Disabling `ByteOrderMark` is also the very last lint error, then once that and #2273, which also fixes the only instance of `ImpliedQuantity` lint error is merged, all lint errors will be fixed and we can enable `failOnError` to prevent potential future issues like #2320. I think #2301 also needs merged though as it fixes error level lint errors in library (well actually it fixes all lint levels in library), but I don't think that has any affect since before using the new KMP plugin it only checks android-only classes which has no lint issues now anyway.

### DIFF
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <lint>
+    <!-- ByteOrderMark has errors in values-b+ja/strings.xml, but it's handled by weblate so we don't really care. -->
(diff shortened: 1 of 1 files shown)


Luna712:
@fire-light42 this should be a fairly easy merge (#2301 also probably), sorry for the ping but trying to get some of my old PRs merged as my setup is getting hard to maintain so trying to get some past PRs merged so ai can clean it up. If there are issues though that's fine it can still wait.

PR #2305, opened 2025-12-06
Use withStyledAttributes

Luna712:
It is safer against memory leaks and doesn't require calling `recycle()`.

This also refactors methods in UIHelper, `getResourceColor`, `colorFromAttribute`, and `adjustAlpha` in order to only call `withStyledAttributes` once and use the other two methods as shortcuts in `getResourceColor`.


fire-light42:
Good pull request, but I do not want to cause merge issues with #2037, So I will leave this to after the download PR is merged.

Luna712:
@fire-light42 I rebased this now after the download PR was merged

PR #2306, opened 2025-12-07
Some minor miscellaneous cleanup

Luna712:


app/src/main/java/com/lagradost/cloudstream3/ui/HeaderViewDecoration.kt
```diff
@@ -3,6 +3,7 @@ package com.lagradost.cloudstream3.ui
 import android.graphics.Canvas
 import android.graphics.Rect
 import android.view.View
+import androidx.core.graphics.withTranslation
 import androidx.recyclerview.widget.RecyclerView
 
 class HeaderViewDecoration(private val customView: View) : RecyclerView.ItemDecoration() {
```
fire-light42: This class is unused, just remove this and setHeaderDecoration instead.

Luna712: Done

PR #2307, opened 2025-12-07
Remove "Too short playback" error

### LINKED ISSUE #2218: Too Short Playback
### Steps to reproduce

If a video is shorter than 20 seconds, attempting to play it triggers a “too short playback” error and the video exits immediately instead of playing.

Image

### Expected behavior

For videos under 20 seconds, the player should not throw an error. Instead, one of the following approaches would improve the experience:

- The video could load in a paused state and only start when the user presses play
- The video could start with a slight delay (e.g., 0.5 seconds) before playback begins.

Since the video is very short, introducing a small start delay would allow it to play smoothly without causing issues.

### Actual behavior

Image

4.6.1

### Android version

15

### Logcat

```java

```

### Other details

this issue is very annoying for adult contents. If it would support below 20 second videos it would be perfect.

### Acknowledgements

Luna712:
Per https://github.com/recloudstream/cloudstream/issues/2218#issuecomment-3537293235 this was added for a certain provider a long time ago, however with some providers such as Invidious, Dailymotion, and Internet Archive, there may often be videos shorter than 20 seconds. This just removes the error entirely as it's no longer necessary and if a certain provider needs to do this, they can handle it a different way.

Fixes #2218

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -557,6 +557,7 @@ abstract class MainAPI {
      * */
     open val loadTimeoutMs: Long? = null
 
+
     /**
      * A set of which ids the provider can open with getLoadUrl()
      * If the set contains SyncIdName.Imdb then getLoadUrl() can be started with


Luna712:
I'm actually not 100% sure if this is necessary or if this should simply be removed. In theory providers can stop playing or showing or even loading links at all in a different way then stopping at the actual player load time.

fire-light42:
> I'm actually not 100% sure if this is necessary or if this should simply be removed. In theory providers can stop playing or showing or even loading links at all in a different way then stopping at the actual player load time.

Yes, I think we should remove this almost arbitrary limit. The limit was only useful for something that is not an issue anymore.

Luna712:
> > I'm actually not 100% sure if this is necessary or if this should simply be removed. In theory providers can stop playing or showing or even loading links at all in a different way then stopping at the actual player load time.
> 
> Yes, I think we should remove this almost arbitrary limit. The limit was only useful for something that is not an issue anymore.

@fire-light42 

Removed entirely now.

PR #2308, opened 2025-12-08
Download selection fix + sub del fix + Del dialog fix

LagradOst:
After some discussion with @Luna712 I decided to fix some minor issues with the old code.

1. Now subtitles are deleted properly using the correct SafeFile api
2. Now delete sections on episodes + movies act way more consistent as the state of "isSelectingItemsRightNow" is now coupled with the type (eg, null = not in selection, empty = no items selected, {a,b,c} = selected a, b and c). I also cleaned up the code around that part too. 
3. Fixed a minor issue with the delete dialog to prevent "delete null?" by reordering the when stmt.   

Yall are free to test it out, but I will probs merge it soon.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadChildFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadChildFragment.kt
@@ -1,16 +1,16 @@
 package com.lagradost.cloudstream3.ui.download
 
 import android.os.Bundle
-import android.os.Handler
-import android.os.Looper
 import android.text.format.Formatter.formatShortFileSize
 import android.view.View
+import androidx.core.view.isGone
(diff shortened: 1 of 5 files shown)


LagradOst:
> This fixes all the issues I previously mentioned, but I also noticed another issue with this (though preexisting maybe even on stable) if you select something the multi delete button bytes don't update when total bytes for a certain item change, until you leave the downloads fragment and return to it. I can look into that one though later if you want.

This has always been the case as the actual byte count on each item is rerendered by hooking private state, that is to say the downloadviewmodel does not recalculate byte count at all. This is also not a bug/feature this pull request seeks out to fix. 

Moreover, I also do not want anyone to touch that before the download pr #2037 is done, as that modifies that part of the business logic quite a bit.

PR #2317, opened 2025-12-11
feat(home):stick profile and search at top

rockhero1234:


fire-light42:
I am unsure about this change from a UI perspective. I can understand the change, but the current UI just looks very clean. Both search and account is not that necessary for quick access.

PR #2326, opened 2025-12-13
Replace the "Update to prereleases" setting with a new button

Luna712:
The current way this setting is done is very confusing on stable. When you have this enabled, whenever you enter stable it will look for pre-release updates, and updating will actually update the pre-release version of the app instead, never doing stable releases itself. We don't need to actually care about pre-release *updates* within stable, so this replaces the former setting with a new button preference, that when clicked runs the exact same as the manual update button preference does, accept only targeting pre-release, and only if it is not installed at all yet, which has the affect of just installing the pre-release version, easily, from within stable.

This also has other benefits:
* We no longer need to have the `is_prerelease` tracked at all in `resValues`.
* We no longer have to track any preference key at all, it's only used to configure the preference click action and visibility, never neads to track, save, or do anything with it.
* We have more consistent behavior on all versions of the app, by explicitly defining in the updater the version we need to pull by `BuildConfig`, which prevents all sorts of potential very odd behavior.
* We no longer get confusing or unexpected behavior on stable.
* We no longer have two conflicting and confusing preferences on pre-release ("Update to prereleases" and "Show app updates").
* We no longer get updates on **CloudStream Debug** which never work anyway and gets quite annoying.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -1201,6 +1201,8 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
             val lastAppAutoBackup: String = getKey("VERSION_NAME") ?: ""
             if (appVer != lastAppAutoBackup) {
                 setKey("VERSION_NAME", BuildConfig.VERSION_NAME)
+                if (lastAppAutoBackup.isEmpty()) return@safe
+
                 safe {
                     backup(this)
                 }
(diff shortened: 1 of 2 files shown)


fire-light42:
Well the toggle may be a mistake, but I still do not see why we need to completely remove the setting. It can be made into a separate button instead.

Luna712:
> Well the toggle may be a mistake, but I still do not see why we need to completely remove the setting. It can be made into a separate button instead.

@fire-light42 

The button wouldnt be a setting though, we could just have something that says install prerelease version or something but it doesnt need to save a key on stable that prerelease is installed or anything, and even a button is kinda an edge case, but even so it would probably reset the translation so that can be cleared and a new message key be added for it as well. The current method conflicts between stable and pre. It doesn't need to show pre updates in stable at all, so we don't need to know when it exists we can just add a quick way to install pre with a link to the pre release version on github or something (or have it install but not track uodates for it and just say "prerelease already installed" or something by tracking if the package is installed using system checks, not any actual setting. At least in my opinion. But if you have a different opinion that's fine too it can be adjusted.

Luna712:
I am working on a new implementation of this using a button

Luna712:
Switched it to a new button shown only on stable now @fire-light42.

PR #2330, opened 2025-12-13
Use a new method to pass if debug to fix debug logging in library

Luna712:
Currently debug logging doesn't work at all, because it never works to know if its compiled with the debug flag, that would be easy to fix, however, that causes some conflict with build and configuration cache. Additionally, libraries should not directly depend on a compile flavor. We instead set a variable in the library from the `Application.onCreate()` method in order for the *app* to tell the *library* whether it is debug or not.

This also adds a new annotation, `InternalAPI` that is used for `AppDebug` since only the app should ever set that. If this part is not wanted, I can remove it, but it seems to be more standard also in other libraries to use an annotation like this for internal-only stuff. 

We could also replace some permanently/unstable API usages with this annotation, such as the ones in `PluginManager`, which is a better usage than `@Deprecated`. Either way whether deprecated or this annotation is used, extensions could suppress it, but using this annotation like this would prevent us internally having to add suppressions, and make it overall easier to manage, as well as be semantically correct, as deprecated implies eventual removal while InternalAPI implies disallowed/discouraged usage permanently. At least in my opinion.


fire-light42 (approved):
Good catch and well written fix!

Issue #2244, opened 2025-11-21
Feature Request: Direct YouTube Trailer Link


xx-ZORO-xx:
### Describe your suggested feature

Hello Developers, I hope you're all doing great!
I have a small request: Currently, the trailer quality within the app is low, which forces users to repeatedly open YouTube to watch the trailer in better quality, access comments and to find additional information. Could you please add a dedicated button or enable a long press on the trailer section that will open the YouTube link directly? This will provide a much better viewing experience and make it easier to access comments and extra information. Thank you!

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

Uvdon:
Yeah I agree

xx-ZORO-xx:
> Yeah I agree

Great, thanks for seeing it the same way!

PR #2332, opened 2025-12-14
open in browser btn:original trailer url instead of extractedlink

rockhero1234:
https://github.com/recloudstream/cloudstream/issues/2244


xx-ZORO-xx:
> #2244

Can we please implement this - that clicking the trailer button directly performs a search with the relevant keywords on YouTube app, instead of opening an external browser. I think it's something called (YouTube deep linking)

PR #2333, opened 2025-12-14
Fix scroll issue on bottom dialogs and add drag handle to UI

### LINKED ISSUE #349: Improve scrolling on through Video Source overview prompt
### Steps to reproduce

When we select custom video players like mpv, Web Video Cast, Web browser (VLC & Internal Player are excluded because this issue doesn't affect them), When try to play something there is a Video Source overview prompt appears.

When scroll on through the Video Source overview prompt from top to very bottom is fine, But try to scroll from bottom to top the Video Source overview prompt get easily dismissed.
That's very annoying! 

### Expected behavior

https://user-images.githubusercontent.com/116799295/215964081-80c7bdc6-3d24-4a8c-b33d-854c93e97a10.mp4

### Actual behavior

https://user-images.githubusercontent.com/116799295/215964102-bf35de45-6af6-4c83-9caa-a1683aefafe0.mp4

4.0.0 PRE • b092116

### Android version

Android 11

### Logcat

_No response_

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #2128: Some menus don't allow scrolling up in landscape mode on mobile
### Steps to reproduce

1. Long click an episode while in landscape mode (or open some menus in settings such as "Preferred video player")
2.  Scroll down in the menu to the bottom
3. Try scrolling back up

### Expected behavior

The menu should properly scroll back up

### Actual behavior

Scrolling up closes the bottom dialog.

4.6.1 acbedd1

### Android version

Android 16

### Logcat

```java

```

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #2210: Scroll issue in mirror list
### Steps to reproduce

Go to any tv stream or movie, click download mirror and after loaded scroll down. But when you try to scroll up it closes the whole menu... 

https://files.catbox.moe/i6ecjx.mp4

This probably happens on other menus too

### Expected behavior

Should scroll normally. My recommendation is to add a top header of some sort that says "download mirrors" so when you hold that header and scroll up or down then it closes the menu 

### Actual behavior

But when you try to scroll up it closes the whole menu...

4.6.1 0579a15 

### Android version

Android 15

### Logcat

```java
No logcat needed
```

### Other details

_No response_

### Acknowledgements

Luna712:
Technically, the only fix necessary was setting `isNestedScrollingEnabled = true` on the ListView in order to make it properly scrollable and not dismiss when scrolling. But this also cleans up the layout and adds a drag handler view to make it more visibly explicit.

Fixes #349
Fixes #2128
Fixes #2210

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -66,7 +66,6 @@ android {
         versionName = "4.6.1"
 
         resValue("string", "commit_hash", getGitCommitHash())
-        resValue("bool", "is_prerelease", "false")
 
         manifestPlaceholders["target_sdk_version"] = libs.versions.targetSdk.get()
 
@@ -120,11 +119,9 @@ android {
     productFlavors {
         create("stable") {
             dimension = "state"
-            resValue("bool", "is_prerelease", "false")
         }
         create("prerelease") {
             dimension = "state"
-            resValue("bool", "is_prerelease", "true")
             applicationIdSuffix = ".prerelease"
             if (signingConfigs.names.contains("prerelease")) {
                 signingConfig = signingConfigs.getByName("prerelease")
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -23,7 +23,6 @@ import android.widget.CheckBox
 import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.Toast
-import androidx.activity.OnBackPressedCallback
 import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.IdRes
 import androidx.annotation.MainThread
@@ -107,6 +106,7 @@ import com.lagradost.cloudstream3.ui.SyncWatchType
 import com.lagradost.cloudstream3.ui.WatchType
 import com.lagradost.cloudstream3.ui.account.AccountHelper.showAccountSelectLinear
 import com.lagradost.cloudstream3.ui.download.DOWNLOAD_NAVIGATE_TO
+import com.lagradost.cloudstream3.ui.home.HomeFragment
 import com.lagradost.cloudstream3.ui.home.HomeViewModel
 import com.lagradost.cloudstream3.ui.library.LibraryViewModel
 import com.lagradost.cloudstream3.ui.player.BasicLink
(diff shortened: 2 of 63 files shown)


fire-light42:
isNestedScrollingEnabled? Are you sure this will not devastate performance?

Luna712:
> isNestedScrollingEnabled? Are you sure this will not devastate performance?

@fire-light42 

Doesn't seem to when testing anyway.

Luna712:
I think I have a better fix though

Luna712:
> I checked, and "isNestedScrollingEnabled" does not cause the issue I was worried about.

This method only partially works also though I have another PR I could also do separately if you want that improves both UX and UI further then this does.

Luna712:
I will just commit to this PR and can revert if you'd rather the other method.

fire-light42:
> > I checked, and "isNestedScrollingEnabled" does not cause the issue I was worried about.
> 
> This method only partially works also though I have another PR I could also do separately if you want that improves both UX and UI further then this does.

Nice, I recommend using our baseadapter/nostateadapter, as listview does not recycle views ect.

Luna712:
> > > I checked, and "isNestedScrollingEnabled" does not cause the issue I was worried about.
> > 
> > 
> > This method only partially works also though I have another PR I could also do separately if you want that improves both UX and UI further then this does.
> 
> Nice, I recommend using our baseadapter/nostateadapter, as listview does not recycle views ect.

This could still use more improvement then I did now but since it doesn't use that adapter at a yet I didn't do it here, we could probably make this whole ListView some RecyclerView but wanted to minimize the amount of changes in just this bug fix, but can change it more if you want also. I was also thinking we should create a BottomSheetDialogListView that handles this and use that instead of this `setOnTouchListener` but decided not to for now. Not sure exactly what you want to so though, but can change it if you want.

app/src/main/java/com/lagradost/cloudstream3/utils/SingleSelectionHelper.kt
```diff
@@ -113,8 +115,37 @@ object SingleSelectionHelper {
         val textView = binding.text1
         val applyButton = binding.applyBtt
         val cancelButton = binding.cancelBtt
-        val applyHolder =
-            binding.applyBttHolder
+        val applyHolder = binding.applyBttHolder
+
+        if (isLayout(PHONE or EMULATOR) && (dialog is BottomSheetDialog)) {
```
fire-light42: While this works well, it will hide the last few items if the user does not fullscreen the dialog.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/utils/SingleSelectionHelper.kt
Luna712: I honestly don't know how to fix that. It technically exists on current also just that it auto expands when swiping down which IMO is in itself bad UX as it causes UI displacement when you may not mean for it to.

Luna712: I just set to isNestedScrollingEnabled again, to many accessibility issues like breaking one handed gestures the other way, even though I managed to fix the above bug this way, I realized it broke like one handed gestures, and other things, and more potential for bugs and harder to maintain so let's just let the system handle it I guess.

PR #2342, opened 2025-12-17
Update media3 to 1.9.2

Luna712:
This also requires minSdk 23 (#2078) first.

It also applies all patches for `UpdatedDefaultExtractorsFactory` and `UpdatedMatroskaExtractor` from upstream, but converting from Java to Kotlin there is some potential I did it wrong so may need more reviewing or testing.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -152,6 +152,14 @@ android {
         resValues = true
     }
 
+    packaging {
+        jniLibs {
+            // Enables legacy JNI packaging to reduce APK size (similar to builds before minSdk 23).
+            // Note: This may increase app startup time slightly.
+            useLegacyPackaging = true
+        }
+    }
+
     namespace = "com.lagradost.cloudstream3"
 }
 
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/CustomRecyclerViews.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/CustomRecyclerViews.kt
@@ -3,6 +3,7 @@ package com.lagradost.cloudstream3.ui
 import android.content.Context
 import android.util.AttributeSet
 import android.view.View
+import androidx.core.content.withStyledAttributes
 import androidx.core.view.children
 import androidx.recyclerview.widget.GridLayoutManager
 import androidx.recyclerview.widget.RecyclerView
@@ -154,10 +155,9 @@ class AutofitRecyclerView @JvmOverloads constructor(context: Context, attrs: Att
 
     init {
         if (attrs != null) {
-            val attrsArray = intArrayOf(android.R.attr.columnWidth)
-            val array = context.obtainStyledAttributes(attrs, attrsArray)
-            columnWidth = array.getDimensionPixelSize(0, -1)
-            array.recycle()
+            context.withStyledAttributes(attrs, intArrayOf(android.R.attr.columnWidth)) {
+                columnWidth = getDimensionPixelSize(0, -1)
+            }
         }
 
         layoutManager = manager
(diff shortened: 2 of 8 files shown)


Luna712:
> This fails on the videos we tried to patch with our UpdatedMatroskaExtractor. See sample_three_seekheads.mkv and sample_recursive_seekhead.mkv.

Nice thanks for samples I was wondering how to test that as yeah wasn't sure. Will look.

fire-light42:
> > This fails on the videos we tried to patch with our UpdatedMatroskaExtractor. See sample_three_seekheads.mkv and sample_recursive_seekhead.mkv.
> 
> Nice thanks for samples I was wondering how to test that as yeah wasn't sure. Will look.

It is perhaps better to push our changes upstream and wait for 1.9.1. However the review team at android is really bad, and only exists to bikeshed. So it is a 50/50 if that is even viable.

Luna712:
> This fails on the videos we tried to patch with our UpdatedMatroskaExtractor. See sample_three_seekheads.mkv and sample_recursive_seekhead.mkv.

@fire-light42 

Should be fixed, tested with those files now, some changes in rollback commits weren't applied (I assumed rollback was only rollback and not other random changes also so didn't check to apply those)



> > > This fails on the videos we tried to patch with our UpdatedMatroskaExtractor. See sample_three_seekheads.mkv and sample_recursive_seekhead.mkv.
> > 
> > 
> > Nice thanks for samples I was wondering how to test that as yeah wasn't sure. Will look.
> 
> It is perhaps better to push our changes upstream and wait for 1.9.1. However the review team at android is really bad, and only exists to bikeshed. So it is a 50/50 if that is even viable.

@fire-light42 

Seeing as how a PR already exists for it upstream and no progress on it perhaps we should still do this for now? As I really am not confident it will be merged by 1.9.1.

Luna712:
@fire-light42 just to mention it since you approved this now, #2078 should be merged first, this also bumps minSdk just to make it pass but the other PR also does `useLegacyPackaging` which without causes the APK size to pretty much double.

Luna712:
@fire-light42 seemed to cause ERROR_CODE_FAILED_RUNTIME_CHECK on most videos when using Software Decoding. Changing to HW works all others gives that. Not sure why as it didnt happen when testing initially.

Update found out why, reverting nextlib should in theory fix it. Happens due to subtitle changes in nextlib. So only fatals if using certain subtitle types due to NextTextRenderer not enable legacy Decoding it seems.

Luna712:
https://github.com/anilbeesetti/nextlib/commit/f11ca49b399d0dcad45e54ff8b2bb0a22b93e033 probably caused it so just going back to 1.9.1-0.10.1 should in theory fix it.

Luna712:
@fire-light42 why was this whole PR reverted when I mentioned above the only part that needed reverted? I just didn't have time to do it then.

fire-light42:
> @fire-light42 why was this whole PR reverted when I mentioned above the only part that needed reverted? I just didn't have time to do it then.

The main issue was that a quick fix was needed to prevent people on pre-release from being unable to play anything, and reverting it was the quickest way to do that without a lot of testing to find and fix the core issue. Moreover, I also found some app crashes, however I only now found out by a lot of testing that it was caused by android OS issues. A full revert is the default policy in case of crashes/denial of service.

Luna712:
> > @fire-light42 why was this whole PR reverted when I mentioned above the only part that needed reverted? I just didn't have time to do it then.
> 
> The main issue was that a quick fix was needed to prevent people on pre-release from being unable to play anything, and reverting it was the quickest way to do that without a lot of testing to find and fix the core issue. Moreover, I also found some app crashes, however I only now found out by a lot of testing that it was caused by android OS issues. A full revert is the default policy in case of crashes/denial of service.

Fair enough. Also I only now realize my comment was a little more direct then I intended, it sounded more like criticism than a question. It was intended as an actual question if I was missing something else. I do apologize for that.

PR #2349, opened 2025-12-20
Improving Extractors and also Improved getQualityFromName()

phisher98:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/BuzzServer.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import com.lagradost.api.Log
+import com.lagradost.cloudstream3.SubtitleFile
+import com.lagradost.cloudstream3.app
+import com.lagradost.cloudstream3.utils.ExtractorApi
+import com.lagradost.cloudstream3.utils.ExtractorLink
+import com.lagradost.cloudstream3.utils.getQualityFromName
+import com.lagradost.cloudstream3.utils.newExtractorLink
+
+class BuzzServer : ExtractorApi() {
```
Luna712, line 13: `@Prerelease` please.

phisher98: Added

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -791,15 +792,26 @@ enum class Qualities(var value: Int, val defaultPriority: Int) {
     }
 }
 
+private val QUALITY_REGEX_MAP = listOf(
+    Regex("""\b(4k|2160p?|2160)\b""", RegexOption.IGNORE_CASE) to Qualities.P2160.value,
+    Regex("""\b1440p?|1440\b""", RegexOption.IGNORE_CASE)     to Qualities.P1440.value,
+    Regex("""\b1080p?|1080\b""", RegexOption.IGNORE_CASE)     to Qualities.P1080.value,
+    Regex("""\b720p?|720\b""", RegexOption.IGNORE_CASE)      to Qualities.P720.value,
+    Regex("""\b480p?|480\b""", RegexOption.IGNORE_CASE)      to Qualities.P480.value
```
diogob003, line 800: * `p?` or `p{,1}` or `p{0,1}`: match `p` between 0 and 1 time (p is optional)
```suggestion
    Regex("""\b(4k|2160p?)\b""", RegexOption.IGNORE_CASE) to Qualities.P2160.value,
    Regex("""\b1440p?\b""", RegexOption.IGNORE_CASE)     to Qualities.P1440.value,
    Regex("""\b1080p?\b""", RegexOption.IGNORE_CASE)     to Qualities.P1080.value,
    Regex("""\b720p?\b""", RegexOption.IGNORE_CASE)      to Qualities.P720.value,
    Regex("""\b480p?\b""", RegexOption.IGNORE_CASE)      to Qualities.P480.value
```

diogob003: Is it really necessary to use Regex here? Won't it affect performance, since it will only match those specific words?
* 4k or 2160 or 2160p
* 1440 or 1440p
* 1080 1080p
* 720 720p
* 480 480p

A simple solution, such as the one below, may offer better performance.
```kotlin
when(qualityName.lowercase()) { 
        "4k", "2160", "2160p" -> Qualities.P2160.value
        "1440", "1440p" -> Qualities.P1440.value
        "1080", "1080p" -> Qualities.P1080.value
        "720", "720p" -> Qualities.P720.value
        "480", "480p" -> Qualities.P480.value
        else -> Qualities.Unknown.value
```

fire-light42: Do not forget to trim qualityName too, as we may get `" 480p "` which will not match anything. However I do not quite understand why we need to change this function? What is the problem with the current approach?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -791,15 +792,26 @@ enum class Qualities(var value: Int, val defaultPriority: Int) {
...
+)
+private var lastResolvedQuality: Int = Qualities.Unknown.value
```
fire-light42, line 802: Wow, this is very unsafe, and creates a plethora of race conditions. Can you explain the reason for storing the "last" quality?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
phisher98: I tried that and it was not working for some and giving 400 Quality Qualities.Unknown.value So tried this and it worked

i was not sure about it why its doing it

PR #2356, opened 2025-12-21
Enable `abortOnError` in lint

Luna712:
I have been working on fixing all error-level lint warnings, and now that that is done, we can enable this. We already do this for library now as well. This will help prevent issues like #2320 in the future.


Luna712:
#2273 fixes the last issue with that though so that will need merged first.

Luna712:
Just for the record the only reason this build passed here was that it was missing #2362, however now that #2273 is merged as well the build should fully pass with this, but #2362 should still be merged first and this can be rebased to verify that  before merging into pre release if wanted.

Luna712:
It seems in the time I opened this more error level lints have been introduced again, someone will have to fix that again.

Luna712:
> It seems in the time I opened this more error level lints have been introduced again, someone will have to fix that again.

Now fixed.

Luna712:
This was a 1 line simple PR that had no runtime affect other than ensuring things work and causing me to have to keep fixing things, since the longer it takes more things have to be fixed, I no longer have any interest in this. If someone else wants to try again feel free.

PR #2358, opened 2025-12-21
Run lint in pull requests

Luna712:


Luna712:
I will do this a different way so it always runs if running locally also for consistency

Luna712:
@fire-light42 I *think* this is what you meant in #2362 ? But now thinking about it again wasn't certain.

Issue #2367, opened 2025-12-22
CI actions: Move away from tibdex/github-app-token and marvinpinto/action-automatic-releases


diogob003:
### Describe your suggested feature

The GitHub actions [tibdex/github-app-token](/tibdex/github-app-token) and [marvinpinto/action-automatic-releases](/marvinpinto/action-automatic-releases) are archived and unmaintained.

It's important to start looking for and testing alternatives.

Alternatives to [tibdex/github-app-token](/tibdex/github-app-token)
 - [actions/create-github-app-token](/actions/create-github-app-token) 

Alternatives to [marvinpinto/action-automatic-releases](/marvinpinto/action-automatic-releases)
 - [softprops/action-gh-release](/softprops/action-gh-release)
 - [semantic-release/semantic-release](/semantic-release/semantic-release)

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

Luna712:
Funny enough I was working on this this morning but realized that they arent exactly the same, like for creating releases the release summary is in a different format that didn't look quite right to me so was looking into a way to fix this.

Luna712:
By the way #2370 does the token generation action (if I did it right, hard for me to test that one) I will work on pre release one too but that one is a bit more difficult. semantic-release is a more complicated setup, but the only one that maintains the same, at least mostly, format for release notes, the other one just uses GitHub auto release generation system and doesn't use commit message format for release notes generation so isn't really a good option.

PR #2370, opened 2025-12-23
Migrate token generation to new action

Luna712:
In GitHub settings, the secret `GH_APP_ID` should be deleted, replaced with a var (https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#defining-configuration-variables-for-multiple-workflows) called `APP_CLIENT_ID` which should be the client ID for the GitHub App rather than the current App ID usage. Additionally, the secret `GH_APP_KEY` should be renamed to `APP_PRIVATE_KEY`. I was just following the exact examples on https://github.com/actions/create-github-app-token.

I have tested all the actions. As a result of my testing in my fork, I also made the actions more usable in forks as long as you have the necessary repositories and signing keys. That should also make testing action changes (such as the rest of #2367) easier for me personally and potentially others in the future as well.

Does part of #2367. I plan to do the rest of that (replace the release action) after this one is done. Making it more generic and usable in forks will make that easier as well.


Luna712:
@fire-light42 this is tested on my fork now. But does require some changes to secrets as explained in the PR description.

PR #2381, opened 2025-12-31
Add new mainUrl DoodStream

saimuelbr:
new mainUrl to DoodStream extractor


fire-light42:
Do not forget to add the extractor to the list. Otherwise it will not do anything.

PR #2390, opened 2026-01-03
feat(extractors): add vidnest extractor

Bnyro:
Can be tested on https://filmpalast.to


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/AsianLoad.kt
```diff
@@ -29,15 +32,14 @@ open class AsianLoad : ExtractorApi() {
...
-                } else if (extractedUrl.endsWith(".mp4")) {
+                } else if (extractedUrl.contains(".mp4")) {
                     extractedLinksList.add(
                         newExtractorLink(
                             source = name,
                             name = name,
                             url = extractedUrl,
                         ) {
                             this.referer = url.replace(" ", "%20")
-                            this.quality = getQualityFromName(sourceMatch.groupValues[2])
```
Bnyro, line 40: This probably never worked because the regex only captures one group, so `groupValues[2]` is always a `NullPointerException`

PR #2396, opened 2026-01-05
Add newExtractorLink usage in ByseSX Extractor

saimuelbr:
The current method used in the extractor ends up creating duplicates.

Ex:
20260104_22h42m50s_grim

If we use newExtractorLink in this way, the extractor will only create one option with multiple resolutions.


phisher98:
@saimuelbr Buddy that's what it supposed to do to create multiple quality links from a single m3u8 link

That's not duplicate it's one which contains all qualities and one only 720p link and user can use priority on it to watch on 720p and accordingly

fire-light42:
Yes, @phisher98 is correct. The function is created to split master m3u8 files for easier link selection or download, as one master m3u8 file may contain several playlists. If we simply return the master m3u8 file, then the quality will be unknown and lower down in the link menu. 

However, if you want to make it dedup links when a m3u8 contains a single quality, then modify `M3u8Helper` to return the master m3u8 file if it only contains a single quality, and mark the master m3u8 with that quality.

PR #2398, opened 2026-01-05
refactor(extractors): simplify and combine jwplayer extraction

Bnyro:
I noticed that a lot of extractors are based on the following pattern:
1. Do some magic to obtain a `script` file containing JWPlayer config
2. Extract stream URLs from this JWPlayer config

Previously, all extractors had their own logic for extracting the JWPlayer config, even though the JavaScript data parsed always looks the same. So there's been a lot of duplicated code that behaved inconsistent (i.e. some only supported `m3u8` links and no other stream types like `mp4` and others didn't parse the subtitles).

This logic is now handled by `JWPlayerHelper.kt`.

I've also noticed that some of the files could be merged, e.g. 
- Filemoon
- Filesim
- Streamwish

contain almost the exact same logic and
- Gamovideo
- Hxfile
- Supervideo
- Vtbe
- Vidhide

also do the exact same thing. I suspect that all these providers are based on the exact same code (only some UI changes), but I haven't moved their extractors into the same class yet because I'd like to hear some other opinions first before doing that.

The deleted files are usages of the JWPlayer API on sites that no longer exist, so I removed them.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/helper/JWPlayerHelper.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        val extractedLinks = sourceMatches.flatMap { link ->
+            if (link.file.contains(".m3u8")) {
+                M3u8Helper.generateM3u8(
+                    source = sourceName,
+                    streamUrl = link.file,
+                    referer = referer.orEmpty(),
+                    headers = headers
+                )
+            } else {
```
diogob003: Without using a `try...catch` around `generateM3u8`, an exception can be thrown and exit from the loop without evaluating the remaining links.
```suggestion
        val extractedLinks = sourceMatches.flatMap { link ->
            if (link.file.contains(".m3u8")) {
                try {
                    M3u8Helper.generateM3u8(
                        source = sourceName,
                        streamUrl = link.file,
                        referer = referer.orEmpty(),
                        headers = headers
                    )
                } catch (e: Exception) {
                    Log.d("JW_PLAYER_HELPER", "Error generating M3U8 links: ${e.message}")
                }
            } else {
```

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/helper/JWPlayerHelper.kt
```diff
@@ -0,0 +1,108 @@
...
+                        source = sourceName,
+                        name = sourceName,
+                        url = link.file,
+                    ) {
+                        this.referer = url
+                        this.headers = headers
+                    }
+                )
+            }
+        }
```
diogob003, line 75: It's not bad, but leaves a lot of cases unhandled.
There are several counterexamples (I just can't remember where to find them all 🙃 ) as in:
```js
var links = {
  "hls3": "https://mmmmmmmmmm.qqqqqqqqqqqq.space/#########/hls3/01/00000/ggggggggg_l/master.txt",
  "hls4": "/stream/zzzzzzzzzzzzzzz/hhhhhhhhhhh/123456789/123456/master.m3u8",
  "hls2": "https://mmmmmmmmmm.qqqqqqqqqqqq.com/hls2/01/00000/ggggggggg_l/master.m3u8?t=##################&s=123456"
};
jwplayer("vplayer").setup({
  sources: [{
    file: links.hls4 || links.hls3 || links.hls2,
    type: "hls"
  }],
  image: "https://pppppppp.ppp/p.jpg",
```
`file:` may have a variable or function that gathers links from other parts of the script.
The links may be incomplete, so add the `mainUrl` prefix from the extractor or use `fixUrl()`.

To avoid dealing with the various exceptions, I usually do it like this:
```kotlin
Regex("""[:=]\s*\"([^\"\s]+(\.m3u8|master\.txt)[^\"\s]*)""").findAll(unpackedScript).forEach { match ->
    val link = match.groupValues[1]
    callback.invoke(
        newExtractorLink(
            source = this.name,
            name   = this.name,
            url    = fixUrl(link)
        ) {
            this.referer = referer ?: mainUrl
        }
    )
}
```
That is, after `:` or `=` get all that's delimited by `"` and has `.m3u8` or `master.txt` inside.

Bnyro: Very good catch, thank you for the review @diogob003 !

One example would be https://github.com/recloudstream/cloudstream/blob/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/VidHidePro.kt#L83, I probably should have remembered that edge case because I wrote the patch over there myself :)


We're now trying to extract stream links by first
- trying to parse the JWPlayer config object and searching for `the file:<url>` pattern
- if that doesn't find anything, we're searching for `hls` streams as you suggested

The problem why we can't drop the first case is that some providers do provide other formats than `HLS`, e.g. I've already seen some `.mp4` streams. So in this case, using only the second method wouldn't work.

okibcn:
I am reworking some extractors. I recently fixed Vidmoly that would benefit from this utility as it fits in the same class of jwplayer config string. 

On the other hand you mention Filemoon, but it is no longer working since now the load of the iframe page is way more complicated and behind several scripts.

Bnyro:
> On the other hand you mention Filemoon, but it is no longer working since now the load of the iframe page is way more complicated and behind several scripts.
>
Yes, true. I only noticed that Streamwish, Filesim and Filemoon use almost the same code, so it's possible that they're all the same provider (i.e. the same company hosting it), that note is not meant to be related to JwPlayer.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/helper/JWPlayerHelper.kt
```diff
@@ -0,0 +1,153 @@
...
+        // jwplayer("vplayer").setup({
+        //  sources: [{
+        //    file: links.hls4 || links.hls3 || links.hls2,
+        //    type: "hls"
+        //  }],
+        if (extractedLinks.isEmpty()) {
+            m3u8Regex.findAll(script).toList().map { match ->
+                val link = match.groupValues[1]
+
+                newExtractorLink(
```
fire-light42, line 92: These are unused and will not be appended to extractedLinks

fire-light42 (changes requested):
Sorry for the late review! 

I will try to get this merged very soon. The code looks good, but it seems like you forgot to add some links.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/helper/JWPlayerHelper.kt
```diff
@@ -0,0 +1,153 @@
...
+import com.lagradost.api.Log
+import com.lagradost.cloudstream3.SubtitleFile
+import com.lagradost.cloudstream3.newSubtitleFile
+import com.lagradost.cloudstream3.utils.AppUtils.tryParseJson
+import com.lagradost.cloudstream3.utils.ExtractorLink
+import com.lagradost.cloudstream3.utils.M3u8Helper
+import com.lagradost.cloudstream3.utils.newExtractorLink
+import kotlin.collections.orEmpty
+
+object JwPlayerHelper {
```
fire-light42, line 15: Don't forget a `@Prerelease` annotation in case extension developers want to use this

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/helper/JWPlayerHelper.kt
Bnyro: That's been an overseight from my side, thanks!

Bnyro:
> Sorry for the late review!
> 
No worries, happy we're finally getting close to merging this :)

fire-light42 (approved):
Looks good and seems to work well. 

I was thinking we could potentially use `getAndUnpack()` inside extractStreamLinks, because packed JS seems to be extremely prevalent. However, that is outside the scope of this pull request. 

Issue #2311, opened 2025-12-09
Cast icon dissappears


Luna712:
### Steps to reproduce

I've noticed this issue happening randomly for a very long time, but it took me forever to figure out a reliable reproduction in order to make this issue, but finally did.
1. First make sure the app is stopped (use force stop)
2. Open a results page on phone (the cast icon should appear)
3. Go back to home page
4. Go to your device home page
5. Click the app icon again (does not always happen if you reopen the app from recents only if you click the icon again)
6. Go back to a results page again

### Expected behavior

Cast icon should remain there.

### Actual behavior

Cast icon is gone now. It only appears again if app is completely restarted, clicking back out of the app, or even swiping it away in recents doesn't always fix it, sometimes only force stop and reopen fixes it.

### Cloudstream version and commit hash

4.6.1 a46b0ac

### Android version

Android 16

### Logcat

```java

```

### Acknowledgements

- [x] I am sure my issue is related to the app and **NOT some extension**.
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [x] I will fill out all of the requested information in this form.

gittinrep:
It seems that backgrounding the app causes an issue with state tracking for the cast context, which causes this boolean to be true:

https://github.com/recloudstream/cloudstream/blob/f84414dbf64b8cfc587ce2f1831c79976a0c75d2/app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragmentPhone.kt#L551

And as such, hiding the button. Essentially, it seems that backgrounding the process makes it so that the cast context no longer reliably scans for available devices when the process is resumed. (The highlighted line also implies that the button will be hidden if no casting devices are found on a network under regular circumstances?)

Luna712:
This was seemingly fixed by the above commit

PR #2412, opened 2026-01-11
Removes check for available devices before displaying chromecast button

gittinrep:
Once the button is selected, the chrome cast service will search for devices and notify the user if none have been found.

Addresses the disappearing button in issue #2311


fire-light42:
🤔 So chromecast does not rescan? Weird, but I guess that works.

gittinrep:
Seems to be a similar issue to a thread I found here: https://github.com/androidx/media/issues/2925

The explanation seems reasonable ¯\\_(ツ)_/¯

fire-light42 (approved):
Everything looks good, and this will be merged after testing.

PR #2416, opened 2026-01-12
Add extra brightness feature

Pawloland:
I was inspired by the extra loudness feature activated with a gesture and wanted something similar, but for video brightness/ gamma correction. It is useful when watching in direct sunlight on the not bright enough screen or when watching very dark movies in general. The color math heavy lifting is done with the code originally from https://github.com/MasayukiSuda/GPUVideo-android library that is MIT licensed. I forked it https://github.com/Pawloland/GPUVideo-android to make it up to date enough to run it's demo sample app on modern Android (in my fork there is only the movie player filter functionality, I removed camera and media file editing option, because I didn't feel like fixing scoped storage and camera API, and the thing that I needed in cloudstream was just the exoplayer part, which I upgraded to Media3).  I copied the source code from my fork into cloudstream codbase with minimal changes, mainly paths in packages and imports, plus relying on PlayerResize enum for scaling types used in cloudstream. There are a lot more effects inside, which I didn't remove. Maybe someone would be intrigued by this and build nice UI for changing contrast, exposure, hue, saturation, vibrance or white balance to name a few.  
Bellow is a demo of it in action when playing [Big Buck Bunny video file](https://download.blender.org/demo/movies/BBB/bbb_sunflower_2160p_60fps_normal.mp4.zip). From my quick empirical tests it behaves as expected on rotations, changes of the scaling to fit, fill and zoom, and it seams to works fine on TV layout. 

https://github.com/user-attachments/assets/1e6df8a2-28e6-45f1-9719-2cf978b6e4fc

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -116,6 +116,7 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
     protected var playerBinding: PlayerCustomLayoutBinding? = null
     private var gpuPlayerView: GPUPlayerView? = null
     private var gpuBrightnessFilter: GlBrightnessFilter? = null
+    private var hasBrightnessBoostError: Boolean = false
 
     private var durationMode: Boolean by UserPreferenceDelegate("duration_mode", false)
 
@@ -210,10 +211,11 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         // Create GPUPlayerView dynamically and attach it to the PlayerView's content frame
         safe {
             val pv = root.findViewById<androidx.media3.ui.PlayerView>(R.id.player_view)
-            val contentId = resources.getIdentifier("exo_content_frame", "id", requireContext().packageName)
+            val packageName = context?.packageName ?: return@safe
+            val contentId = resources.getIdentifier("exo_content_frame", "id", packageName)
             val contentFrame = pv?.findViewById<android.view.ViewGroup>(contentId)
             if (contentFrame != null) {
-                val gpu = GPUPlayerView(requireContext())
+                val gpu = GPUPlayerView(context)
                 val lp = android.widget.FrameLayout.LayoutParams(
                     android.view.ViewGroup.LayoutParams.MATCH_PARENT,
                     android.view.ViewGroup.LayoutParams.MATCH_PARENT
@@ -237,10 +239,6 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         if (player is ExoPlayer) {
             // attach GL renderer filter if available
             gpuPlayerView?.setExoPlayer(player)
-            if (gpuBrightnessFilter == null) {
-                gpuBrightnessFilter = GlBrightnessFilter()
-                gpuPlayerView?.setGlFilter(gpuBrightnessFilter)
-            }
         }
     }
 
@@ -248,6 +246,8 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         // Clean up dynamic GPUPlayerView if created
         safe {
             gpuPlayerView?.onPause()
+            gpuPlayerView?.setGlFilter(null)
+            gpuBrightnessFilter = null
             val parent = gpuPlayerView?.parent as? android.view.ViewGroup
             parent?.removeView(gpuPlayerView)
         }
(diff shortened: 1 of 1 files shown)


brahmkshatriya:
I would suggest removing all the unused classes, for example the filter classes except for GLBrightnessFilter,
Edit: Didnt read your description, but still better to remove unused stuff

Pawloland:
I removed all of the code taken from GPUV library and used a jitpack dependency instead. Jitpack dependency is built from my fork of GPUV. Before merging someone should fork my library fork under recloudstream account and use it instead of mine `com.github.Pawloland:GPUVideo-android` in `libs.versions.toml`. It is the best way, because MIT license of GPUV is kept clearly separate from cloudstream GPL3, under different repo and git commit history isn't lost like it was before.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -199,14 +205,74 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
         val root = super.onCreateView(inflater, container, savedInstanceState) ?: return null
         playerBinding = PlayerCustomLayoutBinding.bind(root.findViewById(R.id.player_holder))
+
+        // Create GPUPlayerView dynamically and attach it to the PlayerView's content frame
+        try {
+            val pv = root.findViewById<androidx.media3.ui.PlayerView>(R.id.player_view)
+            val contentId = resources.getIdentifier("exo_content_frame", "id", requireContext().packageName)
+            val contentFrame = pv?.findViewById<android.view.ViewGroup>(contentId)
+            if (contentFrame != null) {
+                val gpu = GPUPlayerView(requireContext())
```
fire-light42: Do not use requireContext anywhere, it will cause a crash.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -199,14 +205,74 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+            // ignore
+        }
+        gpuPlayerView = null
         playerBinding = null
         super.onDestroyView()
     }
 
+    override fun resize(resize: PlayerResize, showToast: Boolean) {
+        super.resize(resize, showToast)
+        try {
```
fire-light42: I would do a `safe { ... }` to both log and catch throwables.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -199,14 +205,74 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+            if (gpuBrightnessFilter == null) {
+                gpuBrightnessFilter = GlBrightnessFilter()
+                gpuPlayerView?.setGlFilter(gpuBrightnessFilter)
+            }
+        }
+    }
+
     override fun onDestroyView() {
+        // Clean up dynamic GPUPlayerView if created
+        try {
```
fire-light42: `safe {...}`

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -199,14 +205,74 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
     ): View? {
         val root = super.onCreateView(inflater, container, savedInstanceState) ?: return null
         playerBinding = PlayerCustomLayoutBinding.bind(root.findViewById(R.id.player_holder))
+
+        // Create GPUPlayerView dynamically and attach it to the PlayerView's content frame
+        try {
```
fire-light42: `safe {...}`

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -109,6 +113,8 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
     protected open var lockRotation = true
     protected open var isFullScreenPlayer = true
     protected var playerBinding: PlayerCustomLayoutBinding? = null
+    var gpuPlayerView: GPUPlayerView? = null
```
fire-light42: private?

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1374,27 +1465,47 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
-                                    currentRequestedBrightness =
-                                        min(
-                                            1.0f,
-                                            max(currentRequestedBrightness + verticalAddition, 0.0f)
-                                        )
+                                    val nextBrightness =
+                                        (currentRequestedBrightness + verticalAddition).coerceIn(0.0f, if (isBrightnessLocked) 1.0f else 2.0f)
+                                    // Log.e("Brightness", "Current: $currentRequestedBrightness, Next: $nextBrightness")
+                                    // show toast
+                                    if (nextBrightness == 1.0f && isBrightnessLocked && !hasShownBrightnessToast) {
```
fire-light42: Please do not do any total equality checks on floats. Use < or <= instead. In this case I would do it like the volume check `nextRequestedVolume > 1.0 && isCurrentVolumeLocked && !hasShownVolumeToast`.

```kotlin
val nextRequestedBrightness = currentRequestedBrightness + verticalAddition
if(nextRequestedBrightness > 1.0 && isBrightnessLocked && !hasShownBrightnessToast) { ...
```

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1374,27 +1465,47 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+                                        showToast(R.string.slide_up_again_to_exceed_brightness_100)
+                                        hasShownBrightnessToast = true
+                                    } else if (currentRequestedBrightness == 1.0f && nextBrightness > 1.0f && !hasShownBrightnessToast) {
+                                        showToast(R.string.brightness_exceeded_100)
```
fire-light42: This behavior is different from the sliding of volume. I would try to match the behavior of volume and only show a toast on `slide_up_again_to_exceed_brightness_100`.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1074,7 +1152,12 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         } else {
             try {
                 val lp = activity?.window?.attributes
-                lp?.screenBrightness = brightness
+                // use 0.004f instead of 0, because on some devices setting too small value
+                // causes system to override it and in turn system makes the screen apply system brightness level instead
+                // which can be too bright, and it is very hard to fine tune very low brightness, because of it.
+                // Without this clamp, it can jump from almost 0% to 100% brightness when this threshold is crossed.
+                lp?.screenBrightness = brightness.coerceIn(0.004f, 1.0f)
```
fire-light42, line 1156: Do you have any documentation of this bug or have experienced this? I have not seen anyone reporting this bug, nor anything in the docs concerning this.

Pawloland: I actually have one such a device. I don't know why it happens, but it does and this fixed it. I was thinking about creating a separate setting for it, but clamping it barely above 0 in my opinion does not make any visible difference to setting it to true 0, so such a setting would be not really usefull for most people and such a clap would not be noticable either, so it seamed like a good default.

Pawloland:
I didn't notice any stutter, lags or device overheat when using the filter. I now checked with android studio profiler and also didn't notice any significant activity, but I changed the code regardless to apply filter only when the brightness exceeds `1.0f`.

fire-light42 (approved):

LagradOst:
This feature is now removed as it:
1. Forced HDR content to SDR
2. Caused lag

Pawloland:
Hmm, I didn't experience that. Would making a setting that can disable the feature globally be enough of a fix? I didn't go down in library code enough to make it HDR compatible. It would allow using the feature when it doesn't cause harm. Also does it mess with HDR or lags for you, even when 100% brightness isn't exceeded? It should not apply the filter then, so I guess no lag or HDR should be affected then.

LagradOst:
> Hmm, I didn't experience that. Would making a setting that can disable the feature globally be enough of a fix? I didn't go down in library code enough to make it HDR compatible. It would allow using the feature when it doesn't cause harm. Also does it mess with HDR or lags for you, even when 100% brightness isn't exceeded? It should not apply the filter then, so I guess no lag or HDR should be affected then.

@Pawloland 

I was also surprised that it would cause this issue. But from my testing, even if no filter is applied, it still causes these issues. The main issue here I think is that when using the GPUPlayerView from this lib it forces it to render to a GLSurfaceView. I very much think that GLSurfaceView implementation forces it to SDR (https://developer.android.com/media/grow/hdr-playback). 

And by:
```
Specifying the android.view.Surface
By default GLSurfaceView will create a PixelFormat.RGB_888 format surface. If a translucent surface is required, call getHolder().setFormat(PixelFormat.TRANSLUCENT). The exact format of a TRANSLUCENT surface is device dependent, but it will be a 32-bit-per-pixel surface with 8 bits per component.
```

If I read this correctly, this forces it into another format, which I think is causing both lag and SDR even if no filter is applied. Additionally I tried switching the surface mid playback and it resulted in very weird artifacts and visual bugs, so I do not think it is possible (or easily possible) to switch this on mid stream.

PR #2409, opened 2026-01-10
Enhance Hero Card with Dynamic Rating Badge and Larger Layout

Jaidev1805:
image

---
- Added an IMDb rating badge to highlight key metadata at a glance.
- Increased the hero card size to improve visual prominence and readability on large screens.

PR #2420, opened 2026-01-16
Minor fix for HomeHead BG poster dull #2409

phisher98:
image


Jaidev1805:
The texts are unreadable if the background image is too bright or have light colors

phisher98:
Make the text readable without dimming the posters; the current effect makes them look unattractive.

PR #2428, opened 2026-01-19
Android auto

j4ckp0t85:
Hello.

It started as a sort of joke/experiment, it's ending the same way, so it's up to you if you want to cosider this application extension.

I've clearly reported some statements on README file.

Keypoints:

- this is not a full working implementation, it may contain some bugs. i'm releasing it so other devs / you owners could keep improving it

- features: search (typing or vocal), change provider, see favorites (heart), see downloads, see history, see movie / tv episode infos (if available), pick a network source to play, play/pause stream, shift a running stream +-30s

- in order to work, it declares itself as a navigation app. this is a fully native carapp library implementation, with all relative Google design patterns and restrictions (size of images, number of buttons, number of screen navigations allowed)

- both from legal and security aspects, it won't work well if the system recognize you're moving. please, do not use while driving

- since this is a navigation app, it's based on ExoPlayer. so it should takes advantage (if available) of hardware capabilities to play contents

- again, since this is a navigation app, it's not possible to apply a custom overlay (like cursor timeshift bar, subtitles engine)

- this is a natural extension of the mobile app, that still acts as the master actor for the configurations part. it inherits the language and should display localized messages. to gracefully accepts the connections, i had to move the initialization of the request client to be independent from the mobile app

- since it's an unsigned app, it will be visible on android dhu emulator, using a device like AAWIRELESS or making an internal testing app deploy on Google Play Developer Console

- the version number is increased because i did some deployment tests

- there is an "about me" section with a sample asset video, this is just a personal test experiment i did to test the player

**Final statements**

I'm not responsible of improper usage of this application.

The application could stop working as of Google changes.

I'm not going to provide any kind of future support on this, i'm releasing it to you to continue (eventually) to improve it

Fun fact: mostly this was possible thanks to the help of Nano Banana Pro :)

Cheers


fire-light42:
Extremely impressive pull request! If it works well enough I would not oppose merging this. 
However, due to the size of the pull request it may take some time to review.

PR #2432, opened 2026-01-20
Initial AI Policy

fire-light42:
What do you think about this new policy regarding AI? 

@Luna712 @phisher98 @KingLucius @diogob003 @LagradOst @rockhero1234 @Kraptor123 
Other people are of course also welcome to leave feedback ❤️

### DIFF
--- a/AI-POLICY.md
+++ b/AI-POLICY.md
@@ -1,0 +1,11 @@
+# AI Policy
+
+AI is a great tool. However, we want you to follow these rules regarding usage of AI in order to ensure the quality of both code and discussions.
+
+1. Always state any AI usage in pull requests and issues. 
+
+2. Always test code before making a pull request. We do not want to test your AI generated code. 
+
+3. Listen to humans over computers. Contributors to CloudStream know this codebase better than an AI. 
+
+4. You should be able to explain and fix any code you submit. We do in-depth reviews and will reject low effort contributions.


Kraptor123:
I think it's great to have a policy about it. AI is just a tool; use it wisely, don't trust it blindly, and know what it does before opening PRs. If you don't understand what the code does, don't share it. At least that's what I'm trying to do.

phisher98:
This is a sensible policy. Code must be properly tested before opening a pull request, especially when AI assistance is used.

rockhero1234:
Must needed policy as I am seeing some ai generated code lately. AI is not bad as long as you know what that is doing.

j4ckp0t85:
AI, like humans, both have a same principle: for best results both parts must prompt and understand clearly and correctly in depth what's written behind every single word.

Humans have more "neural" connections, let's say from a top-down complete perspective, AI owns the code (if using the proper model).

It's their natural language, they know the weakness and strong parts.
And sometimes they are too much genuine giving solutions ;)

Otherwise, when there's a misunderstand, things are going wrong

diogob003:
I agree, AI is just a tools, used it wisely. I used often to understand HOW some code works, or alternatives ways to write little chunks of code.

I would like to include to the policy **Keep it simple and small**. Do not let AI mess up hundreds of files to implement minor things.

fire-light42:
> I would like to include to the policy **Keep it simple and small**. Do not let AI mess up hundreds of files to implement minor things.

That is the general policy for all pull requests.

Issue #2031, opened 2025-10-18
Add Kitsu if possible besides MAL and AL.


hellowhatis22-lang:
### Describe your suggested feature

If possible add Kitsu besides MAL and AL 

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

fire-light42:
Can you please specify the exact website.

hellowhatis22-lang:
https://kitsu.app/explore/anime

hellowhatis22-lang:
> Can you please specify the exact website.

https://kitsu.app/explore/anime

PiterWeb:
I want to work on this. 

I have seen the source code of the App and looks like [it has already some code](https://github.com/recloudstream/cloudstream/blob/5c0f7159732fb00aefab21b09e729bca3afb8dd0/app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/KitsuApi.kt) for Kitsu but doesn't provide a way to auth and no UI in the sync providers section.

PiterWeb:
This is what I have done today

![IMG_20260107_153542.jpg](https://github.com/user-attachments/assets/a87e1b7c-5a32-46dc-beab-0209f1bfa494)

![IMG_20260107_153517.jpg](https://github.com/user-attachments/assets/5cdb4e0f-49a2-4de4-a9d0-82927a23f2a6)

PiterWeb:
User login now works
![IMG_20260107_205350.jpg](https://github.com/user-attachments/assets/bc45c02a-811b-4dd0-af71-b0987d159e24)

hellowhatis22-lang:
Great work. You should do pull request

hellowhatis22-lang:
Want kitsu in next update. I am happy for the kitsu.

PiterWeb:
Thank you. Yeah I will do a pull request, I want to finish all (if posible) the kitsu features first
> Great work. You should do pull request

PiterWeb:
The kitsu tab now works

![IMG_20260108_195430.jpg](https://github.com/user-attachments/assets/3331f507-a5b5-4d1e-a25d-5403499e0e56)

hellowhatis22-lang:
Waiting for this updates😊

PiterWeb:
[...]

PR #2440, opened 2026-01-21
Kitsu added as sync provider

PiterWeb:
This pull request is trying to resolve the issue #2031.
In the issue I posted screenshots on what it looks like implemented

Implemented functionality:

- User login with kitsu credentials
- Kitsu library
- Update anime sync in left-side sync panel


app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/KitsuApi.kt
```diff
@@ -1,8 +1,631 @@
...
+                it.id,
+                "$mainUrl/anime/${it.id}/",
+                attributes.posterImage?.large ?: attributes.posterImage?.medium
+            )
+        }
+    }
+
+    override suspend fun load(auth : AuthData?, id: String): SyncResult? {
+        val auth = auth?.token?.accessToken ?: return null
+        id.toIntOrNull() ?: return null
```
fire-light42: Did you mean to write `val id = id.toIntOrNull() ?: return null` because this is confusing to read even if its only purpose is to return null on non-int ids.

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/KitsuApi.kt
```diff
@@ -1,8 +1,631 @@
...
+        val limit = 500
+        val initUrl = "$apiUrl/library-entries?filter[userId]=$userId&filter[kind]=anime&include=anime&page[limit]=$limit&page[offset]=0&fields[anime]=${animeSelectedFields.joinToString(",")}&fields[libraryEntries]=${libraryEntriesSelectedFields.joinToString(",")}"
+
+        val urlChannel = Channel<String>(1) // Capacity 1 to not block thread on first send
+        urlChannel.send(initUrl)
+
+        val fullList = mutableListOf<KitsuNode>()
+        val listMutex = Mutex(false)
+
+        coroutineScope {
```
fire-light42: Why? From the looks of it a simple while loop should work fine?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -1878,6 +1904,10 @@ interface LoadResponse {
             this.addSimklId(SimklSyncServices.Mal, id.toString())
         }
 
+        fun LoadResponse.addKitsuId(id: Int?) {
```
fire-light42, line 1888: Add a prerelease annotation.

fire-light42 (changes requested):
Very good code, I will leave a more detailed review after I have tested it.

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/KitsuApi.kt
PiterWeb: I write it like that to return null if the id failed on converting to int. Now that I look at it, it is confusing hahaha. I can change it to 

```kt
if (id.toIntOrNull() == null) {
 return null
}
```
or not doing the comprobation. In fact I think this is the best option because Kitsu uses numbers inside a string so in a future they might introduce other strings that are not numeric values

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/KitsuApi.kt
PiterWeb: Yeah, right now it can be a normal loop, the initial reason of that was because I misunderstood the way the Kitsu api worked and to achieve a decent result in time I need a lot of concurrency.  

I will modify to use a simple loop

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
PiterWeb: Do you mean a comment in top saying "Prerelease" ?

fire-light42: No, we use the annotation `@Prerelease`

app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/KitsuApi.kt
fire-light42: Yes, do that as it becomes clear what exactly you are trying to do. Otherwise one might think that you can refactor the code by removing your current statement.

PiterWeb:
In the way of doing the changes requested I have also corrected the updateStatus functionality when None is selected in the status to delete that anime from the library

PR #2445, opened 2026-01-24
Fix audio track selection when same lang/id is used in 2+ audio tracks. Add codec and channel configuration in track selection

okibcn:
### Summary
Successfully resolved two critical issues in the Android Cloudstream app's audio track selection system:

1. Audio track dialog always checked the first track in a given language instead of the currently playing track
2. Missing detailed information about audio tracks (codec and channel configuration)

### Issues Resolved

**Issue 1: Incorrect Audio Track Selection**
Problem: The audio track dialog always marked the first audio track as selected, regardless of which track was actually playing.

Root Cause: There wasn't any variable storing the current track index. So, there wasn't any way to know the current selected track. The only way was to select the pair language/id, but that doesn't work when 2 or more tracks use the same language and id.

Solution: Created getCurrentAudioTrackIndex() in CS3IPlayer.kt to query ExoPlayer's current selection when no stored index exists. This is used in CS3IPlayer's `setPreferredAudioTrack(trackLanguage: String?, trackIndex: Int?)` and GeneratorPlayer's `showTracksDialogue()`. Also `setPreferredAudioTrack` no longer expects the id as it is useless. Now it uses the desired deterministic audio track index. 

**Issue 2: Missing Audio Track Information**
Problem: Audio tracks displayed only basic language information without codec or channel details.

Root Cause: The AudioTrack data class only contained id, label, and language fields - no technical specifications.

Solution: Created a public method `getAudioFormats(): List<Format>?` in CS3IPlayer class to access ExoPlayer's Format objects, which contain complete technical details.

### 1. In CS3IPlayer.kt
​
`getCurrentAudioTrackIndex()` now automatically detects the current track from ExoPlayer when there's no stored index

`getAudioFormats()` publicly exposes audio formats so other classes can access detailed information

`setPreferredAudioTrack()` now receives trackIndex (Int) instead of id (String), simplifying the logic

### 2. In GeneratorPlayer.kt
​
The audio dialog now displays very comprehensive information:

```kotlin
"[$index] $language $codec $channels"
```
With user-friendly labels:

**Channels**: "mono", "stereo", "5.1", "7.1" (instead of raw numbers)

**Codec:** clean extracts from MIME types

**Language:** translated to full language name

### 3. In IPlayer.kt
​
Added `import androidx.media3.common.Format`

New interface function `getAudioFormats(): List<Format>?` that CS3IPlayer must implement

### Result
Users can now see clear and useful information about each available audio track, such as:
```
[0] English aac stereo
[1] Spanish ac3 5.1
[2] Japanese opus stereo
```
And the dialog correctly marks the track that is currently playing. Excellent work solving these issues!

A debug version is available as an artifact [here](https://github.com/okibcn/cloudstream/actions/runs/21308005175/artifacts/5241900116).

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -15,7 +15,6 @@ import android.widget.FrameLayout
 import androidx.annotation.MainThread
 import androidx.annotation.OptIn
 import androidx.appcompat.app.AlertDialog
-import androidx.core.net.toUri
 import androidx.media3.common.C.TIME_UNSET
 import androidx.media3.common.C.TRACK_TYPE_AUDIO
 import androidx.media3.common.C.TRACK_TYPE_TEXT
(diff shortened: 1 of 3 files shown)


okibcn:
This PR removes the additional SW decoding options introduced earlier. 

This PR will be closed and resubmitted without removing the previous additions.

Issue #2427, opened 2026-01-19
Unable to select audio tracks when multiple tracks have the same Track Id **[Solution inside]**


okibcn:
### Steps to reproduce

1. Install repo ` https://raw.githubusercontent.com/Sushan64/NetMirror-Extension/refs/heads/builds/Netflix.json`
2. Download Netmirror plugin
3. Open PrimeVideo Plugin from the list of available plugins
4. Search in the plugin for any blockbuster such as `fallout`
5. Play any episode
6. In the player tap on the change Audio Track button at the bottom
7. Select the second Spanish (or second French) track and apply changes
8. Tap Again in the tap on the change Audio Track button
ERROR: Cloudstream has not selected the right track, but the first one since there is more than one track with the same ID. Having more than one track with the same ID is common in Spanish and French content.

<img width="1487" height="669" alt="Image" src="https://github.com/user-attachments/assets/30d6ee11-0788-44b1-8554-1f0b07f809b2" />

### Expected behavior

Cloudstream should provide exoplayer with the right audio track and the second audio track should be selected and played.

### Actual behavior

Cloudstream is not playing the selected track

### Cloudstream version and commit hash

4.6.1-PRE build ee1009a

### Android version

Android 14

### Logcat

```java

```

### Other details

**ROOT CAUSE ANALYSIS:**
`override fun setPreferredAudioTrack(trackLanguage: String?, id: String?)` uses the track ID for selecting the audio track. But it is common to see several tracks with the same Id. By using the Id it is impossible to use alternate tracks sharing the same Id. Only unique value per track should be used, and the Track Index seems to be the only value capable of ensure success.

**PROPOSED SOLUTION:**
The function above in file `CS3IPlayer.kt` should receive the desired audio track index which is unique per track in a stream and not the Id.

This is the proposed function:
```kotlin
override fun setPreferredAudioTrack(trackLanguage: String?, trackNum: Int?) {
    preferredAudioTrackLanguage = trackLanguage

    val player = exoPlayer ?: return
    val tracks = player.currentTracks

    // Find the audio group
    val audioGroup = tracks.groups
        .firstOrNull { it.type == TRACK_TYPE_AUDIO }

    // If a track number is provided, try to override by index
    if (audioGroup != null && trackNum != null) {
        if (trackNum in 0 until audioGroup.length) {

            val override = TrackSelectionOverride(
                audioGroup.mediaTrackGroup,
                listOf(trackNum)   // <-- select by track number
            )

[...]

PR #2447, opened 2026-01-24
Fix track selection when same lang/id is used in 2+ audio tracks. Codec & channel configuration in track selection. Solves issue #2427

okibcn:
# Pull Request Report: Improve Audio Track Selection without Reflection or Global Variables

## Overview

This PR refactors the audio track selection mechanism to avoid using reflection and eliminates the global `currentAudioTrackIndex` variable. The new approach uses a composite ID format (`formatId:formatIndex`) to uniquely identify audio formats within ExoPlayer track groups. 

The original motivation is to solve issue #2427 

## Motivation

1. **Avoid Reflection**: The previous implementation relied on reflection to access ExoPlayer's internal `audioFormat` property, which is fragile and can break with library updates
2. **Remove Global State**: Eliminated the `currentAudioTrackIndex` global variable that was used to track the currently playing audio track
3. **Improve Reliability**: Use ExoPlayer's public API consistently for better maintainability and stability

### 1. IPlayer.kt

**Changes**: Extended `AudioTrack` data class

**Before**:

```kotlin
data class AudioTrack(
    override val id: String?,
    override val label: String?,
    override val language: String?,
    override val sampleMimeType: String?,
) : Track
```

**After**:

```kotlin
data class AudioTrack(
    override val id: String?,
    override val label: String?,
    override val language: String?,
    override val sampleMimeType: String?,
    val channelCount: Int?,
    val formatIndex: Int?,
) : Track
```

**Impact**: Added `channelCount` and `formatIndex` fields to `AudioTrack` to support richer audio track information and enable unique identification.

***

### 2. CS3IPlayer.kt

**Major Changes**:

- Refactored `setPreferredAudioTrack()` method
- Updated `Format.toAudioTrack()` extension
- Modified `getVideoTracks()` to include format indices and detect current audio track

#### Audio Track Conversion \& Detection

**Before**:

```kotlin
private fun Format.toAudioTrack(): AudioTrack {
    return AudioTrack(
        this.id?.stripTrackId(),
        this.label,
        this.language,
        this.sampleMimeType
    )
}

override fun getVideoTracks(): CurrentTracks {
    val allTracks = exoPlayer?.currentTracks?.groups ?: emptyList()
    val audioTracks = allTracks.filter { it.type == TRACK_TYPE_AUDIO }.getFormats()
        .map { it.first.toAudioTrack() }
    
    return CurrentTracks(
        exoPlayer?.videoFormat?.toVideoTrack(),
        exoPlayer?.audioFormat?.toAudioTrack(),  // Uses reflection-based audioFormat
        currentTextTracks,
        videoTracks,
        audioTracks,
        textTracks
    )
}
```

**After**:

```kotlin
private fun Format.toAudioTrack(formatIndex: Int?): AudioTrack {
    return AudioTrack(
        this.id,
        this.label,
        this.language,
        this.sampleMimeType,
        this.channelCount,
        formatIndex ?: 0,
    )
}

override fun getVideoTracks(): CurrentTracks {
    val allTrackGroups = exoPlayer?.currentTracks?.groups ?: emptyList()
    
    var currentAudioTrack: AudioTrack? = null
    val audioTracks = allTrackGroups.filter { it.type == TRACK_TYPE_AUDIO }
        .flatMap { group ->
            (0 until group.mediaTrackGroup.length).map { formatIndex ->
                val format = group.mediaTrackGroup.getFormat(formatIndex)
                val audioTrack = format.toAudioTrack(formatIndex)
                
                // Detect currently selected track without reflection
                if (group.isTrackSelected(formatIndex)) {
                    currentAudioTrack = audioTrack
                }
                
                audioTrack
            }
        }
    
    return CurrentTracks(
        exoPlayer?.videoFormat?.toVideoTrack(),
        currentAudioTrack,  // Now detected without reflection
        currentTextTracks,
        videoTracks,
        audioTracks,
        textTracks
    )
}
```

**Key Improvements**:

- Added `formatIndex` parameter to `toAudioTrack()` with channel count
- `getVideoTracks()` now iterates through all formats and uses `group.isTrackSelected(formatIndex)` to detect the current track
- Eliminates reflection-based `exoPlayer?.audioFormat` access
- Provides complete track list with format indices for each audio track

#### Set Preferred Audio Track

**Before**:

```kotlin
override fun setPreferredAudioTrack(trackLanguage: String?, id: String?) {
    preferredAudioTrackLanguage = trackLanguage
    if (id != null) {
        val audioTrack =
            exoPlayer?.currentTracks?.groups?.filter { it.type == TRACK_TYPE_AUDIO }
                ?.getTrack(id)  // Uses helper that finds track by stripped ID
        if (audioTrack != null) {
            exoPlayer?.trackSelectionParameters = exoPlayer?.trackSelectionParameters
                ?.buildUpon()
                ?.setOverrideForType(
                    TrackSelectionOverride(audioTrack.first, audioTrack.second)
                )
                ?.build() ?: return
            return
        }
    }
    
    // Fallback
    exoPlayer?.trackSelectionParameters = exoPlayer?.trackSelectionParameters
        ?.buildUpon()
        ?.setPreferredAudioLanguage(trackLanguage)
        ?.build() ?: return
}
```

**After**:

```kotlin
override fun setPreferredAudioTrack(trackLanguage: String?, id: String?) {
    preferredAudioTrackLanguage = trackLanguage
    
    val formatId = id?.substringBeforeLast(":")
    val trackFormatIndex = id?.substringAfterLast(":")?.toIntOrNull() ?: -1
    
    exoPlayer?.currentTracks?.groups
        ?.filter { it.type == TRACK_TYPE_AUDIO }
        ?.find { group ->
            (0 until group.mediaTrackGroup.length).any { index ->
                group.mediaTrackGroup.getFormat(index).id == formatId
            }
        }
        ?.let { group ->
            exoPlayer?.trackSelectionParameters
                ?.buildUpon()
                ?.setOverrideForType(TrackSelectionOverride(group.mediaTrackGroup, trackFormatIndex))
                ?.build()
        }
        ?.let { newParams ->
            exoPlayer?.trackSelectionParameters = newParams
            return
        }
    
    // Fallback
    exoPlayer?.trackSelectionParameters = exoPlayer?.trackSelectionParameters
        ?.buildUpon()
        ?.setPreferredAudioLanguage(trackLanguage)
        ?.build()
}
```

**Key Improvements**:

- Parses composite ID format (`formatId:formatIndex`)
- Searches for the group containing the format with matching `formatId`
- Uses `formatIndex` to select the specific format within the group
- Gracefully falls back to language-based selection if ID-based selection fails
- Uses chained `?.let` for safe null handling

***

### 3. GeneratorPlayer.kt

**Major Changes**: Updated track selection dialog logic

#### Initial Audio Track Selection

**Before**:

```kotlin
var audioIndexStart = currentAudioTracks.indexOf(tracks.currentAudioTrack).takeIf {
    it != -1
} ?: currentVideoTracks.indexOfFirst {
    tracks.currentAudioTrack?.id == it.id
}
```

**After**:

```kotlin
var audioIndexStart = currentAudioTracks.indexOfFirst { track ->
    track.id == tracks.currentAudioTrack?.id &&
    track.formatIndex == tracks.currentAudioTrack?.formatIndex
}.takeIf { it != -1 } ?: 0
```

**Improvement**: Now correctly identifies the current audio track by matching both `id` AND `formatIndex`, since multiple formats can share the same group ID.

#### Audio Track Display

Enhanced to show detailed audio information including codec and channel configuration:

```kotlin
audioArrayAdapter.addAll(currentAudioTracks.mapIndexed { index, track ->
    val language = track.language?.let { fromTagToLanguageName(it) ?: it } 
        ?: track.label 
        ?: "Audio"
    
    val codec = track.sampleMimeType?.let { mimeType ->
        when {
            mimeType.contains("mp4a") || mimeType.contains("aac") -> "aac"
            mimeType.contains("ac-3") || mimeType.contains("ac3") -> "ac3"
            mimeType.contains("eac3") -> "eac3"
            // ... more codecs
        }
    } ?: "codec?"
    
    val channelConfig = when (track.channelCount ?: 0) {
        1 -> "mono"
        2 -> "stereo"
        6 -> "5.1"
        8 -> "7.1"
        else -> "${track.channelCount}Ch"
    }
    
    "[$index] $language $codec $channelConfig"
})
```

#### Apply Audio Track Selection

```kotlin
binding.applyBtt.setOnClickListener {
    val currentTrack = currentAudioTracks.getOrNull(audioIndexStart)
    player.setPreferredAudioTrack(
        currentTrack?.language, 
        "${currentTrack?.id}:${currentTrack?.formatIndex}"  // Composite ID
    )
    // ...
}
```

***

### ID Format

The composite ID format `"formatId:formatIndex"` is used throughout:

- **`formatId`**: The ExoPlayer group ID (same for all formats in a group)
- **`formatIndex`**: The index of the specific format within its group

### Selection Flow

1. User selects audio track in UI dialog
2. `GeneratorPlayer` creates composite ID: `"${track.id}:${track.formatIndex}"`
3. `CS3IPlayer.setPreferredAudioTrack()` receives composite ID
4. Parses ID to extract `formatId` and `formatIndex`
5. Finds the group containing a format with matching `formatId`
6. Applies `TrackSelectionOverride` with the group and `formatIndex`
7. Falls back to language-based selection if any step fails

### Benefits

- **No Reflection**: Uses only public ExoPlayer APIs
- **No Global State**: Tracks current audio using `group.isTrackSelected(formatIndex)`
- **Atomic Selection**: Composite ID ensures correct format selection
- **Graceful Degradation**: Falls back to language-based selection on failure
- **Better UX**: Displays detailed audio information (codec, channels)

## Testing Recommendations

1. Test with videos containing multiple audio tracks in the same language
2. Verify audio track persistence across episode changes
3. Test fallback to language-based selection
4. Verify UI displays correct currently-playing audio track
5. Test with videos containing various audio codecs (AAC, AC3, EAC3, Opus, etc.)

[^1]: IPlayer.kt
[^2]: UpdatedMatroskaExtractor.kt
[^3]: GeneratorPlayer.kt

<div align="center">⁂</div>

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
@@ -1450,7 +1450,7 @@ class GeneratorPlayer : FullScreenPlayer() {
                 var audioIndexStart = currentAudioTracks.indexOfFirst { track ->
                     track.id == tracks.currentAudioTrack?.id && 
                     track.formatIndex == tracks.currentAudioTrack?.formatIndex
-                }.takeIf { it != -1 } ?: 0
+                }.coerceAtLeast(0)
 
                 val audioArrayAdapter =
                     ArrayAdapter<String>(ctx, R.layout.sort_bottom_single_choice)


app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
```diff
@@ -3,6 +3,7 @@ package com.lagradost.cloudstream3.ui.player
 import android.content.Context
 import android.graphics.Bitmap
 import android.util.Rational
+import androidx.media3.common.Format
```
fire-light42: Please do not include any media3 specific structures here as IPlayer is supposed to be a generic player interface.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -382,36 +383,85 @@ class CS3IPlayer : IPlayer {
...
-                            audioTrack.first,
-                            audioTrack.second
+        // Try to use the index directly if provided
+        if (trackIndex != null) {
+            // Get all audio tracks flattened in the same order as getVideoTracks()
+            val audioTracks = exoPlayer?.currentTracks?.groups
+                ?.filter { it.type == TRACK_TYPE_AUDIO }
+            
+            if (audioTracks != null) {
+                val allFormats = audioTracks.flatMapIndexed { groupIndex, group ->
```
fire-light42: I do not understand why groupIndex is included here.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
```diff
@@ -302,7 +303,13 @@ interface IPlayer {
...
 
     /** If no trackLanguage is set it'll default to first track. Specifying the id allows for track overrides as the language can be identical. */
-    fun setPreferredAudioTrack(trackLanguage: String?, id: String? = null)
+    fun setPreferredAudioTrack(trackLanguage: String?, trackIndex: Int? = null)
+
+    // Gets a deterministic unique track number
+    fun getCurrentAudioTrackIndex(): Int?
+
+    // Gets a list of ExoPlayer Formats
+    fun getAudioFormats(): List<Format>?
```
fire-light42: This api is a bit weird, it should be a part of `getVideoTracks`.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -382,36 +383,85 @@ class CS3IPlayer : IPlayer {
             ?: return
     }
 
-    override fun setPreferredAudioTrack(trackLanguage: String?, id: String?) {
+    override fun setPreferredAudioTrack(trackLanguage: String?, trackIndex: Int?) {
```
fire-light42: I do not understand why this needs to be changed at all, would it not be possible to just remove stripTrackId in toAudioTrack and getTrack.

.github/workflows/debugBuild.yml
```diff
@@ -0,0 +1,82 @@
+name: Debug Build
```
fire-light42: This should be removed before the final version.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
```diff
@@ -302,7 +303,13 @@ interface IPlayer {
     fun setMaxVideoSize(width: Int = Int.MAX_VALUE, height: Int = Int.MAX_VALUE, id: String? = null)
 
     /** If no trackLanguage is set it'll default to first track. Specifying the id allows for track overrides as the language can be identical. */
-    fun setPreferredAudioTrack(trackLanguage: String?, id: String? = null)
+    fun setPreferredAudioTrack(trackLanguage: String?, trackIndex: Int? = null)
+
+    // Gets a deterministic unique track number
+    fun getCurrentAudioTrackIndex(): Int?
```
fire-light42: `getVideoTracks` should already contain this information.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
okibcn: @fire-light42  I didn't find any other way to have the codec (aac, eac3, dts, vorbis,..) and channel configuration (mono, stereo, 5.1, 7.1) of the audio track. I can't see an easy way to extract that audio info from exoplayer. 

Note: Cloudstream provides video data such as resolution and no info about audio. That is fine in a phone, but not for the TV app. For the TV app more info such as HDR, framerate, and audio info are important when selecting video/audio tracks.


EDIT: I see in the comments below what you mean. You suggest adding the get audio track formats using getVideoTracks.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
okibcn: That groupIndex is there because you need two different “index systems” at the same time:

a flat index (0, 1, 2, 3, …) over all audio tracks, which is what trackIndex refers to; and the real location of the track inside ExoPlayer: which group it belongs to and its index inside that group, which is what TrackSelectionOverride needs.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
okibcn: @fire-light42  That would involve changing the data returned by getVideoTracks(). At this time it is returning:
```js
        return CurrentTracks(
            exoPlayer?.videoFormat?.toVideoTrack(),
            exoPlayer?.audioFormat?.toAudioTrack(),
            currentTextTracks,
            videoTracks,
            audioTracks,
            textTracks
        )
```
That will change the definition of that function as it would no longer return a CurrentTrack type.
What should be the allowed return?

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
okibcn: @fire-light42 I’m unclear on the purpose of stripTrackId, so I didn’t remove it. I wasn’t sure **why** that data was being stripped originally.”

If that makes the id actually unique for each track, that should solve the issue in a simple way.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
okibcn: Noted. I will try to use it.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
okibcn: @fire-light42 Tested removing the stripTrackId and it didn't work:
At this time in the master cloudstream branch id string is something like `aac:spanish` for multiple tracks. If we remove stripTrackId, then the id is `0:aac:spanish`, but it is the same for multiple tracks as the first 0 is not the logical global track index.Maybe the problem is related to the generation of the track id.

In my solution, track id is simply the number of the track. Since it is a number, the logical way is to pass it as Int? and not as String?

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
fire-light42: I need to clarify that I fully understand what it does, but at the moment it is unused and therefore useless. Replace groupIndex with a random value and it will have no effect on the control flow or result.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
fire-light42: Checkout #2454 for a good implementation of this. You are only using the sampleMimeType and channelCount so just pass that into the structure instead of adding another API.

app/src/main/java/com/lagradost/cloudstream3/ui/player/IPlayer.kt
fire-light42: No, just change the VideoTrack and AudioTrack like #2454

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
fire-light42: Looks like media3 is a bit stupid. However, the logical way to pass this would be a pair of int and string, as we want to index into a specific trackgroup, not index all the tracks. A global index will cause a lot of problems.

okibcn:
@fire-light42 I have simplified the change, IPlayer.kt is now unaffected.

1. **onTracksChanged** is now changing the id of each audio track to make them unique. 
2. **getVideoTracks** provides the right information without changing input or output formats.
3. **setPreferredAudioTrack** now has a simpler logic and uses only the data it needs, without the need to provide groupIndex
4. **showTracksDialogue()** now properly detects the current track and displays extra info such as codec and speakers configuration.

Please read updated PR report for more details.

okibcn:
Today I have updated the PR to match the current project master.

It works and now uses the field added to AudioTrack data class.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -468,8 +480,22 @@ class CS3IPlayer : IPlayer {
         val videoTracks = allTracks.filter { it.type == TRACK_TYPE_VIDEO }
             .getFormats()
             .map { it.first.toVideoTrack() }
-        val audioTracks = allTracks.filter { it.type == TRACK_TYPE_AUDIO }.getFormats()
-            .map { it.first.toAudioTrack() }
+        
+        val audioTracks = allTracks.filter { it.type == TRACK_TYPE_AUDIO }
```
fire-light42: What is this? It is not the same as exoPlayer?.audioFormat

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1421,6 +1447,17 @@ class CS3IPlayer : IPlayer {
             exoPlayer?.addListener(object : Player.Listener {
                 override fun onTracksChanged(tracks: Tracks) {
                     safe {
+                        // Fix Format.id of new audio tracks with unique id
+                        var uniqueTrackId = 0
+                        tracks.groups.filter { it.type == TRACK_TYPE_AUDIO }.forEach { group ->
```
fire-light42: What is this? Using reflection to add metadata? Even if the id is a bit wonky just use `stripTrackId` with an index like exoplayer does. This is just a global index again.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -140,6 +140,7 @@ class CS3IPlayer : IPlayer {
             field = value
         }
 
+    private var currentAudioTrackIndex: Int = -1
```
fire-light42: What?

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
fire-light42: See https://github.com/recloudstream/cloudstream/pull/2447#discussion_r2727513494

fire-light42 (changes requested):
Good job on the channelCount, however please just add another parameter to setPreferredAudioTrack called trackIndex. You do not need to use reflection.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
okibcn: No, they are completely different, one is a list of track groups, the other is a list of media formats.

okibcn:
The new version avoids the use of reflection and there isn't any global currentAudioTrackIndex.

PR #2449, opened 2026-01-24
Fix subtitle selection

CranberrySoup:
Prevents user options getting erased and properly saves and selects the chosen language.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerSubtitleHelper.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerSubtitleHelper.kt
@@ -11,10 +11,8 @@ import androidx.media3.ui.SubtitleView
 import com.lagradost.cloudstream3.SubtitleFile
 import com.lagradost.cloudstream3.ui.subtitles.SaveCaptionStyle
 import com.lagradost.cloudstream3.ui.subtitles.SubtitlesFragment.Companion.setSubtitleViewStyle
-import com.lagradost.cloudstream3.utils.SubtitleHelper
-import com.lagradost.cloudstream3.utils.SubtitleHelper.fromCodeToLangTagIETF
+import com.lagradost.cloudstream3.utils.SubtitleHelper.fromLanguageToTagIETF
 import com.lagradost.cloudstream3.utils.UIHelper.toPx
-import me.xdrop.fuzzywuzzy.FuzzySearch
 
 enum class SubtitleStatus {
     IS_ACTIVE,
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerSubtitleHelper.kt
```diff
@@ -47,6 +50,56 @@ data class SubtitleData(
...
     }
 
+    /** Returns true if langCode is the same as the IETF tag */
+    fun matchesLanguage(langCode: String): Boolean {
+        return getIETF_tag() == langCode
+    }
+
+    /** Tries hard to figure out a valid IETF tag based on language code and name. Will return null if not found. */
+    fun getIETF_tag(): String? {
+        val tag = fromCodeToLangTagIETF(this.languageCode)
```
diogob003: Did you checked **data class SubtitleFile**? I feel like the same thing is done 2 times, in SubtitleFile we already try to convert lang -> ietf tag

https://github.com/recloudstream/cloudstream/blob/c1a2ae8704aa85c1110aca9750610fefbd647b92/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt#L1113-L1129

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerSubtitleHelper.kt
```diff
@@ -47,6 +50,56 @@ data class SubtitleData(
...
+        // First go for exact matches
+        SubtitleHelper.languages.forEach { language ->
+            if (language.languageName.equals(cleanedLanguage, ignoreCase = true) ||
+                language.nativeName.equals(cleanedLanguage, ignoreCase = true) ||
+                // Also match exact IETF tags
+                language.IETF_tag.equals(cleanedLanguage, ignoreCase = true)
+            ) {
+                return language.IETF_tag
+            }
+        }
```
diogob003: This is basically what is implemented in the `library` , the key difference is that we turn `languages` in `indexMapLanguageName` and `indexMapNativeName` then get it from the map instead of `.equals(cleanedLanguage)`. By doing it this way, the performance improved.

https://github.com/recloudstream/cloudstream/blob/c1a2ae8704aa85c1110aca9750610fefbd647b92/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt#L81-L104

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerSubtitleHelper.kt
```diff
@@ -47,6 +50,56 @@ data class SubtitleData(
...
+
+            // Arbitrary cutoff at 80.
+            if (cleanedLanguage.contains(language.languageName, ignoreCase = true) ||
+                cleanedLanguage.contains(language.nativeName, ignoreCase = true) || score > 80
+            ) {
+                if (score > closestMatch.second) {
+                    closestMatch = language.IETF_tag to score
+                }
+            }
+        }
```
diogob003: I find it more useful to implement the **FuzzySearch** within the **library**, so it can be used by extensions as well, and not just by the app itself.

My suggestion is to replace the `halfMatch` code with the `FuzzySearch` code. That's what I started doing but never finished or submitted a request 😬.

https://github.com/recloudstream/cloudstream/blob/c1a2ae8704aa85c1110aca9750610fefbd647b92/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt#L81-L103

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerSubtitleHelper.kt
diogob003: If I remember correctly, `SubtitleFile` is always turned into `SubtitleData`
SubtitleFile: what we get from plugins and providers
SubtitleData: the app's internal representation (that has more properties)

CranberrySoup:
Moved logic to `getLanguageDataFromName` :+1:

PR #2456, opened 2026-01-27
Feat: Zoom

fire-light42:
This is an updated patch of #1907 as zoom is a very cool feature. The main difference is how I did the transform as the old pull request did not feel right https://github.com/recloudstream/cloudstream/pull/1907#issuecomment-3289688607. The main reason it felt incorrect was that the zooming was done around the center of the screen, instead of the center of the pinch. I honestly did not want to reinvent the postScale function using pure math, so I just used the matrix operations and then extracted translation+scale. 

I also fixed some issues with black borders https://github.com/recloudstream/cloudstream/pull/1907#issuecomment-3314604831 and added an outline like youtube.

It should work fine on all horizontal videos, but I have noticed some weird behavior on vertical videos. It is also a bit wonky when zooming in when in the resize option "Zoom" as it instantly zooms out.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -101,6 +101,14 @@ import kotlin.math.min
 import kotlin.math.round
 import kotlin.math.roundToInt
 
+// You can zoom out more than 100%, but it will zoom back into 100%
+const val MINIMUM_ZOOM = 0.95f
+
+// How sensitive the auto zoom is to center at the min zoom
+const val ZOOM_SNAP_SENSITIVITY = 0.07f
+
+// Maximum zoom to avoid getting lost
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -1221,15 +1235,263 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+            // Set the transform to the correct x and y
+            newMatrix.postTranslate(
+                expectedTranslationX - translationX,
+                expectedTranslationY - translationY
+            )
+            zoomMatrix = newMatrix
+
+            val scaledAspect = scale * aspect
+            if (!animation) {
+                if ((scaledAspect - 1.0f).absoluteValue < 0.07f) {
```
CranberrySoup: I would like to see more comments describing magic values and advanced logic.

PR #2459, opened 2026-01-27
revert: circular progress bars

rockhero1234:


rockhero1234:
I also reverted check icon as that is linked to circular bar and other minor improvements revert are linked to circular bar

fire-light42:
> I also reverted check icon as that is linked to circular bar

No, what I mean is that it reverts a lot of refactoring work done in the same files. For example:
1. DownloadAdapter NoStateAdapter is being reverted into a ListAdapter
2. The colors of watchProgress is being reverted
3. The colors of download_delete_appbar is being reverted
4. GeneratorPlayer gets an extra setViewPos

And possibly other things I did not catch. Please do a manual revert to avoid rolling back refactoring work like https://github.com/recloudstream/cloudstream/commit/38296bfb1a791d07e4611068b292a2537c51dbad#diff-0fdb1731c5d1df329f2e95269eeec156915ba66bcc7d0e58e5f24ac3e28c0d7aR1

PR #2461, opened 2026-01-28
Adding Metdata on Player (Initial Draft) (TV)

phisher98:
image

Fade in 8 sec after paused

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -241,8 +241,6 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         return root
     }
 
-    private var hasStartedPlaying = false
-
     private fun scheduleMetadataVisibility() {
         val metadataScrim = playerBinding?.playerMetadataScrim ?: return
         val ctx = metadataScrim.context ?: return
@@ -260,7 +258,6 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
         }
 
         val isPaused = currentPlayerStatus == CSPlayerLoading.IsPaused
-        metadataScrim.animate().cancel()
         val token = ++metadataVisibilityToken
 
         if (isPaused) {
@@ -281,7 +278,10 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
                     .alpha(0f)
                     .setDuration(300L)
                     .setInterpolator(AccelerateDecelerateInterpolator())
-                    .withEndAction { metadataScrim.isVisible = false }
+                    .withEndAction {
+                        metadataScrim.alpha = 0f      // force final state
+                        metadataScrim.isVisible = false
+                    }
                     .start()
             }
         }
@@ -2227,7 +2227,6 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
     }
 
     protected fun uiReset() {
-        hasStartedPlaying = false
         metadataVisibilityToken++
         playerBinding?.playerMetadataScrim?.let {
             it.animate().cancel()
--- a/app/src/main/res/layout/player_custom_layout_tv.xml
+++ b/app/src/main/res/layout/player_custom_layout_tv.xml
@@ -63,7 +63,7 @@
                 android:textColor="#B3FFFFFF"
                 android:textSize="14sp"
                 android:maxLines="2"
-                tools:text="Animation • Action • Adventure • 2025 • 7.6"/>
+                tools:text="Animation • Action • Adventure • 2025 • ⭐ 7.6"/>
 
             <!-- SYNOPSIS -->
             <TextView


app/src/main/res/layout/player_custom_layout_tv.xml
```diff
@@ -9,6 +9,75 @@
     tools:orientation="vertical">
 
 
+    <FrameLayout
```
fire-light42, line 11: Please check the layout tests, as this will crash on other layouts even if it is not used. 
`R.layout.player_custom_layout,R.layout.player_custom_layout_tv, R.layout.trailer_custom_layout` must all have matching xmls to make the bind not crash.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -563,6 +593,7 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
     override fun onResume() {
         enterFullscreen()
         verifyVolume()
+        playerBinding?.root?.findViewById<View>(R.id.player_metadata_scrim)?.isVisible = false
```
fire-light42: Why?

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1541,6 +1541,58 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+            logoView.isVisible = false
+            titleView.isVisible = true
+            titleView.text = load.name
+        }
+
+        val meta = arrayOf(
+            load.tags?.takeIf { it.isNotEmpty() }?.joinToString(", "),
+            load.year?.toString(),
+            if (!load.type.isMovieType())
+                episode?.let { "S${it.season} • E${it.episode}" }
```
fire-light42: Both season and episode may be null. Use getShortSeasonText instead.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1541,6 +1541,58 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+        val logoView = overlay.findViewById<ImageView>(R.id.player_movie_logo)
+        val metaView = overlay.findViewById<TextView>(R.id.player_movie_meta)
+        val descView = overlay.findViewById<TextView>(R.id.player_movie_overview)
+
+        val load = viewModel.getLoadResponse() ?: return
+        val episode = currentMeta as? ResultEpisode
+
+        val logoUrl = load.logoUrl
+
+        if (!logoUrl.isNullOrBlank()) {
```
fire-light42: Use bindLogo

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
phisher98: My bad

app/src/main/res/layout/player_custom_layout_tv.xml
phisher98: I tried it but will check and maybe add in all layout

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
phisher98: Got it

app/src/main/res/layout/player_custom_layout_tv.xml
fire-light42: It is not about UI, but that it will crash. You can add it but make the visibility = gone.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
fire-light42: It was a question, I am wondering what this fixes, and if the metadataScrim animation may interfere with it.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
phisher98: Done

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
phisher98: I think I left it while understanding the code

Removed I didn't see any issues after removed

app/src/main/res/layout/player_custom_layout_tv.xml
phisher98: Added and Fixed

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -235,10 +238,38 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
             }
         }
-
         return root
     }
 
+    private fun scheduleMetadataVisibility() {
+        val metadataScrim = playerBinding?.playerMetadataScrim ?: return
+
+        if (!requireContext().shouldShowPlayerMetadata()) {
```
fire-light42: Please write `val ctx = metadataScrim.context ?: return` then ctx.shouldShowPlayerMetadata ... as requireContext will crash the application.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -235,10 +238,38 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+        }
+
+        if (isLayout(PHONE)) {
+            metadataScrim.isVisible = false
+            return
+        }
+
+        val isPaused = currentPlayerStatus == CSPlayerLoading.IsPaused
+
+        metadataScrim.animate().cancel()
```
fire-light42: I assume this is to cancel the postDelayed? Unfortunately animate().cancel will not cancel the metadataScrim.postDelayed.

fire-light42: The normal solution is to have a global counter variable that we increment, then when the lambda is invoked we check if that counter has not been incremented.

fire-light42 (changes requested):
The new UI looks really good, however I found a major UX bug. The problem is that the postDelayed never get cancelled, so pausing-unpausing repeatedly for 8 seconds will cause it to invoke the lambda several times even if you try to navigate. 

Moreover, I also want to ask if it intentional to show the "Scrim" even when you navigate the UI, as you can pull up the UI by pressing Up on the dpad while paused.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
phisher98: Updated

fire-light42 (approved):
The code looks good, will do some further testing later.

phisher98:
@fire-light42 Any more changes needed ?

phisher98:
> Looks good and works well but the metadataScrim should fade in and out as opposed to instantly appearing and disappearing.
> 
> If possible, see if you can make a more distinct visual difference between the synopsis and the meta row. If they are different it will be easier to distinguish where the synopsis starts at a glance. Notice how we display the synopsis in other places to make reading easier.
> 
> The yellow star indicating the ratings looks a bit out of place with the white font. You might not need to change this if you visually redesign the meta row.

Made some changes and Removed Star Icon

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -235,10 +238,55 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
...
+                    .setDuration(500L)
+                    .setInterpolator(DecelerateInterpolator())
+                    .start()
+                hidePlayerUI()
+            }, 8000L)
+        } else {
+            if (metadataScrim.isVisible) {
+                metadataScrim.animate()
+                    .alpha(0f)
+                    .setDuration(300L)
```
fire-light42, line 279: If you quickly unpause and pause this animation is cancelled midway leading to half-transparent text.

app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
```diff
@@ -235,10 +238,55 @@ open class FullScreenPlayer : AbstractPlayerFragment() {
                 requestUpdateBrightnessOverlayOnNextLayout()
             }
         }
-
         return root
     }
 
+    private var hasStartedPlaying = false
```
fire-light42: Looks like you accidentally left some code from testing, should this be kept?

phisher98: Removed

fire-light42 (approved):

PR #2462, opened 2026-01-28
revert circular progress,check icon & reapply accent color to watchprogress

rockhero1234:


fire-light42:
Well done pull request, however are you sure you want to revert all circular progress bars? For example, I find the circular white button resume on movies and tv-series much better as it removes the previous blue bar that was disconnected from the button. 

How do you feel about the different circular progress bars, and why do you want to revert it? Is there a UX or UI reason?

rockhero1234:
After using circular progress i found some ux issues
- it looks bad when progress is less than 2-3% as it shows a little dot.
- same when progress is not enough for check icon and around it it show a circle that also looks unfit.
- I realised they are bit hard to judge
- we also loose the purpose of accent color.
If we are reverting any circular bar we should revert all It will be bad ux if we kept some.
Thank You ❤️ for review

fire-light42:
> After using circular progress i found some ux issues
> 
>     * it looks bad when progress is less than 2-3% as it shows a little dot.
> 
>     * same when progress is not enough for check icon and around it it show a circle that also looks unfit.
> 
>     * I realised they are bit hard to judge
> 
>     * we also loose the purpose of accent color.
>       If we are reverting any circular bar we should revert all It will be bad ux if we kept some.
>       Thank You ❤️ for review

Then I would argue that this is not applicable to:
image

As it shows the exact details, and complements the download circular bar. In comparison this is not more readable, nor more conforming to the rest of the UI. This has the exact same problems, but makes it worse as it separates the action and information:
image

While I understand your concerns on the posters, I think this specific part of the "circular progress bars" really fits the current design to a T.

rockhero1234:
Screenshot_20260129-215640_CloudStream Beta
This looks bad as I said before and one of my major concern. Linear progress bar makes things clear

rockhero1234:
@fire-light42 will this be merged or i need to change something?

fire-light42:
I do not want to close pull requests people clearly have spent time on. Unfortunately I do not see the linear progress bar as an improvement over the circular progress bar. 

Yes, it does not look very good when at a low percentage, but that can be fixed by 'rounding up' low values to a minimum threshold.

fire-light42:
Perhaps there is another solution which we both can agree looks better? You would not need to implement it in code, concept art would work for a discussion.

Luna712:
@fire-light42 I think combined with #2493, this looks quite good anyway. But maybe there is a solution for everything to look better also.

PR #2467, opened 2026-01-30
Adding Download Metadata (Large Initial)

phisher98:
image

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/downloader/DownloadObjects.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/downloader/DownloadObjects.kt
@@ -69,6 +69,8 @@ object DownloadObjects {
         @JsonProperty("score") var score: Score? = null,
         @JsonProperty("description") val description: String?,
         @JsonProperty("cacheTime") val cacheTime: Long,
+        @JsonProperty("airDate") val airDate: Long? = null,
+        @JsonProperty("runtime") val runtime: Int? = null,
         override val id: Int,
     ) : DownloadCached(id) {
         @JsonProperty("rating", access = JsonProperty.Access.WRITE_ONLY)


app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
```diff
@@ -366,11 +376,129 @@ class DownloadAdapter(
...
+        val data = card.data
+
+        binding.apply {
+            episodePoster.loadImage(data.poster)
+            episodeText.text = root.context.getNameFull(data.name, data.episode, data.season)
+
+            val ratingText = data.score?.toString()
+
+            episodeRating.isVisible = !ratingText.isNullOrBlank()
+            episodeRating.text = ratingText?.let { "Rated: $it" }
```
Bnyro, line 394: nit: not translatable, i.e. you should use `getString(R.string.rated_format, it)` instead

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
```diff
@@ -366,11 +376,129 @@ class DownloadAdapter(
...
+
+            episodeRuntime.isVisible = data.runtime != null && data.runtime > 0
+            episodeRuntime.text = data.runtime?.let { "$it min" }
```
Bnyro: This only formats minutes well, you should however support longer durations (e.g. 20 hours or 5 seconds).

So you should instead use `secondsToReadable` https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/syncproviders/AccountManager.kt#L142

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadAdapter.kt
```diff
@@ -382,38 +510,47 @@ class DownloadAdapter(
...
     @SuppressLint("NotifyDataSetChanged")
     fun setIsMultiDeleteState(value: Boolean) {
         if (isMultiDeleteState == value) return
         isMultiDeleteState = value
         notifyDataSetChanged() // This is shit, but what can you do?
     }
 
-    private fun toggleIsChecked(checkbox: CheckBox, itemId: Int) {
+    private fun toggleIsChecked(checkbox: CheckBox?, itemId: Int) {
+        if (checkbox == null) return
```
Bnyro: Why allow to pass `null` here? You should instead make sure that you never call `toggleIsChecked` with `null` as `checkbox` parameter.

app/src/main/res/layout-v23/download_child_episode_large.xml
Bnyro, line 1: Btw, why did you add a separate layout for SDK23 and above? What's the difference between these layouts?

phisher98: Ide automatically created it I don't know why 

Should I delete it ?

Bnyro: Yes, it's unnecessary

phisher98: Done

PR #2469, opened 2026-01-31
Readd extra brightness feature as optional setting

Pawloland:
Readd extra brightness feature added in #2416 and removed in c1a2ae8704aa85c1110aca9750610fefbd647b92
 as optional setting and fix brightness icon animation on gesture in a player.

In the meantime the zoom feature was added, and it is incompatible with the brightness filter in the current form. When the filter is disabled in settings (it is by default), then zooming works fine. With filter enabled, only the 3 preprogrammed modes are working.


Pawloland:
It now uses simple view overlay and supports changing scaling with a button or by pinch to zoom&pan. Pip window resizing also works correctly. Ready for review and merge on my side.

PR #2477, opened 2026-02-02
Update gradle to 9.3.1

Luna712:


fire-light42:
Sisyphus would be happy with gradle.

PR #2491, opened 2026-02-11
add Videa.hu extractor

KaifTaufiq:
Created New Videa.hu extractor

image

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Videa.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Videa.kt
@@ -16,8 +16,6 @@ class Videa : ExtractorApi() {
     override val requiresReferer = false
 
     private val videaSecret = "xHb0ZvME5q8CBcoQi6AngerDu3FGO9fkUlwPmLVY_RTzj2hJIS4NasXWKy1td7p"
-    private var key = ""
-    private var cookie = ""
 
     override suspend fun getUrl(
         url: String,
@@ -26,11 +24,11 @@ class Videa : ExtractorApi() {
         callback: (ExtractorLink) -> Unit
     ) {
         var currentUrl = url
-        var found = false
-
+        var key = ""
+        var lastUrl: String? = null
         // Handle redirect loop until we get valid XML
-        while (!found) {
-            val webUrl = getXmlUrl(currentUrl) ?: return
+        while (true) {
+            val webUrl = getXmlUrl(currentUrl) { cookie -> /* no-op, cookie not used */ } ?: return
             val response = app.get(webUrl)
             val rawBytes = response.body.bytes()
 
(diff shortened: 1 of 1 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Videa.kt
```diff
@@ -0,0 +1,206 @@
+// Adapted for CloudStream - taken from https://github.com/vargalex/ResolveURL/blob/fix/videa-resolver-add-cookie/script.module.resolveurl/lib/resolveurl/plugins/videa.py
+package com.lagradost.cloudstream3.extractors
+
+import com.lagradost.cloudstream3.app
+import com.lagradost.cloudstream3.SubtitleFile
+import com.lagradost.cloudstream3.utils.*
+import android.util.Base64
```
fire-light42: No android imports please. The lib is used for cross platform development, and as such will not compile if you use the android imports. Instead use the helper base64 functions.

KaifTaufiq: Replaced
image

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Videa.kt
```diff
@@ -0,0 +1,206 @@
...
+ * Extractor for Videa.hu video hosting service
+ * Handles encrypted XML responses and redirect chains
+ */
+class Videa : ExtractorApi() {
+    override val name = "Videa"
+    override val mainUrl = "https://videa.hu"
+    override val requiresReferer = false
+
+    private val videaSecret = "xHb0ZvME5q8CBcoQi6AngerDu3FGO9fkUlwPmLVY_RTzj2hJIS4NasXWKy1td7p"
+    private var key = ""
```
fire-light42: Why is there class variables? Extractors are supposed to be stateless. CloudStream uses the same extractor to dispatch several getUrl in parallel, and as such this looks like a multithreading issue waiting to happen.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Videa.kt
```diff
@@ -0,0 +1,206 @@
...
+                val xsHeader = response.headers["X-Videa-Xs"] ?: return
+                key += xsHeader
+                rc4DecryptBytes(rawBytes, key)
+            }
+
+            // Check for redirect in XML error
+            val redirectMatch = """<error.*?"noembed".*>(.*)</error>""".toRegex().find(videaXml)
+
+            if (redirectMatch != null) {
+                currentUrl = redirectMatch.groupValues[1]
```
fire-light42, line 57: Could you please also include a check to prevent inf recursion. E.g. && redirectMatch.groupValues[1] != currentUrl

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Videa.kt
```diff
@@ -0,0 +1,206 @@
...
+            } else {
+                found = true
```
fire-light42: Just use break?

fire-light42 (changes requested):

PR #2493, opened 2026-02-12
better episode poster ui

rockhero1234:
#2462 need to merge before this
Screenshot_20260212-144213_CloudStream Debug
Thumbnail is more visible now


voidforcex9:
Screenshot_2026-02-12-15-26-18-24_bba4627ecec1346e77f282610a5b88e8
Hii buddy i tried your pr, the season and episode number number is showing correctly but I think there is no need of that pause icon on bottom left side of the thumbnails. How about replacing it with ratings or just removing it?

Dexbuddy:
Great work on the UI improvements. The new layout looks much cleaner and more refined.

I have a small suggestion regarding the thumbnails. Instead of showing the Season and Episode number like S1:E1 in the bottom right corner, maybe consider displaying the episode runtime there in HH:MM format. It could give users a quick sense of the episode length at a glance and add more practical context.

saimuelbr:
> Great work on the UI improvements. The new layout looks much cleaner and more refined.
> 
> I have a small suggestion regarding the thumbnails. Instead of showing the Season and Episode number like S1:E1 in the bottom right corner, maybe consider displaying the episode runtime there in HH:MM format. It could give users a quick sense of the episode length at a glance and add more practical context.

that's a good point, but i think this need added in description episode to be more clean

phisher98:
@fire-light42 could you please review this as well?

PR #2500, opened 2026-02-14
Replace Google suggest API with TheMovieDB multi-search endpoint

NivinCNC:
Replace Google suggest API with TheMovieDB multi-search endpoint to avoid junky suggestions (Now the suggestions are only related to movies, series, animes etc)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchSuggestionApi.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchSuggestionApi.kt
@@ -1,18 +1,33 @@
 package com.lagradost.cloudstream3.ui.search
 
+import com.fasterxml.jackson.annotation.JsonProperty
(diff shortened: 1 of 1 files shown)


BoakyeDarlington:
> Replace Google suggest API with TheMovieDB multi-search endpoint to avoid junky suggestions (Now the suggestions are only related to movies, series, animes etc)

Can u please increase the speed of the search results. My data connection is very good but I have to wait some seconds to see the result which makes it irrelevant for some people. Nowadays we mostly type faster so if the result delay it won't help. Pls increase the speed. Thank you

BoakyeDarlington:
> Replace Google suggest API with TheMovieDB multi-search endpoint to avoid junky suggestions (Now the suggestions are only related to movies, series, animes etc)

I've also noticed this, when typing the result won't pop up unless I finish or pause it for some sec to show the results. It needs to pop up when am still typing like chrome browser

Issue #2403, opened 2026-01-07
Feature Request: Setting to reorder extensions in search tab


xx-ZORO-xx:
### Describe your suggested feature

Hello Developers, I hope you're all doing great!
I have a small request: Currently, when searching for movies/shows the list of source providers (extensions) appears in a random order. It would be very helpful if we had the option to rearrange the provider list according to our preference. [Something like the reoder section in ultima extension] This way, the desired providers would always appear at the top.

![Image](https://github.com/user-attachments/assets/2a64d516-0757-44d4-89d6-ef476d35266e)
.
![Image](https://github.com/user-attachments/assets/d800e0bd-f4af-4202-83e5-7fd1add3e9a7)

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

fire-light42:
Search results are given in the chronological order of response time. As such they are not subject to any deterministic behavior and will race to be the first result. This is done to show a result as fast as possible, but will cause races between extensions. 

However, I am open to the idea of pinned extensions having a priority slot on top. A full ordering system is overkill.

xx-ZORO-xx:
> Search results are given in the chronological order of response time. As such they are not subject to any deterministic behavior and will race to be the first result. This is done to show a result as fast as possible, but will cause races between extensions.
> 
> However, I am open to the idea of pinned extensions having a priority slot on top. A full ordering system is overkill.

Thanks! The pinned extensions idea is perfect. please implement it whenever possible.
Looking forward to the update!

xx-ZORO-xx:
> Search results are given in the chronological order of response time. As such they are not subject to any deterministic behavior and will race to be the first result. This is done to show a result as fast as possible, but will cause races between extensions.
> 
> However, I am open to the idea of pinned extensions having a priority slot on top. A full ordering system is overkill.

Bro any updates? @fire-light42 
(pinned extensions in search tab)

Issue #2263, opened 2025-11-26
Allow pinned providers to appear first in search


Luna712:
### Describe your suggested feature

I think it makes sense to either add a separate pinning system for search, adding a way to prioritize providers in search, or just use the existing pin implementation and pin the providers at the top of search also. I always find it difficult to find the right provider in search and skim over it a few times, even though I only have a few providers installed in the first place.

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

xx-ZORO-xx:
Yes 👍🏻

Issue #756, opened 2023-11-09
Extension Pinning feature


EHMilon:
### Describe your suggested feature

 Add provider pining option in CS3.
In **Setting** menu under **Provider** section, you should create a setting for **Pin extension** .
### Pin Extension
- Pin extension search first 
- Search only pin extension.

Sometime, our favorite's provider search position index at the bottom, and we need to scroll & scan to find our favorite provider's. And It will also help to minimalistic our search result. I think this feature will be very useful for us.

### Other details

![Screenshot_20231110-0126](https://github.com/recloudstream/cloudstream/assets/107425940/017cd91e-b24e-459d-946a-0c88bcc0046f)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

fire-light42:
https://github.com/recloudstream/cloudstream/issues/225
Ui wise that slot is taken by provider specific settings

jackuars7:
This is a great idea, where I need to see results from my preferred providers at the top

EHMilon:
@LagradOst

Issue #225, opened 2022-11-28
pin source 


Light7654:
### Describe your suggested feature

add a feature to pin your favourite source 

### Other details

like pinning superstream and hdtoday

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Blatzar:
What would pinning a source do?

C10udburst:
> What would pinning a source do?

pin it ofc

Indranil012:
> What would pinning a source do?

I think it's like when searching the pinned provider will be always on top.

eightyy8:
+1 
It's annoying having to scroll through providers in the search results.

jackuars7:
+1 for this too

Gthbacc:
+1 for this.

kurwahuj123:
+1

Rayforest:
I think the op is requesting for a separate provider filter option. Just like we have filters for the movies, series, anime, livestreams, nsfw etc. Idk about the op, but I would very much love to see this being added. A separate filter named 'favourites' or something similar in which we can add our favourite providers (like sorastream, superstream, HDToday etc)

PR #2501, opened 2026-02-14
Search result provider pining feature added.

swapnilkuwar:
In homepage we are already pinning the providers, same provider order we can use to pin the search results.

Closes: #756,
Closes: #225,
Closes: #2263,
Closes: #2403


xx-ZORO-xx:
Bro, I think it's not working @fire-light42, @swapnilkuwar.
No as such feature is visible.

swapnilkuwar:
Its not directly visible in UI.
in homepage whatever providers you have pinned the same order will be used in the search screen.

swapnilkuwar:
https://github.com/recloudstream/cloudstream/pull/2507
@fire-light42 this PR is created to focus the first search item in case of pinned providers.

PR #2510, opened 2026-02-18
Revert "Revert "Update media3 to 1.9.2""

Luna712:
Reverts recloudstream/cloudstream#2509
https://github.com/recloudstream/cloudstream/pull/2342#issuecomment-3917766504


Luna712:
@fire-light42 I've tested that this does fix the crashes I'm aware of. Unless there was something else also?

fire-light42:
It is not necessary to downgrade the NextLib, we only need to override the NextRenderersFactory buildTextRenderers.

Luna712:
> It is not necessary to downgrade the NextLib, we only need to override the NextRenderersFactory buildTextRenderers.

@fire-light42 

I guess thats true. Can that be a separate patch, or someone else can commit to this one? I apologize I'm unable to do much more work here right now due to being much more busy IRL now so it will be awhile before I could get to that if so, and I would like media3 updated anyway to fix some safe mode crashes that happens sometimes in particular.

PR #2511, opened 2026-02-18
Remove dead Extractors

saimuelbr:
removing dead sites


fire-light42:
Please also remove them from the list, otherwise you will get compile errors.

saimuelbr:
> Please also remove them from the list, otherwise you will get compile errors.

yeah, i forget send the ExtractorApi.kt with the first commit

(i'll fix the errors of build)

saimuelbr:
fixed

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/VidNest.kt
```diff
@@ -1,50 +0,0 @@
...
-
-import com.lagradost.cloudstream3.app
-import com.lagradost.cloudstream3.utils.ExtractorApi
-import com.lagradost.cloudstream3.utils.ExtractorLink
-import com.lagradost.cloudstream3.utils.M3u8Helper
-import com.lagradost.cloudstream3.utils.newExtractorLink
-
-class Vidnest : AsianLoad() {
-    override var name = "Vidnest"
-    override var mainUrl = "https://vidnest.io"
```
fire-light42, line 11: This link still seems to be up and running as a video host. Are we sure it is dead?

saimuelbr: vidnest is online, i should revert this change?

fire-light42: That would be nice, as we do not want to remove working extractors.

saimuelbr: right, i'll add again

saimuelbr: > That would be nice, as we do not want to remove working extractors.

i rename the extractor to VidNest.kt

saimuelbr:
will be merged?

PR #2515, opened 2026-02-19
Tv landscape rotation

onurcvnoglu:
On TV, videos that are meant to be watched in portrait mode (such as short series) were previously rotating automatically in the player. Instead of that behavior, a manual rotate button has now been added.


app/src/main/res/layout/player_custom_layout_tv.xml
```diff
@@ -1,33 +1,33 @@
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
fire-light42, line 1: Please do not change files that are not relevant to this pull request. Especially for a small change.

PR #2517, opened 2026-02-19
Bump jsoup and zipline libs

Luna712:


fire-light42 (commented):
Might need to bump nicehttp to avoid possible conflict.

Luna712:
> Might need to bump nicehttp to avoid possible conflict.

@fire-light42

Blatzar/NiceHttp#14

fire-light42 (approved):

PR #2528, opened 2026-02-21
offline pin:minor fix and icon change to fit properly

rockhero1234:
photo_6111573389894946461_y

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadButtonSetup.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadButtonSetup.kt
@@ -149,8 +149,7 @@ object DownloadButtonSetup {
                                 uri = Uri.EMPTY,
                                 id = it.id,
                                 parentId = it.parentId,
-                                name = it.name
-                                    ?: (act.getString(R.string.episode) + it.episode),
+                                name = it.name ?: (act.getString(R.string.downloaded_file)),
                                 season = it.season,
                                 episode = it.episode,
                                 headerName = parent.name,
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadButtonSetup.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadButtonSetup.kt
@@ -149,7 +149,8 @@ object DownloadButtonSetup {
                                 uri = Uri.EMPTY,
                                 id = it.id,
                                 parentId = it.parentId,
-                                name = it.name ?: act.getString(R.string.downloaded_file),
+                                name = it.name
+                                    ?: (act.getString(R.string.episode) + it.episode),
                                 season = it.season,
                                 episode = it.episode,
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadButtonSetup.kt
```diff
@@ -149,7 +149,8 @@ object DownloadButtonSetup {
                                 uri = Uri.EMPTY,
                                 id = it.id,
                                 parentId = it.parentId,
-                                name = it.name ?: act.getString(R.string.downloaded_file),
+                                name = it.name
+                                    ?: (act.getString(R.string.episode) + it.episode),
```
fire-light42: This contains three fundamental issues.

1. This may get triggered on a movie without a name.
2. There is no spacing between episode and the number resulting in "Episode1"
3. Episode number and season are already shown in the player? It would result in something like:
```
Episode 1 - 1920x1080
Breaking Bad "S1:E1" - Episode 1 ✓
```

app/src/main/res/layout/player_custom_layout.xml
```diff
@@ -180,12 +180,12 @@
...
-                            tools:text="Hello world " />
+                            tools:text="Hello world" />
                         <ImageView
                             android:id="@+id/offline_pin"
-                            android:layout_width="16dp"
-                            android:layout_height="16dp"
-                            android:layout_marginStart="2dp"
+                            android:layout_width="15dp"
+                            android:layout_height="15dp"
+                            android:layout_marginStart="1dp"
```
fire-light42, line 188: Should the TV player not get changed too?

fire-light42 (changes requested):

app/src/main/res/layout/player_custom_layout.xml
rockhero1234: That's already aligned

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadButtonSetup.kt
rockhero1234: I changed subName if name is downloaded file it may work

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1761,7 +1761,7 @@ class GeneratorPlayer : FullScreenPlayer() {
 
             is ExtractorUri -> {
                 headerName = meta.headerName
-                subName = meta.name
+                subName = if(meta.name != context?.getString(R.string.downloaded_file)) meta.name else ""
```
fire-light42: Why?

rockhero1234: I can't recollect it, I removed it for now. You can test to merge

PR #2546, opened 2026-03-01
Auto-scroll to first non-empty search item in case of pinned providers.

hsp1020:
This PR re-submits the changes from #2507, which was closed because the head repository was deleted.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt
@@ -521,7 +521,38 @@ class SearchFragment : BaseFragment<FragmentSearchBinding>(
                     }
 
                     submitList(newItems)
-                    //notifyDataSetChanged()
+                    
+                    val firstNonEmptyIndex = newItems.indexOfFirst { it.list.list.isNotEmpty() }
+                    if (firstNonEmptyIndex != -1 && pinnedOrder.isNotEmpty()) {
+                        val masterRecycler = binding.searchMasterRecycler
+
+                        masterRecycler.post {
+                            try {
+                                masterRecycler.scrollToPosition(firstNonEmptyIndex)
+
+                                masterRecycler.post {
+                                    val parentVH = masterRecycler.findViewHolderForAdapterPosition(firstNonEmptyIndex)
+                                    val parentView = parentVH?.itemView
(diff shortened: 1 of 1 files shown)


hsp1020:
I opened this PR several weeks ago, but haven't received any feedback yet. Is there a reason for the delay? I'd appreciate it if you could let me know when you might be able to review it. Thanks.

fire-light42:
Sorry, there have simply been too many pull requests for me to review everything in a timely manner.

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt
```diff
@@ -521,7 +521,38 @@ class SearchFragment : BaseFragment<FragmentSearchBinding>(
...
+                                        ?.findViewById<androidx.recyclerview.widget.RecyclerView>(R.id.home_child_recyclerview)
+
+                                    childRecycler?.post {
+                                        val firstChild =
+                                            childRecycler.layoutManager?.findViewByPosition(0)
+                                                ?: childRecycler.findViewHolderForAdapterPosition(0)?.itemView
+
+                                        firstChild?.let { view ->
+                                            view.isFocusableInTouchMode = true
+                                            view.requestFocus()
```
fire-light42, line 548: Do not focus the item unless necessary. This steals focus from the user on a TV, and it focuses when the user is on a phone (which should never happen).

app/src/main/java/com/lagradost/cloudstream3/ui/search/SearchFragment.kt
```diff
@@ -521,7 +521,38 @@ class SearchFragment : BaseFragment<FragmentSearchBinding>(
...
                     submitList(newItems)
-                    //notifyDataSetChanged()
+                    
+                    val firstNonEmptyIndex = newItems.indexOfFirst { it.list.list.isNotEmpty() }
+                    if (firstNonEmptyIndex != -1 && pinnedOrder.isNotEmpty()) {
+                        val masterRecycler = binding.searchMasterRecycler
+
+                        masterRecycler.post {
+                            try {
+                                masterRecycler.scrollToPosition(firstNonEmptyIndex)
```
fire-light42, line 531: This makes it impossible for users to scroll down while the results are loading. Make sure the user inputs take precedence over anything you are doing automatically.

This is best tested by installing many providers, you will see it becomes impossible to do anything while new results appear.

fire-light42 (changes requested):
Good idea, but requires refinement for a better UX.

PR #2547, opened 2026-03-01
Prefer player default live position for HLS/DASH

NivinCNC:
When starting a single HLS (M3U8) or DASH stream and no explicit start position was provided (playbackPosition == 0), set playbackPosition to TIME_UNSET so ExoPlayer can choose its default/live position. This prevents forcing a 0 ms start for live/non-playlist streams and lets the player resume at the correct live/default offset.

I made this PR because when i play HLS/DASH live streams that has larger Playback window eg: 1hr, the stream plays from 0ms rather than playing live which always  made to make me fastforward to live. This PR fixes this


fire-light42:
Looks good but I will need to do some testing before merging :+1:

Issue #2527, opened 2026-02-21
Kitsu account not syncing anime episode


hellowhatis22-lang:
### Steps to reproduce

When a episode watched the the checked mark appeared but right after kitsu account not syncing episode. 

### Expected behavior

Kitsu account should be sync after each episode watched. 

### Actual behavior

Watched episode is not syncing in kitsu account. 

### Cloudstream version and commit hash

4.62

### Android version

Android 14 

### Logcat

```java

```

### Acknowledgements

- [x] I am sure my issue is related to the app and **NOT some extension**.
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [x] I will fill out all of the requested information in this form.

PiterWeb:
Since I implemented the kitsu sync I have been using it, and it works properly with a exception that probably is what is causing this bug for you.

Kitsu has in their API docs an URL that it is kitsu.io/api/edge/, I have tested that kitsu.io sometimes is down but the same URL with kitsu.app (which is not the "official" API URL) works perfect.

The bug might be because when you are watching the episode in that moment kitsu.io is down and can't update the episode count.

I can try to add the kitsu.app URL as fallback. When kitsu.io fails

hellowhatis22-lang:
As you can see below images. I have watched anime blue miburo episode 9 but in kitsu didn't sync. In the beginning kitsu works totally fine. About two weeks this problem occurs. If possible please fix it. 

![Image](https://github.com/user-attachments/assets/f027e7cb-bec3-43ac-b9a7-3116ffdfbbc3)

![Image](https://github.com/user-attachments/assets/ce95a1a6-ec11-448a-81c0-4b1769fbb425)

hellowhatis22-lang:
> Since I implemented the kitsu sync I have been using it, and it works properly with a exception that probably is what is causing this bug for you.
> 
> Kitsu has in their API docs an URL that it is kitsu.io/api/edge/, I have tested that kitsu.io sometimes is down but the same URL with kitsu.app (which is not the "official" API URL) works perfect.
> 
> The bug might be because when you are watching the episode in that moment kitsu.io is down and can't update the episode count.
> 
> I can try to add the kitsu.app URL as fallback. When kitsu.io fails

[...]

PR #2552, opened 2026-03-06
Add fallback url for kitsu sync

PiterWeb:
As I mentioned in #2527 kitsu.io sometimes is down while kitsu.app is working and are equivalent but kitsu.io is the one that is documented as the api url domain.

So I added a fallback when any of the requests targeting kitsu.io fails it tries with kitsu.app.

This may fix the issue about not syncing anime.


PiterWeb:
Yeah you are very right, I will investigate on interceptors or create a function like you said to do the code DRY.

> If you need to log errors you should use our built-in function logError(throwable: Throwable).

OK, thank you for pointing out.

PiterWeb:
I have moved the fallback logic to an OkHttp interceptor. Now it looks so much clean :D

PR #2553, opened 2026-03-08
Fixed shared pool, closes #2082

### LINKED ISSUE #2082: Changing theme doesn't immediately change colors in all places
### Steps to reproduce

1. Change theme from dark or light or vice versa
2. Go somewhere like to extensions or the home page in some cases

### Expected behavior

The colors should be changed to new theme colors.

### Actual behavior

The remain on the old theme colors until, in some cases like the home page, you leave an reenter fragment again, or in other cases until you restart the app entirely.

Image
Image
Image
Image
Image

4.6.1 6c30af3

### Android version

Android 16

### Logcat

```java

```

### Other details

_No response_

### Acknowledgements

LagradOst:
This should fix all the sharedPool issues without clearing them. The problem is that views are shared between context barriers, creating nonsense bugs like coil not rendering images, or cards rendering in the wrong theme. This is done with a `WeakHashMap<Context, RecyclerView.RecycledViewPool>` to map the context to the correct pool, and avoid storing a strong reference to the context. Moreover I used a lambda function as the constructor as it will be used several times. I have no idea why android do not do this by default. The implementation also uses synchronized to avoid any runtime crash, but this should never happen given that all UI should be done on the same thread.  

While I do not expect any bugs given that this fixed my issues in QN, there might be some views that are not rendered correctly in some specific circumstance. As such any testing would be appreciated. If you can get a view to render with the wrong image or with the wrong theme, then ping me.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/BaseAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/BaseAdapter.kt
@@ -1,5 +1,6 @@
 package com.lagradost.cloudstream3.ui
 
+import android.content.Context
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
@@ -11,6 +12,7 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.recyclerview.widget.RecyclerView.ViewHolder
 import androidx.viewbinding.ViewBinding
 import coil3.dispose
+import java.util.WeakHashMap
 import java.util.concurrent.CopyOnWriteArrayList
 
 open class ViewHolderState<T>(val view: ViewBinding) : ViewHolder(view.root) {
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/BaseAdapter.kt
```diff
@@ -22,6 +24,33 @@ abstract class NoStateAdapter<T : Any>(
...
+ * The reason for this complicated structure is that a pool should not be shared between contexts
+ * as it makes coil fuck up, and theming.
+ * */
+fun newSharedPool(lambda: RecyclerView.RecycledViewPool.() -> Unit = { }): Pair<WeakHashMap<Context, RecyclerView.RecycledViewPool>, RecyclerView.RecycledViewPool.() -> Unit> =
+    WeakHashMap<Context, RecyclerView.RecycledViewPool>() to lambda
+
+/** Sets the shared pool of the recyclerview */
+fun RecyclerView.setRecycledViewPool(pool: Pair<WeakHashMap<Context, RecyclerView.RecycledViewPool>, RecyclerView.RecycledViewPool.() -> Unit>) {
+    val ctx = context ?: return
+    synchronized(pool.first) {
```
fire-light42, line 38: Locking a context object _may_ cause deadlocks or other unpredictable issues, since the context can be used literally everywhere else. 

In practice it seems to work well but I want to avoid any risks of weird issues arising from this.

fire-light42: Nevermind, I got confused by the nested types. You are locking on `WeakHashMap<Context, RecyclerView.RecycledViewPool>`, which will not cause issues.

PR #2570, opened 2026-03-17
Add source hiding

CranberrySoup:
- Adds a quality profile settings menu
- Adds "Hide sources with a negative priority" option
- Adds "Hide sources with errors" option

The code contains redundant sortLinks and toDisplayLink calls and it could be made more efficient, but that would increase code complexity. The caches in QualityDataHelper.kt reduces sortLinks runtime to 0-2ms even with 200 links. 

Admittedly the code is quite messy but I could not make it better. Suggestions welcome.


app/src/main/java/com/lagradost/cloudstream3/ui/player/source_priority/QualityDataHelper.kt
```diff
@@ -53,13 +56,21 @@ object QualityDataHelper {
         val types: Set<QualityProfileType>
     )
 
+
+    // Map profile and name to priority
+    val sourcePriorityCache: HashMap<Int, HashMap<String, Int>> = hashMapOf()
```
fire-light42: This should be a ConcurrentHashMap for better thread safety.

app/src/main/java/com/lagradost/cloudstream3/ui/player/source_priority/QualityDataHelper.kt
```diff
@@ -107,8 +126,24 @@ object QualityDataHelper {
...
+        if (value == setting.defaultValue) {
+            removeKey(folder, setting.key)
+        } else {
+            setKey(folder, setting.key, value)
+        }
     }
 
+    inline fun <reified T : Any> getProfileSetting(profile: Int, setting: ProfileSettings<T>): T {
+        val folder = "$currentAccount/$VIDEO_PROFILE_SETTINGS/$profile"
+        val value = getKey<T>(folder, setting.key)
```
Luna712, line 144: I am not positive either way but does this work with kotlinx serialization?

fire-light42 (approved):

PR #2574, opened 2026-03-23
episode playback preferences for remembering source and subtitle selections.

onurcvnoglu:
Remembers your selected source and subtitle options for each series. Since multiple sources and subtitles may be available and the default selection may not always be accurate, your choice will be automatically applied to subsequent episodes. This way, you won’t need to reselect them every time you switch episodes.
Some parts of this feature were developed with the assistance of AI, but all implementations have been reviewed and thoroughly tested.


app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -130,6 +132,91 @@ import kotlinx.coroutines.launch
 import java.io.Serializable
 import java.util.Calendar
 
+private const val PLAYER_EPISODE_PREFERENCES = "player_episode_preferences"
+
+internal data class EpisodePlaybackPreference(
+    val sourceDisplayName: String? = null,
+    val subtitleOriginalName: String? = null,
+    val subtitleSuffix: String? = null,
```
fire-light42: The suffixes are not stable, they get assigned by order of arrival. 
If you want a more stable subtitle saving you should save the link instead of suffix. Subtitle links are generally _relatively_ stable, but it is not a perfect solution.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1335,20 +1515,38 @@ class GeneratorPlayer : FullScreenPlayer() {
...
-                            subtitlesGroupedList.getOrNull(subtitleGroupIndex - 1)?.value?.getOrNull(
-                                subtitleOptionIndex
-                            )?.let {
+                            selectedSubtitle?.let {
                                 setSubtitles(it, true)
                             } ?: false
                         }
+
+                        if (subtitleGroupIndex <= 0 || selectedSubtitle != null) {
+                            persistSelectedSubtitlePreference(currentMeta, selectedSubtitle)
```
fire-light42: Just move the saving logic to setSubtitles, that makes more sense and is already where we store subtitle preference data.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -130,6 +132,91 @@ import kotlinx.coroutines.launch
 import java.io.Serializable
 import java.util.Calendar
 
+private const val PLAYER_EPISODE_PREFERENCES = "player_episode_preferences"
+
+internal data class EpisodePlaybackPreference(
```
fire-light42: These functions can be moved to a separate file inside an object (something like EpisodePreferenceHelper) to reduce cluttering. 

This would also make it easy to know when episode preferences are managed. Something like `EpisodePreferenceHelper.someMethod` is a dead giveaway of what happens as opposed to `resolveStoredSubtitlePreference` managing *some* subtitle preference.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -520,6 +688,20 @@ class GeneratorPlayer : FullScreenPlayer() {
             hasRequestedStamps = false
 
         loadExtractorJob(link.first)
+        val preferredSubtitle = if (sameEpisode) {
```
fire-light42, line 539: This is unnecessarily complex. Move logic to getAutoSelectSubtitle, otherwise we have logic for automatically selecting subtitles outside of getAutoSelectSubtitle. Then it is impossible for maintainers to know how each preference system supersedes each other.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -230,6 +322,82 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+        )
+    }
+
+    private fun updateEpisodePreference(
+        meta: Any?,
+        update: EpisodePlaybackPreference.() -> EpisodePlaybackPreference,
+    ) {
+        val parentId = getEpisodePreferenceParentId(meta) ?: return
+        val updatedPreference = update(getEpisodePreference(meta) ?: EpisodePlaybackPreference())
+        setKey(
```
fire-light42: This data is persistent and impossible to remove user-wise. It will cause backup files to grow endlessly (more than they already are).

You should expire and remove data after some time (max a month). It is unlikely that the user even wants to keep it longer than that, especially with sources changing so much.

fire-light42 (changes requested):
First review, what I had time for right now.

Generally the code is way too complex for something so simple. It should be easy to read and follow along, and understand how it integrates with the current automatic subtitle selection system (`preferredAutoSelectSubtitles`) and automatic priority source selection system. 

Use comments to explain the intended behavior of complex code.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -2075,106 +2259,15 @@ class GeneratorPlayer : FullScreenPlayer() {
...
-
-            // If the file is downloaded then do not select auto select the subtitles
-            // Downloaded subtitles cannot be selected immediately after loading since
-            // player.getCurrentPreferredSubtitle() cannot fetch data from non-loaded subtitles
-            // Resulting in unselecting the downloaded subtitle
-            if (set.lastOrNull()?.origin != SubtitleOrigin.DOWNLOADED_FILE) {
-                autoSelectSubtitles()
-            }
-        }
+        observeLoadingLinks()
```
fire-light42, line 2377: Please refrain from refactoring existing code unless required.
Unnecessary changes makes it more difficult to review and track down bugs, and it creates merge conflicts.

It is much better to keep each pull request scoped to solve one specific issue.

app/src/main/java/com/lagradost/cloudstream3/ui/player/EpisodePreferenceHelper.kt
```diff
@@ -0,0 +1,262 @@
...
+     */
+    fun getEpisodePreference(meta: Any?): EpisodePlaybackPreference? {
+        val parentId = getEpisodePreferenceParentId(meta) ?: return null
+        val preference = getKey<EpisodePlaybackPreference>(
+            "$currentAccount/$PLAYER_EPISODE_PREFERENCES",
+            parentId.toString()
+        ) ?: return null
+
+        // Expire preferences older than 30 days
+        if (System.currentTimeMillis() - preference.savedAt > PREFERENCE_MAX_AGE_MS) {
```
fire-light42, line 186: This does not delete any saved data.

The core issue is that we don't want user preferences to grow forever. 
Therefore we must clean unwanted data, not just return null.

I suggest you place a removeKey here and then make the clear cache button in settings also clear all redundant keys. That way we can make it possible for users to clean this themselves, without looking through all keys all the time to check expiration date.

app/src/main/java/com/lagradost/cloudstream3/ui/player/EpisodePreferenceHelper.kt
```diff
@@ -0,0 +1,262 @@
...
+            subtitle.originalName == preference.subtitleOriginalName &&
+                    subtitle.url == preference.subtitleUrl
+        }?.let { subtitle ->
+            return ResolvedEpisodeSubtitlePreference(
+                subtitle = subtitle,
+                blockFallback = false,
+            )
+        }
+
+        // Priority 2: Match by originalName + nameSuffix (cross-episode, preserves "Turkish 3" selection)
```
fire-light42, line 105: Do not use nameSuffix for any fallbacks, it could lead to incorrect automatic subtitle selection, which would mislead the user.

Only use url, name and language.

app/src/main/java/com/lagradost/cloudstream3/ui/player/EpisodePreferenceHelper.kt
```diff
@@ -0,0 +1,262 @@
...
+     * Matching priority:
+     * 1. Exact match by originalName + URL (most stable, same episode re-selection)
+     * 2. Match by originalName + nameSuffix (cross-episode: e.g. "Turkish 3" stays "Turkish 3")
+     * 3. Match by originalName only (fallback if suffix not found)
+     * 4. Match by languageTag + nameSuffix (fallback with suffix preference)
+     * 5. Match by languageTag only (broadest fallback)
+     *
+     * If subtitles were explicitly disabled, returns blockFallback=true
+     * to prevent auto-selection from overriding the user's choice.
+     */
```
fire-light42, line 73: Great comments!

app/src/main/java/com/lagradost/cloudstream3/ui/player/EpisodePreferenceHelper.kt
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Maximum age for episode preferences in milliseconds (30 days).
+ * Preferences older than this are automatically expired and removed
+ * to prevent backup files from growing endlessly.
+ */
+private const val PREFERENCE_MAX_AGE_MS = 30L * 24 * 60 * 60 * 1000
+
+internal data class EpisodePlaybackPreference(
+    val sourceDisplayName: String? = null,
```
fire-light42, line 21: Please use annotations, e.g., `@JsonProperty("sourceDisplayName")` before each variable in objects you are storing.

This is to signify to the reader that the content is going to be serialized and to ensure that no issues occur if someone renames the variables in the future.

PR #2582, opened 2026-03-28
Bump DGP and KGP libs

Luna712:
Final compatibility with AGP 9 (though noting it may be awhile before we can fully upgrade to AGP 9, my PRs to recloudstream/gradle are still needed as well for extensions among a few other things there).


Luna712:
@fire-light42 this *should* be okay to merge btw. I did test Dokka as well, but there is a small likelihood Dokka could still have issues. If so we should be able to easily fix it though.

PR #2584, opened 2026-03-29
Quality: Subtitle deletion matches on substring extension, can delete non-subtitle files

Nam0101:
## Problem

`isMatchingSubtitle` checks extensions with `name.contains(...)` instead of checking the actual file suffix. This can classify non-subtitle files as subtitles (e.g. `movie.srt.mp4` contains `.srt`) and delete them in `deleteMatchingSubtitles`, causing real data loss.

**Severity**: `high`
**File**: `app/src/main/java/com/lagradost/cloudstream3/utils/SubtitleUtils.kt`

## Solution

Match only by true file extension. Replace the extension check with `endsWith` on a normalized filename: `val lower = name.lowercase()` `val hasValidExtension = allowedExtensions.any { lower.endsWith(it) }` Keep deletion gated on this stricter check.

## Changes

- `app/src/main/java/com/lagradost/cloudstream3/utils/SubtitleUtils.kt` (modified)

## Testing

- [x] Existing tests pass
- [x] Manual review completed
- [x] No new warnings/errors introduced

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/SubtitleUtils.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/SubtitleUtils.kt
@@ -43,8 +43,7 @@ object SubtitleUtils {
         cleanDisplay: String
     ): Boolean {
         // Check if the file has a valid subtitle extension
-        val lowerName = name.lowercase()
-        val hasValidExtension = allowedExtensions.any { lowerName.endsWith(it) }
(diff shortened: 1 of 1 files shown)


fire-light42:
Nice catch! I doubt it has accidentally deleted anything, but it is a very unnecessary risk.

Remember to mark your pull requests as AI assisted as per our [AI Policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md) if you used AI for anything in the pull request.

PR #2588, opened 2026-03-30
Bump actions

Luna712:


.github/workflows/prerelease.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         echo "key_pwd=$KEY_PWD" >> $GITHUB_OUTPUT
 
     - name: Setup Gradle
-      uses: gradle/actions/setup-gradle@v5
+      uses: gradle/actions/setup-gradle@v6
```
fire-light42: Looks like v6 removed configuration cache (temporarily): https://blog.gradle.org/github-actions-for-gradle-v6. There could be value in waiting for v7 while staying on v5. I think our configuration cache works now.

What do you think?

Luna712: Very good point. I somehow missed that. We should probably wait yeah.

Luna712: I updated this to just bump the other two actions for now.

PR #2594, opened 2026-04-01
Dokka Documentation

jimlyas:
This PR include

## Dokka Documentation Generation

Removing `docs` module, Now Dokka documentation can be generated directly from the root project. 

```shell
# Before
./gradlew :docs:dokkaGeneratePublicationHtml

# After
./gradlew dokkaGeneratePublicationHtml
```

The generated documentation:

|Module|ScreenShoot|
|---|---|
|Root|All Module|
|App|App Module|
|Library|Library Module|

There's some package documentation named `packages.md` for `app` and `library` modules **generated by AI**.


fire-light42:
Great contributions! 

However please keep these two changes as separate pull requests to make it easier to review and merge accordingly. It also helps with future debugging and reverting.

Luna712:
Note to @fire-light42 before merging this I would *prefer* (though definitely up to you) this wait until after AGP 9, as I have spent months testing and working on AGP 9 changes including Dokka support and other stuff, I really dont want to have to rebegin testing compatibility with the merged projects and removal of docs which does change the functionality a bit. Again up to you though.

Overall this is a nice contribution though and this is not to say otherwise.

fire-light42:
Of course, those who have made pull requests should not worry about having to adapt to new changes due to a slow review process.

app/packages.md
```diff
@@ -0,0 +1,233 @@
...
+
+Extensions management:
+- ExtensionsFragment - Extensions UI
+- PluginsFragment, PluginDetailsFragment - Plugin management
+- ExtensionsViewModel, PluginsViewModel - State management
+
+# Package com.lagradost.cloudstream3.ui.settings.testing
+
+Testing utilities:
+- TestFragment - Testing UI
```
fire-light42, line 187: I am unsure about AI generated documentation for these things. 

It may look good at a glance, but it might not be very good long term. First of all it creates a maintenance task, which requires maintainers and contributors to keep this updated every time something is changed. 

Secondly, it is not very helpful. In this example we have "Testing UI", what does it mean? It tells nothing more than the file name currently. 

I think if we want more documentation about what files do we need to introduce a code standard wherein people use doc comments more and in an organized manner A workflow could then automatically generate appropriate documentation. This would make intuitive sense for contributors and be useful even if people only look at the code.

jimlyas: I fully agree, the AI-generated documentation is far from decent.

I just thought it might be a nice addition to the project's documentation, by having some definition of what does the package contains

PR #2597, opened 2026-04-02
Gradle Dependency Verification

### LINKED ISSUE #2589: CloudStream is missing dependency verification
Cloudstream is currently missing [dependency verification](https://docs.gradle.org/current/userguide/dependency_verification.html).

This would prevent supply chain attacks and add more trust in the build process. This is especially useful for us as we fully rely on automated builds for pre-release. Adding it would increase dependency management overhead but I feel like the increased security is worth the cost.

Should we add it? Opinions are welcome!

jimlyas:
This PR implement Gradle dependency verification.

Could probably close #2589


fire-light42:
@Luna712 👀

Luna712:
Thanks for rhe ping @fire-light42. I will verify this later today.

Great contribution though @jimlyas. Thanks for working on this.

**UPDATE** Sorry didnt get to this yet but will within the next few days.

PR #2598, opened 2026-04-02
fix(download): remove deleted/cancelled items from the Downloads list in real time (#1227)

### LINKED ISSUE #1227: Download does not get removed when cancelling from a notification.
### Steps to reproduce

This was found and explained in https://github.com/recloudstream/cloudstream/pull/1177. The main problem is that the UI is still showing episodes or movies that was deleted by the user if the user cancels an download. This is a preexisting bug, but #1177 made it worse by also affecting episodes instead of just movies. 

### Expected behavior

The downloadDeleteEvent or similar should be hooked to update the UI in case of deletion from any source.

### Actual behavior

Invalid UI 

ab379ab31c30069aac3afdeec76410e69ea6bc95

### Android version

N/A

### Logcat

_No response_

### Other details

_No response_

### Acknowledgements

- [X] I am sure my issue is related to the app and **NOT some extension**.

LeC-D:
Fixes #1227.

### Problem

When a download is cancelled or deleted, `VideoDownloadManager.downloadDeleteEvent` fires — but until now **only the download button** (`BaseFetchButton`) listened to it. The **Downloads page itself never reacted**, so a deleted episode/movie card stayed on screen until the user navigated away and back. That stale card is the actual bug reported in #1227 ("UI still showing episodes/movies that were deleted").

### Fix

**`DownloadViewModel`** — the core #1227 fix. Subscribes to `downloadDeleteEvent` at the activity-scoped (shared) ViewModel level and removes the deleted id from the header + child lists live, so both `DownloadFragment` (headers) and `DownloadChildFragment` (children) refresh with no manual reload. It also clears the id from multi-select state. Uses `postValue` only, since the event fires on the downloader's background thread. This hooks the event "to update the UI in case of deletion from any source" (result page, queue, notification, …), as the issue asks.

**`BaseFetchButton`** — resets the button via `resetView()` on a matching `persistentId`; subscription wired in `onAttached/DetachedFromWindow`, mirroring the existing `downloadStatusEvent` handling. This keeps an on-screen download button from showing a stale state after a deletion elsewhere.

### How to reproduce (the #1227 symptom)

1. Download one or more episodes/movies.
2. Open the **Downloads** page.
3. Cancel/delete an item from another surface (result page button, download queue, or the download notification).
4. **Before:** the deleted card stays visible in the Downloads list until you leave and re-enter the page. **After:** it disappears immediately.

I'm happy to attach a before/after screen recording if useful.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
@@ -68,6 +68,14 @@ class DownloadViewModel : ViewModel() {
     private val _selectedItemIds = ConsistentLiveData<Set<Int>?>(null)
     val selectedItemIds: LiveData<Set<Int>?> = _selectedItemIds
 
+    // Retained so onDownloadDeleted can rebuild the lists from disk after a
+    // delete triggered from anywhere in the app. We only ever store the
+    // application context (never an Activity) to avoid leaking it past the
+    // ViewModel's lifetime. currentChildFolder is the folder backing the child
+    // list currently on screen, or null when no child screen is open.
+    private var appContext: Context? = null
+    private var currentChildFolder: String? = null
+
     init {
         // Keep the Downloads list in sync when a download is deleted/cancelled from
         // anywhere in the app (result page button, queue, notification, etc.). See
@@ -206,6 +214,7 @@ class DownloadViewModel : ViewModel() {
     }
 
     fun updateHeaderList(context: Context) = viewModelScope.launchSafe {
+        appContext = context.applicationContext
         // Do not push loading as it interrupts the UI
         //_headerCards.postValue(Resource.Loading())
 
@@ -368,8 +377,16 @@ class DownloadViewModel : ViewModel() {
         })
     }
 
-    fun updateChildList(context: Context, folder: String) = viewModelScope.launchSafe {
-        _childCards.postValue(Resource.Loading()) // always push loading
+    fun updateChildList(
+        context: Context,
+        folder: String,
+        pushLoading: Boolean = true
+    ) = viewModelScope.launchSafe {
+        appContext = context.applicationContext
+        currentChildFolder = folder
+        // Push loading when first opening the screen, but not on a live refresh
+        // (e.g. after a delete) where clearing the list would flicker the UI.
+        if (pushLoading) _childCards.postValue(Resource.Loading())
 
         val visual = withContext(Dispatchers.IO) {
             context.getKeys(folder).mapNotNull { key ->
@@ -402,7 +419,7 @@ class DownloadViewModel : ViewModel() {
     }
 
     /**
-     * Removes a deleted/cancelled download from the currently displayed lists in real time.
+     * Refreshes the Downloads screen in real time when a download is deleted/cancelled.
      *
      * Issue #1227: cancelling a download (or deleting it from the result page, the download
      * queue, or a notification) fires [VideoDownloadManager.downloadDeleteEvent], but until
@@ -410,24 +427,34 @@ class DownloadViewModel : ViewModel() {
      * itself never reacted, so the deleted episode/movie card stayed on screen until the user
      * navigated away and back. We observe the same event here, at the activity-scoped (shared)
      * ViewModel level, so both [DownloadFragment] (headers) and [DownloadChildFragment]
-     * (children) refresh without a manual reload.
+     * (children) refresh.
      *
-     * The event is invoked synchronously on the downloader's background thread, so this only
-     * uses thread-safe LiveData postValue.
+     * We rebuild the lists from disk rather than filtering the in-memory lists by id. Filtering
+     * cannot handle a series correctly: deleting an *episode* fires the event with the child id,
+     * which must shrink (or remove) the parent *header* rather than match a header id, and the
+     * child list is frequently in the [Resource.Loading] state when the delete originates from
+     * outside the child screen. Reloading reads [VideoDownloadManager.getDownloadFileInfo],
+     * whose [KEY_DOWNLOAD_INFO] the delete has already cleared, so the gone episode/movie
+     * disappears and the header aggregates (download count, size) are recomputed.
+     *
+     * The event is invoked synchronously on the downloader's background thread; the reloads
+     * below hop onto [viewModelScope] and only post via thread-safe LiveData postValue.
      */
     private fun onDownloadDeleted(id: Int) {
-        // The id space is shared: movie/series headers use their own id and individual
-        // episodes use the child id, so filtering both lists by id covers every card type.
-        // filter() returns null only while the list is still Loading; postHeaders/postChildren
-        // treat null as "no change", so there is nothing to update in that case.
-        postHeaders(_headerCards.success?.filter { it.data.id != id })
-        postChildren(_childCards.success?.filter { it.data.id != id })
-
         // Keep multi-select state consistent: forget the removed id if it was selected.
         val currentSelection = selectedItemIds.value
         if (currentSelection?.contains(id) == true) {
             _selectedItemIds.postValue(currentSelection - id)
             updateSelectedBytes()
+        }
+
+        // Nothing has been shown yet (no list ever loaded), so there is nothing to refresh.
+        val context = appContext ?: return
+        updateHeaderList(context)
+        // Refresh the child screen too, without flashing a loading state. currentChildFolder is
+        // cleared once the child screen closes (clearChildren), so this is a no-op afterwards.
+        currentChildFolder?.let { folder ->
+            updateChildList(context, folder, pushLoading = false)
         }
     }
 
@@ -616,6 +643,7 @@ class DownloadViewModel : ViewModel() {
     }
 
     fun clearChildren() {
+        currentChildFolder = null
         _childCards.postValue(Resource.Loading())
     }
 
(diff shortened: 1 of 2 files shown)


fire-light42 (changes requested):
Please review our AI Policy https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md

LeC-D:
Hi @fire-light42, thanks for pointing to the AI policy.

To be transparent: this fix was developed with AI assistance (GitHub Copilot / Claude for context exploration). I'm fully responsible for understanding and verifying the code — the approach directly mirrors the existing `downloadStatusEvent` and `downloadProgressEvent` patterns already in `BaseFetchButton`, so the change is straightforward and intentional. I've reviewed and tested the logic manually.

Let me know if you have any concerns about the implementation itself!

LeC-D:
Updated the PR description to include the AI usage disclosure directly in the body, as required by the [AI Policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md). Thanks for flagging this, @fire-light42!

LeC-D:
Rebased on upstream `master` (2026-04-03) to bring the branch back up to date.

fire-light42:

I will test and review soon. The code looks good, but getting these things right can be tricky.

Do not concern yourself with merge conflicts or keeping the branch up to date, I will resolve any conflicts if necessary :+1:

fire-light43:
I can not replicate the stale download state. 
Can you please show a video of it before and after?

LeC-D:
Quick status update to clear up confusion about what this PR actually does — the original title/description undersold its scope.

**The code already covers the full #1227 fix** (2 files, the diff hasn't changed):

- **`DownloadViewModel`** — the actual #1227 fix: subscribes to `downloadDeleteEvent` at the activity-scoped (shared) ViewModel level and removes the deleted id from the header + child lists live, so both `DownloadFragment` and `DownloadChildFragment` refresh with no manual reload. Also clears it from multi-select state. `postValue` only (the event fires on the downloader's background thread).
- **`BaseFetchButton`** — resets the button via `resetView()` on a matching `persistentId`; subscription wired in `onAttached/DetachedFromWindow`, mirroring the existing `downloadStatusEvent` handling.

Together this hooks `downloadDeleteEvent` "to update the UI in case of deletion from any source," as the issue asks.

@fire-light43 — I think the earlier difficulty reproducing came from the old description, which pointed at the *button* rather than the *Downloads list*. The #1227 symptom is on the list. Repro:

1. Download some episodes/movies.
2. Open the **Downloads** page.
3. Cancel/delete an item from another surface (result page, queue, or notification).
4. Before: the deleted card stays in the list until you leave and re-enter. After: it disappears immediately.

I've updated the title and description to match. Happy to attach a before/after recording if that helps.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
```diff
@@ -389,6 +401,36 @@ class DownloadViewModel : ViewModel() {
...
+     * The event is invoked synchronously on the downloader's background thread, so this only
+     * uses thread-safe LiveData postValue.
+     */
+    private fun onDownloadDeleted(id: Int) {
+        // The id space is shared: movie/series headers use their own id and individual
+        // episodes use the child id, so filtering both lists by id covers every card type.
+        // filter() returns null only while the list is still Loading; postHeaders/postChildren
+        // treat null as "no change", so there is nothing to update in that case.
+        postHeaders(_headerCards.success?.filter { it.data.id != id })
+        postChildren(_childCards.success?.filter { it.data.id != id })
```
fire-light42: This does not work for TV shows because `_childCards` are 'Loading' when this gets triggered.

fire-light42 (changes requested):

LeC-D:
Thanks for taking the time to reproduce this and for the precise pointer — you were right. Filtering the in-memory lists by id was the wrong approach for TV shows, for the two reasons your comment exposed:

1. `_childCards` is often still in `Loading` when `downloadDeleteEvent` fires, so filtering `.success` was a no-op.
2. Deleting an episode emits the event with the **child** id, never the header id — so the series header could never be updated by an id filter.

I've reworked `onDownloadDeleted` (commit `b0513631`) to **rebuild the lists from disk** instead of filtering the cache: it re-runs `updateHeaderList` + `updateChildList`. After a deletion the `KEY_DOWNLOAD_INFO` entry is gone (or the file is removed), so `getDownloadFileInfo` returns `null`, the item drops out, and the header aggregates are recomputed. This works the same for a normal delete and for cancelling an in-progress download.

Implementation notes: I retain the application `appContext` (no Activity reference → no leak) and `currentChildFolder`, and added a `pushLoading=false` flag so the live refresh doesn't flash a Loading state.

Ready for another look when you have a moment 🙏

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
```diff
@@ -389,6 +418,46 @@ class DownloadViewModel : ViewModel() {
...
+     * The event is invoked synchronously on the downloader's background thread; the reloads
+     * below hop onto [viewModelScope] and only post via thread-safe LiveData postValue.
+     */
+    private fun onDownloadDeleted(id: Int) {
+        // Keep multi-select state consistent: forget the removed id if it was selected.
+        val currentSelection = selectedItemIds.value
+        if (currentSelection?.contains(id) == true) {
+            _selectedItemIds.postValue(currentSelection - id)
+            updateSelectedBytes()
+        }
```
fire-light42:
```suggestion
        updateSelectedItems { it?.minus(id) }
```
Same logic, much cleaner.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
```diff
@@ -389,6 +418,46 @@ class DownloadViewModel : ViewModel() {
...
+
+        // Nothing has been shown yet (no list ever loaded), so there is nothing to refresh.
+        val context = appContext ?: return
+        updateHeaderList(context)
+        // Refresh the child screen too, without flashing a loading state. currentChildFolder is
+        // cleared once the child screen closes (clearChildren), so this is a no-op afterwards.
+        currentChildFolder?.let { folder ->
+            updateChildList(context, folder, pushLoading = false)
+        }
```
fire-light42:
```suggestion
        postChildren(_childCards.success?.filterNot { it.data.id == id })
```
Unnecessary to save a folder or re-fetch children if you just need to delete ids.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
```diff
@@ -389,6 +418,46 @@ class DownloadViewModel : ViewModel() {
...
+        // Keep multi-select state consistent: forget the removed id if it was selected.
+        val currentSelection = selectedItemIds.value
+        if (currentSelection?.contains(id) == true) {
+            _selectedItemIds.postValue(currentSelection - id)
+            updateSelectedBytes()
+        }
+
+        // Nothing has been shown yet (no list ever loaded), so there is nothing to refresh.
+        val context = appContext ?: return
+        updateHeaderList(context)
```
fire-light42:
```suggestion
        val context = CloudStreamApp.context ?: return
        updateHeaderList(context)
```
We use the CloudStreamApp global context here, otherwise the context variable will cause memory leaks.

fire-light42 (changes requested):
It should be mergeable when the comments are addressed. 

If you have time it would be very nice if you could add a corresponding feature for when new downloads are started.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
LeC-D: Applied in 60cfebd4 — collapsed to `updateSelectedItems { it?.minus(id) }`.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
LeC-D: Applied in 60cfebd4 — now uses `CloudStreamApp.context`, and the retained `appContext`/`currentChildFolder` fields are gone entirely.

app/src/main/java/com/lagradost/cloudstream3/ui/download/DownloadViewModel.kt
LeC-D: Applied in 60cfebd4 — the child list now just drops the deleted id via `postChildren(_childCards.success?.filterNot { it.data.id == id })`, and the `pushLoading` parameter on `updateChildList` is reverted since nothing re-fetches the folder anymore. The header list is still rebuilt from disk so series counts/sizes stay correct when an episode (child id) is deleted.

LeC-D:
All three suggestions applied in 60cfebd4 (net effect: −31 lines vs the previous revision — `CloudStreamApp.context`, `updateSelectedItems { it?.minus(id) }`, and in-memory child filtering instead of the folder re-fetch, which also let me revert the `pushLoading` parameter).

On the follow-up idea — live-adding items when a new download *starts*: agreed it would round this out nicely. I'd rather do it as a separate PR than grow this one, for two reasons: it needs its own event wiring (`downloadStatusEvent` rather than `downloadDeleteEvent`, with care not to rebuild on every pause/resume transition), and I'd like this fix to stay small and easy to verify against #1227. Happy to open that follow-up once this lands, if that works for you.

fire-light42:
Perfect! I will merge after testing

fire-light42 (approved):

PR #2604, opened 2026-04-03
AGP 9!

Luna712:
After months of testing and numerous intervals of PRs and work on this we should finally be ready for AGP 9 with built-in Kotlin.


Luna712:
This *should* ***finally*** be ready now!

fire-light42 (approved):
Nice!

PR #2618, opened 2026-04-05
Trying Auto Skip

phisher98:
Trying to implement auto-skip; used AI for guidance.

Added a setting to disable it as well.


fire-light42:
Please check out the new API at https://github.com/recloudstream/cloudstream/commit/1d03b05a7cceeb9656d28f27227f957f8097251d

Luna712:
Also, if this is done, I think it would make more sense for consistency with current behavior to be disabled by default. Just a suggestion though but not to adamant about it.

phisher98:
Auto Skip Disabled by Default

fire-light42 (changes requested):
While I have no doubt this works fine, the logic is hard to follow given the indirect recursion and global variables. I think it would be much cleaner if this was a simple for loop inside a suspended context. Then you only need a single "Job" variable that can be canceled.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1997,76 @@ class GeneratorPlayer : FullScreenPlayer() {
...
 
     override fun onDestroyView() {
+        clearSkipChapterAutoClick()
         binding = null
         super.onDestroyView()
     }
 
     var skipAnimator: ValueAnimator? = null
     var skipIndex = 0
+    private var currentTimestamp: VideoSkipStamp? = null
```
fire-light42: This local variable is unnecessary, given that you can pass around it by value, and that skipIndex already keeps track when the stamp changes.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1997,76 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+    }
+
+    private fun updateSkipChapterButtonText(
+        timestamp: VideoSkipStamp? = null,
+        secondsRemaining: Int? = null
+    ) {
+        val displayTimestamp = timestamp ?: currentTimestamp
+
+        val text = displayTimestamp?.let { ts ->
+            if (isAutoSkipPopupEnabled() && secondsRemaining != null) {
```
fire-light42: Simplify to `secondsRemaining != null` given that it should only be called with a non-null value whenisAutoSkipPopupEnabled.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1997,76 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+    private fun isAutoSkipPopupEnabled(): Boolean {
+        val ctx = context ?: return false
+        return PreferenceManager.getDefaultSharedPreferences(ctx).getBoolean(
+            ctx.getString(R.string.auto_skip_popup_key),
+            false
+        )
+    }
+
+    private fun updateSkipChapterButtonText(
+        timestamp: VideoSkipStamp? = null,
```
fire-light42: Require this paramater to be non-null.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1997,76 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+                    if (skipIndex != currentIndex || currentTimestamp != timestamp) return@launch
+
+                    updateSkipChapterButtonText(timestamp, secondsRemaining)
+                    delay(1000)
+                }
+
+                if (isActive && skipIndex == currentIndex && currentTimestamp == timestamp) {
+                    player.handleEvent(CSPlayerEvent.SkipCurrentChapter)
+                }
+            } finally {
```
fire-light42, line 2056: 1. You need to catch the `CancellationException`
2. Only call clearSkipChapterAutoClick if you are `isActive`

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1997,76 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+        playerBinding?.skipChapterButton?.setText(text)
+    }
+
+    private fun clearSkipChapterAutoClick() {
+        skipAutoClickJob?.cancel()
+        skipAutoClickJob = null
+    }
+
+    private fun startSkipChapterAutoClick(timestamp: VideoSkipStamp, currentIndex: Int) {
+        if (skipAutoClickJob?.isActive == true) {
```
fire-light42: This should not be the case, given the previous popup will cancel itself when it notices that the index is incorrect. So I think you should remove this line.

fire-light42 (changes requested):
Good changes, but I think you can simplify this even more! ❤️

phisher98:
Done

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1998,70 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+
+                    updateSkipChapterButtonText(timestamp, secondsRemaining)
+                    delay(1000)
+                }
+
+                if (isActive && skipIndex == currentIndex) {
+                    player.handleEvent(CSPlayerEvent.SkipCurrentChapter)
+                }
+            } catch (e: CancellationException) {
+                throw e
```
fire-light42: No, you should not just rethrow it! That will crash the entire app. Write:
```kotlin
} catch (_: CancellationException) {
// If we get canceled, do nothing. 
}
```

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -2029,13 +2085,11 @@ class GeneratorPlayer : FullScreenPlayer() {
                 addListener(onEnd = {
                     if (show) {
                         if (!isShowing) {
-                            // Automatically request focus if the menu is not opened
```
fire-light42: Why remove comments?

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1996,12 +1998,70 @@ class GeneratorPlayer : FullScreenPlayer() {
...
+    }
+
+    private fun startSkipChapterAutoClick(timestamp: VideoSkipStamp, currentIndex: Int) {
+        clearSkipChapterAutoClick()
+        if (!isAutoSkipPopupEnabled()) {
+            updateSkipChapterButtonText(timestamp)
+            return
+        }
+
+        skipAutoClickJob = viewModel.viewModelScope.launch {
```
fire-light42: Use launchSafe to avoid any sort of crash.

fire-light42 (changes requested):
Good improvements, however while testing it I found that it is impossible to cancel it manually. Normally you either have a 5s countdown with a cancel button, or auto skip instantly. TV devices often cancel it immediately on any key-press. 

So my question is: Is this the intended experience?

phisher98:
> Good improvements, however while testing it I found that it is impossible to cancel it manually. Normally you either have a 5s countdown with a cancel button, or auto skip instantly. TV devices often cancel it immediately on any key-press.
> 
> So my question is: Is this the intended experience?

Yes I guess do we need that ?

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
phisher98: oops mistakenly added back

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
phisher98: Done

fire-light42:
What I mean is that autoskip after 5s is a bit weird from a UX perspective. Here is an image of what I think autoskip should act like on TV, given my experience with the feature from other services.

Capture

Where as phone has a similar cancel when you press the screen. Otherwise what is the point of a 5s timer?

PR #2627, opened 2026-04-07
Fix livestreams

CranberrySoup:
- Fixes the issue of live streams closing if you seek too far (PlaylistStuckException).
- Removes previews from live streams (they get outdated)
- Changes seekbar to display as full when you are sufficiently live.
- Seeking beyond the optimal live point makes the player seek back to the optimal live point

Knowing when the user is live is actually difficult, because most livestreams are only a sliding window of 10-30 second media segments. I therefore have to calculate where the optimal player point is based on new video chunks and their durations. 

From my testing it works well on all livestreams, with the caveat of the seekbar not being very useful in short live streams due to the auto-seek. I did not see a good way to circumvent this, and the seekbar is currently useless in short live streams regardless.

This could potentially fail on specially crafted live streams, but it would at worst make the video unseekable.


app/src/main/java/com/lagradost/cloudstream3/ui/player/live/LiveManager.kt
```diff
@@ -0,0 +1,97 @@
...
+const val CHUNK_VARIANCE = 3000L
+
+// A livestream chunk from the player, the time we get it and the duration can be used to calculate
+// the expected live timestamp.
+class LivestreamChunk(
+    durationMs: Long, val receiveTimeMs: Long = System.currentTimeMillis()
+) {
+    // We want to be PREFERRED_LIVE_OFFSET ms after the latest update, but we cannot be ahead of the middle point.
+    // If we are ahead of the middle point we will reach the end before the new chunk is expected to be released.
+    val targetPosition = maxOf(0,minOf(
```
CranberrySoup, line 26: This could be calculated using an average of several chunks. It would increase accuracy by a bit but increase complexity.

app/src/main/java/com/lagradost/cloudstream3/ui/player/live/LiveHelper.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.annotation.OptIn
+import androidx.media3.common.Player
+import androidx.media3.common.Timeline
+import androidx.media3.common.util.UnstableApi
+import com.lagradost.cloudstream3.mvvm.debugWarning
+import java.util.WeakHashMap
+
+object LiveHelper {
+    private val liveManagers = WeakHashMap<Player, LiveManager>()
+    private val listeners = WeakHashMap<Player, Player.Listener>()
```
fire-light42: I think this would make more sense as a Pair, to make them synced.

fire-light42 (approved):

PR #2630, opened 2026-04-08
Promote AcraApplication deprecations to error level

Luna712:
I know this is really early, but I personally PRed replacements for these to every single repo that used AcraApplication and listed in recloudstream/cs-repos. Therefore to make it less likely something starts using it again, I would prefer to do this earlier now, since it only affects build, not runtime.

If you decide you don't want to yet that is understandable, and this can be closed.


fire-light42:
These are good changes, but I still want at the very least a week of waiting, maybe even a month before marking as errors in builds. 

Changes from upstream which force abrupt code changes are very irritating and I really want to lessen that feeling as much as possible. Making pull requests to various repositories to address this is great!

I think the risk of new AcraApplication usage after the deprecation is less than the risk/irritation of abrupt changes, which is why I want to wait so long.

With the now large backlog of pull requests I think this can stay opened while I review them.

Luna712:
That is absolutely fair and makes sense.

PR #2631, opened 2026-04-09
Feat: TheIntroDBSkip + Bugfix

fire-light42:
This was a quick 20min coding adventure implementing #2622 Also fixed some issues I found in the existing API.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/SkipAPI.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/SkipAPI.kt
@@ -17,7 +17,8 @@ enum class SkipType(@StringRes val res: Int) {
     MixedOpening(R.string.skip_type_mixed_op),
     MixedEnding(R.string.skip_type_mixed_ed),
     Credits(R.string.skip_type_creddits),
-    Intro(R.string.skip_type_creddits),
+    Intro(R.string.skip_type_intro),
+    Preview(R.string.skip_type_preview),
 }
 
 data class SkipStamp(
@@ -60,7 +61,7 @@ abstract class SkipAPI {
     }
 
     companion object {
-        private val skipApis: List<SkipAPI> = listOf(AniSkip(), IntroDbSkip())
+        private val skipApis: List<SkipAPI> = listOf(AniSkip(), TheIntroDBSkip(), IntroDbSkip())
         private val cachedStamps = ConcurrentHashMap<Int, List<VideoSkipStamp>>()
 
         /** Get all video timestamps from an episode */
--- a/app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/TheIntroDBSkip.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/TheIntroDBSkip.kt
@@ -2,12 +2,14 @@ package com.lagradost.cloudstream3.utils.videoskip
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.lagradost.cloudstream3.LoadResponse
+import com.lagradost.cloudstream3.LoadResponse.Companion.getImdbId
 import com.lagradost.cloudstream3.LoadResponse.Companion.getTMDbId
 import com.lagradost.cloudstream3.LoadResponse.Companion.isMovie
 import com.lagradost.cloudstream3.TvType
 import com.lagradost.cloudstream3.ui.result.ResultEpisode
 import com.lagradost.cloudstream3.app
 
+/** https://theintrodb.org/docs */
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/TheIntroDBSkip.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.lagradost.cloudstream3.LoadResponse.Companion.getTMDbId
+import com.lagradost.cloudstream3.LoadResponse.Companion.isMovie
+import com.lagradost.cloudstream3.TvType
+import com.lagradost.cloudstream3.ui.result.ResultEpisode
+import com.lagradost.cloudstream3.app
+
+/** https://theintrodb.org/docs */
+class TheIntroDBSkip : SkipAPI() {
+    override val name = "TheIntroDB"
+    override val supportedTypes = setOf(TvType.TvSeries, TvType.Cartoon, TvType.Anime, TvType.Movie)
```
Luna712: Also supports AsianDrama.

app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/TheIntroDBSkip.kt
```diff
@@ -0,0 +1,67 @@
...
+        val recap: List<Stamp> = emptyList(),
+        @JsonProperty("credits")
+        val credits: List<Stamp> = emptyList(),
+        @JsonProperty("preview")
+        val preview: List<Stamp> = emptyList(),
+    )
+
+    data class Stamp(
+        @JsonProperty("start_ms")
+        val startMs: Long,
```
Luna712: Both this and endMs can also be null. If startMs is null that is 00:00, if endMs is null, that is the very end of the video. If I am understanding the API correctly anyway.

app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/TheIntroDBSkip.kt
```diff
@@ -0,0 +1,76 @@
...
+                ?: data.getImdbId()?.let { imdbId -> "imdb_id=$imdbId" }
+                ?: return null
+
+        val url = if (data.isMovie()) {
+            "$mainUrl/v2/media?$idSuffix"
+        } else {
+            val season = episode.season ?: return null
+            "$mainUrl/v2/media?$idSuffix&season=$season&episode=${episode.episode}"
+        }
+        val root = app.get(url).parsed<Root>()
```
Luna712, line 38: We don't seem to in other parts of this SkipAPI (and elsewhere also; but maybe we should?) but wouldnt this be better/safer to use parsedSafe? I am not positive though.

Luna712: Seems like maybe not necessary since the main SkipAPI calls stamps method using safe handling actually though it may still help track down potential issues via logcat if it uses parsedSafe so it logs where parsed in particular failed though may not be necessary.

app/src/main/java/com/lagradost/cloudstream3/utils/videoskip/SkipAPI.kt
```diff
@@ -60,7 +61,7 @@ abstract class SkipAPI {
     }
 
     companion object {
-        private val skipApis: List<SkipAPI> = listOf(AniSkip(), IntroDbSkip())
+        private val skipApis: List<SkipAPI> = listOf(AniSkip(), TheIntroDBSkip(), IntroDbSkip())
```
Luna712, line 64: Hmm. Probably not for this PR in particular, but something I noticed is IntroDB is more particular about what it has, TIDB has more an often more accurate, but some things have start timestamps on TIDB but not IntroDB while having end timestamps on IntroDB but not TIDB. I wander if there is a way to mix what API it can get for different SkipStamp types...

Otherwise perhaps IntroDB should be tried first here since its smaller what it does have may in fact be better than TIDB at times...?

Luna712:
Not sure if related to this (or SkipAPI at all) or not but when testing, and fast forwarding and rewinding a lot I got a random OOM for the first time ever in this context.
Screenshot_20260409_104719_CloudStream Debug(1)

Luna712 (approved):
Works very well on what I tested (I tested on ones that have all the timestamps (Except Preview,  one probably exists but I couldnt fine any), and tested on at least one for: TV Series, Anime, Movie, and AsianDrama versions). Only those other suggestions to think about for the future. I am not convinced that OOM was related to this or the SkipAPI at all, it hasnt happened since anyway.

**UPDATE** Found one with Preview and it works as well.

PR #2644, opened 2026-04-11
Add plugin hash validation

fire-light42:
Fixes a long standing issue of corrupted plugin downloads whenever the server host does not cooperate. This problem usually occurs when GitHub is rate limiting users.

Testable with the official repository: https://github.com/recloudstream/extensions which already has plugin hashes.
More repositories should automatically get plugin hashes when they use the latest gradle version.


app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -92,7 +100,8 @@ data class PluginData(
             null,
             null,
             null,
-            File(this.filePath).length()
+            File(this.filePath).length(),
+            getFileHash()
```
fire-light42: Not happy doing IO whenever we want to create a SitePlugin from a local plugin. Not sure what a better alternative would be.

fire-light42:
@Luna712 Might be interested, reviews are much appreciated.

Luna712 (commented):
I don't really have the knowledge to review this in to much more detail. I mean, it looks fine to me, but you are better than I am at this lol...

fire-light42:
No worries :+1:

PR #2653, opened 2026-04-13
Change default installer to legacy

fire-light42:
I have seen people report issues with the default APK installer. We need to ensure that the "new" installer is fully reliable.

I am switching the default to the more reliable legacy installer until we have fixed the new installer.


Luna712:
The new installer still works for me. But yeah I suppose this is a good idea.

ByAyzen:
I saw a lot of users who had problems with new method, thats happening especially on tv boxes and older devices.

PR #2654, opened 2026-04-13
Add permissions to workflows

Luna712:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

This was found by codeql, its not really an issue for us here but it is still probably a good idea to make explicit. This was fixed by the CodeQL auto fix which is technically GitHub Copilot AI, but I also made changes to it to fix the permissions it actually needs. This should cover all that it actually needs, but if it is missing something that should be very easy to fix.

### DIFF
--- a/.github/workflows/build_to_archive.yml
+++ b/.github/workflows/build_to_archive.yml
@@ -9,6 +9,9 @@ on:
       - '**/wcokey.txt'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 concurrency:
   group: "Archive-build"
   cancel-in-progress: true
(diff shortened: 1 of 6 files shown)


fire-light42:
Does the git push workflows not require `contents: write`?
For example: `update_locales.yml`

Luna712:
> Does the git push workflows not require `contents: write`?

They shouldnt because it uses the different GitHub app token not `GITHUB_TOKEN`. I couldn't test those ones though. But its a very easy fix if they end up needing it so left them out for now.

fire-light42 (approved):

PR #2655, opened 2026-04-13
Fix SSL issues

fire-light42:
We want to check all SSL traffic to prevent security issues and we want it to be secure by default. This *can* break some existing extensions, but the change is worth it to ensure everyone stays secure by default.

Extension developers can opt out by using `insecureApp`, which comes with an appropriate warning.
I think it would be best to merge this quickly and publish the April stable release shortly after, to maximize security and minimize extension breakage. 

Thoughts @LagradOst @Luna712 ?

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -1169,8 +1169,9 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        app.initClient(this)
-        val settingsManager = PreferenceManager.getDefaultSharedPreferences(this)
+        app.initClient(this, ignoreSSL = false)
+        @OptIn(UnsafeSSL::class)
+        insecureApp.initClient(this, ignoreSSL = true)
 
         setLastError(this)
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1169,8 +1169,9 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        app.initClient(this)
-        val settingsManager = PreferenceManager.getDefaultSharedPreferences(this)
```
Luna712, line 1173: Was this removal intentional?

Luna712: Overall looks good to me though. And yes should probably publish release earlier this time.

fire-light42: My mistake, it was not intentional :+1:

Luna712 (approved):

LagradOst (approved):

Luna712:
@fire-light42 just saw something else when working on a player rewrite... https://github.com/recloudstream/cloudstream/blob/636d2507f72ce571b37bce8c30766658309074c0/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt#L1877-L1887 does something need done there also? As it seems like it always true (to ignore ssl there also)

fire-light42:
> @fire-light42 just saw something else when working on a player rewrite...
> 
> https://github.com/recloudstream/cloudstream/blob/636d2507f72ce571b37bce8c30766658309074c0/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt#L1877-L1887
> does something need done there also? As it seems like it always true (to ignore ssl there also)

The problem with ignoreSSL is that someone can potentially do a mitm attack, and that is problematic when sending sensitive information or fetching sensitive content. 

However the attack surface from playing a video is not as critical, given that the client only consumes a video from the server. This should not lead to any remote code execution or leak too sensitive information. As such is a known "issue", we ignore to play all videos. Nonetheless, if someone has a real problem with this we will remove it.

PR #2656, opened 2026-04-14
Always use `exitProcess()` to exit the app

Luna712:
Even on phones, when the app enters safe mode and the user clicks the back button to exit, it does not exit safe mode upon restarting if the "confirm exit" setting is enabled and they exit from the dialog. Always using `exitProcess()` fixes this.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -674,9 +674,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
         val confirmBeforeExit = settingsManager.getInt(getString(R.string.confirm_exit_key), -1)
 
         if (confirmBeforeExit == 1 || (confirmBeforeExit == -1 && isLayout(PHONE))) {
-            // finish() causes a bug on some TVs where player
-            // may keep playing after closing the app.
-            if (isLayout(TV)) exitProcess(0) else finish()
+            exitProcess(0)
             return
         }
 
@@ -692,9 +690,8 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
                         putInt(getString(R.string.confirm_exit_key), 1)
                     }
                 }
-                // finish() causes a bug on some TVs where player
-                // may keep playing after closing the app.
-                if (isLayout(TV)) exitProcess(0) else finish()
+
+                exitProcess(0)
             }
 
         builder.show().setDefaultFocus()


app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -674,9 +674,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
         val confirmBeforeExit = settingsManager.getInt(getString(R.string.confirm_exit_key), -1)
 
         if (confirmBeforeExit == 1 || (confirmBeforeExit == -1 && isLayout(PHONE))) {
-            // finish() causes a bug on some TVs where player
-            // may keep playing after closing the app.
-            if (isLayout(TV)) exitProcess(0) else finish()
+            exitProcess(0)
```
fire-light42, line 677: exitProcess kills the entire app immediately, which includes background tasks (such as downloading) and the offline video player. There is not even a closing animation which looks very weird to users. We should really avoid this whenever possible. Finish is the normal android lifecycle call, which lets the OS handle the app appropriately. 

I think the best compromise would be calling exitProcess if the user is in safe mode.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
Luna712: finish() is unreliable, the app used to use exitProcess everywhere but was changed. even on phone ive noticed with finish() it will keep playing after closing sometimes too, but I kinda ignored it because it wasn't all that common. But with the safe mode issue also I figured it may also be a good idea to deal with it. Not sure what would fully be a good solution to fix all the potential issues here without fully killing it then.

fire-light42: "keep playing" as in the video audio is continuing without the app being open? We have had many issues on firestick with this (hence `if isLayout(TV)`), but I have never heard about it on phone.

Luna712: Yes I encountered it on phone twice before. Which wasn't much compared to how TVs were and couldn't reliably reproduce it either. Although looking into it more now, maybe `finishAffinity()` is the proper usage here?

Luna712: Or maybe `finishAndRemoveTask()`?

fire-light42: I do not know the exact differences between finish, finishAffinity and finishAndRemoveTask, but one of them is probably correct. We want to exit gracefully and without interrupting downloads.

fire-light42: But perhaps it is impossible to keep downloads while interrupting the player. :thinking:

Luna712: I will look into this more in depth later.

PR #2658, opened 2026-04-14
Bump nextlibMedia3

Luna712:


Luna712:
@fire-light42 this should probably be merged to sync it with the media3 dependancy btw. The only change in this version of nextlib was bumping the media3 requirement but they forgot to publish to maven until recently so it didn't get in my media3 bump.

PR #2670, opened 2026-04-15
Bump compilesdk to 37

Luna712:
37 *shouldn't* have any changes that will break anything here. But still going to split compileSdk and targetSdk out.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -50,7 +50,7 @@ zipline = "1.24.0"
 jvmTarget = "1.8"
 jdkToolchain = "17"
 minSdk = "23"
-compileSdk = "36"
+compileSdk = "37"
 targetSdk = "36"
 
 [libraries]


Luna712:
```
* What went wrong:
Execution failed for task ':app:checkPrereleaseDebugAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
   > An issue was found when checking AAR metadata:
     
       1.  Dependency 'androidx.biometric:biometric:1.4.0-alpha06' requires libraries and applications that
           depend on it to compile against version 36.1 or later of the
           Android APIs.
     
           :app is currently compiled against android-36.
     
           Recommended action: Update this project to use a newer compileSdk
           of at least 36.1, for example 37.
     
           Note that updating a library or application's compileSdk (which
           allows newer APIs to be used) can be done separately from updating
           targetSdk (which opts the app in to new runtime behavior) and
           minSdk (which determines which devices the app can be installed
           on).
```

Seems we may need this even sooner than I thought? Which is confusing why I get that when building the current version of the app just now and why it just started... we do use `androidx.biometric:biometric:1.4.0-alpha06` as the current version though.

**UPDATE** this was actually my own issue and not an issue with not doing this. Though there probably still is no harm in this whenever but it isn't 100% urgent.

fire-light42 (approved):
Works well when testing, targetSdk is no doubt more difficult to change.

PR #2687, opened 2026-04-17
Bump targetsdk to 37

Luna712:


Luna712:
@fire-light42 btw, as far as I could find there wasn't really any breaking changes with this that changes the behavior away from what we already use unless I did miss something.

PR #2689, opened 2026-04-17
Major rework to player

Luna712:
This practically rebuilds the entire player (sans IPlayer/CS3IPlayer), aimed to remove the FullScreenPlayer inheritance from ResultFragmentPhone, use BaseFragment for ResultFragmentPhone and the player, and make the player easier to maintain and expand in the future, as well overall cleanup to code readability and adding documentation to methods in the player. Also aimed to not fully crash if the layouts have minor differences as well.

Eventually FullScreenPlayer could be entirely factored out into helper methods for GeneratorPlayer, and AbstractPlayerFragment removed completely, but to reduce the chance of harder to resolve merge conflicts with removed files, I decided to not do it in this PR.


Luna712:
This is mostly ready for review. Though there may still be a few bugs since I wrote a lot of the code from scratch in many places (some of the code is stuff I worked on ~6 months ago, for about two months but gave up as I ran into some issues I couldn't figure but finally finished it and expanded to all the new features since then now).

If you feel this should be done a different way I could do that also, but this was just designed to make it easier in the future.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -2120,17 +2092,14 @@ class GeneratorPlayer : FullScreenPlayer() {
...
-                var lastTopIndex = -1
+                val lastTopIndex = -1
                 playerEpisodeList.addOnScrollListener(object : RecyclerView.OnScrollListener() {
-                    @SuppressLint("SetTextI18n", "DefaultLocale")
                     override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
                         val layoutManager =
                             recyclerView.layoutManager as? LinearLayoutManager ?: return
                         val topIndex = layoutManager.findFirstCompletelyVisibleItemPosition()
                         if (topIndex != RecyclerView.NO_POSITION && topIndex != lastTopIndex) {
-                            lastTopIndex = topIndex
```
fire-light42, line 2136: Bug?

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -79,676 +31,117 @@ const val NEXT_WATCH_EPISODE_PERCENTAGE = 90
...
-    var piphide: FrameLayout? = null
-    var subtitleHolder: FrameLayout? = null
-    var currentPlayerStatus = CSPlayerLoading.IsBuffering
+    // Stored pre-initialization so subclasses can set them before onBindingCreated.
+    private var _player: IPlayer = CS3IPlayer()
 
-    @LayoutRes
-    protected open var layout: Int = R.layout.fragment_player
+    /** The shared [PlayerView] host that owns all player state and view references. */
+    protected lateinit var playerHostView: PlayerView
```
fire-light42: You use playerHostView like a nullable with checks, but mark it as lateinit? Why not just mark this as nullable, and force the check it with the `?` operator.

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerView.kt
```diff
@@ -0,0 +1,788 @@
...
+        }, delayMs)
+    }
+
+    /** Cancels any pending auto-hide scheduled by [scheduleAutoHide]. */
+    fun cancelAutoHide() {
+        autoHideToken++
+        autoHideHandler.removeCallbacksAndMessages(null)
+    }
+
+    companion object {
```
fire-light42: Normally companion object is in the top of the file for readability.

fire-light42 (commented):
I have not looked in-depth at each line yet, given the sheer magnitude of this diff. It is hard to keep track of what is refactored, moved or unchanged. However, I like what I see so far.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
Luna712: It was unused at that point and lint in Android Studio was just warning about it which is why it was removed.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: That was just a design decision. Probably a poor design decision though and you make a good point.

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerView.kt
Luna712: Right my bad, will do. Thanks!

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: Hmm, actually I think I did this to make it more explicit with checks using isInitialized instead which I figured would be easier debugging if the need arises then tracing where it is becoming null, it would be essier to trace if its initialized at all, I did write this part of the code 6 months ago though. I can change it if you want though? Still not sure if it was the best choice or not myself though, so if you do think otherwise I will definitely change it.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
fire-light42: It is not, read the line above this. Android Studio has many bugs in its linter that other languages dont.

fire-light42: Here is an example where my android studio gives `Assigned value is never read` on `value = true`

```kotlin
var value = false
arrayOf(1).flatMap { _ ->
    value = true // false positive gives `Assigned value is never read` 
    listOf(1)
}
print(value)
```

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
fire-light42: The problem is that other people will use playerHostView without checking if it is isInitialized. You can easily add a getter and setter to trace if you want.

Luna712: Absolutely fair enough.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
Luna712: Didn't seem to cause any issue for me but I restored it now.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: Done

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerGestureHelper.kt
```diff
@@ -0,0 +1,1137 @@
...
+               (if (rmin >= 0 || h >= 0) forceLetters(rmin) + ":" else "") +
+               forceLetters(rsec)
+    }
+
+    /** Touch gestures */
+
+    fun setupTouchGestures() {
+        val holder = playerView.playerHolder ?: return
+        @SuppressLint("ClickableViewAccessibility")
+        holder.setOnTouchListener { v, event -> handleGesture(v, event) }
```
fire-light42: `holder.setOnTouchListener(::handleGesture)`

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerGestureHelper.kt
```diff
@@ -0,0 +1,1137 @@
...
+     * @param event              The motion event (should have pointerCount >= 2 or [lastPan] != null).
+     * @param ctx                Context used to create the [ScaleGestureDetector] on first call.
+     * @param onFirstPointerDown Called on [MotionEvent.ACTION_POINTER_DOWN] (e.g. hide player UI).
+     * @param onGestureEnd       Called when the gesture ends (e.g. reset caller touch state).
+     * @return Always true (event consumed).
+     */
+    fun handleZoomPanGesture(
+        event: MotionEvent,
+        ctx: Context,
+        onFirstPointerDown: () -> Unit = {},
```
fire-light42: No need to default it to `{}` when you have one callee.

fire-light42 (changes requested):
Very good pull request! I did not find any big glaring issues in the code or when testing.

The only bugs I found was:
1. player_video_title_rez is not hidden when locked
2. Brightness and volume sliders are still accessible when locked 
3. player_metadata_scrim is visible on trailers on phone

Please also note that while your formatting is great, it is not standard to android studio, and will cause the next person to touch the file to get +-500. Please format it with android studio or figure out some setting to disallow it on these files. 

I have also noticed that you removed a lot of comments, I am fine if you reword it. However, functions like `verifyVolume` lose context of *why* it is done, similar for `currentZoomMatrix`, and for the comments in `createScaleGestureDetector`. You should never remove a "why" comment, as it makes it harder to read the code later and understand the decision, intent, and why something else is not done. 

To be clear, doc comments that explain the "what" is done, e.g. functionally is entirely free to update, as they do not contain any inherit information that can not be gleaned from the code. 

While the clean code book is not perfect, and has a lot of misguided advice, the comments sections is worth reading:  https://bitsyncvortex.github.io/Books/pdfs/ComputerScience/CleanCode.pdf page 55+

Luna712:
Thank you for the review! I will try and finish this up tomorrow.

Luna712:
Is "player_video_title_rez is not hidden when locked" an issue on pre as well? I don't see where it hides on pre either. I can fix it here though anyway and may be just missing it.

Also not sure what you mean by "Please also note that while your formatting is great, it is not standard to android studio, and will cause the next person to touch the file to get +-500. Please format it with android studio or figure out some setting to disallow it on these files."

Because I wrote it within Android Studio in the first place and it does not try to reformat or anything.

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerGestureHelper.kt
Luna712: Done

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerGestureHelper.kt
Luna712: Done

Luna712:
The 3 bugs mentioned *should* be fixed now. Comments restored as well. Still not sure what was meant by formatting though.

Also, another side effect of this now is trailer_custom_layout should be able to be its own thing. It will no longer cause inflation errors if deviating from the other player layouts and the other player layouts no longer need anything only used in trailer either. I could do a follow-up PR to clean this up if wanted unless you want to still keep it exactly in sync even though no longer needed?

fire-light42:
> The 3 bugs mentioned _should_ be fixed now. Comments restored as well. 

What I gave as examples is fixed, but there is comments in almost every function and around them that has been removed. 

It is very inconstant with what comments are removed and what is not. Please go through every comment in the entire pull request you removed and add them back within reason. Even small comments like `// +- 5%` that should be self evident when reading `0.05f` makes it easier to parse, as kotlin does not always have great typing for units. 

> Still not sure what was meant by formatting though.

It looks like you have use a different code formatter, I tried to find some online formatter to compare but it looks every code formatter for kotlin is custom and there is no standard. However in general it looks like you:

```kotlin
open fun f() {
    throw NotImplementedError() // This is the general kotlin style in CloudStream
}
```
vs
```kotlin
open fun f() { throw NotImplementedError() } // This is not
```

And have code like bindViews that adds a lot of space to make it line up:
```kotlin
exoDuration                  = root.findViewById(androidx.media3.ui.R.id.exo_duration)
exoFfwdText                  = root.findViewById(R.id.exo_ffwd_text)
```

Please check your settings in android studio ❤️

Luna712:
@fire-light42 fixed comments (hopefully I didn't miss any...) and formatting as best I could. My PC broke down yesterday so I was more so guessing what it should look like now, I will get a new PC within the next couple weeks, but can't actually see what Android Studio does again right now or check my settings there. If you want, feel free to commit to this, or can wait until I get a new PC again, but I hope it doesnt create massive merge conflicts in the meantime... I do apologize for that bit of inconvenience with this.

Most of the formatting now though aligns with at least some other parts of the app, so if it still has issues, other unrelated parts could probably use some formatting fixes eventually as well.

fire-light42:
Very good changes, unfortunately I found another bug. You updated the function isValidTouch, and now it is not working at all. Please revert that function. Otherwise this is mergable, which is surprising for such a large pull request. The code is really good, compared to what we have now.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -476,12 +459,12 @@ class CS3IPlayer : IPlayer {
         )
     }
 
+    private var currentAudioTrack: AudioTrack? = null
```
fire-light42: Please move this into the function, it makes no sense to be a member variable.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -79,676 +31,142 @@ const val NEXT_WATCH_EPISODE_PERCENTAGE = 90
...
-        throw NotImplementedError()
-    }
+    var subView: SubtitleView?
+        get() = playerHostView?.subView
+        set(value) { playerHostView?.subView = value }
 
-    open fun subtitlesChanged() {
-        throw NotImplementedError()
-    }
+    protected open var hasPipModeSupport: Boolean
```
fire-light42: This is a circular definition, as it is only used in `playerHostView?.hasPipModeSupport = hasPipModeSupport`

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -79,676 +31,142 @@ const val NEXT_WATCH_EPISODE_PERCENTAGE = 90
...
-    }
+    var playerPausePlay: ImageView?
+        get() = playerHostView?.playerPausePlay
+        set(value) { playerHostView?.playerPausePlay = value }
 
-    open fun onTracksInfoChanged() {
-        throw NotImplementedError()
-    }
+    /** The underlying [androidx.media3.ui.PlayerView] widget (named to avoid conflict with our [PlayerView]). */
+    var playerView: androidx.media3.ui.PlayerView?
```
fire-light42: It makes no sense for this to be a setter, all views are set in bindViews and never changed.

Luna712:
> Very good changes, unfortunately I found another bug. You updated the function isValidTouch, and now it is not working at all. Please revert that function. Otherwise this is mergable, which is surprising for such a large pull request. The code is really good, compared to what we have now.

@fire-light42

Can you please provide reproduction steps so I can test this when I fix it? What do you do that isn't working properly?

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
Luna712: Hmm. I probably did that as another instance of blindly trusting Android Studio's linting, my bad thanks for pointing it out.

fire-light42:
> Can you please provide reproduction steps so I can test this when I fix it? What do you do that isn't working properly?

Swiping to pull up the system UI from both the top and right when in horizontal mode seeks the video or changes the brightness/volume. We purposefully invalidate those touches for better UX, but this pull request allows all touches, even those opening the system UI.

fire-light42:
I also want to add that while you fixed the trailer popup issue, it still hides the seekbar when paused in a few sec. Just hiding the view is not enough.

Luna712:
> I also want to add that while you fixed the trailer popup issue, it still hides the seekbar when paused in a few sec. Just hiding the view is not enough.

Fixed

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: Done

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: Done

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
Luna712: Done.

Luna712:
> > Can you please provide reproduction steps so I can test this when I fix it? What do you do that isn't working properly?
> 
> Swiping to pull up the system UI from both the top and right when in horizontal mode seeks the video or changes the brightness/volume. We purposefully invalidate those touches for better UX, but this pull request allows all touches, even those opening the system UI.

This should now be fixed as well.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
fire-light42: This was an example, I mean on all views. `playerPausePlay`, `subView`, `playerPausePlay`, and maybe some others I have missed.

Luna712: Oh right sorry, my bad. I think those were the only four though. Done now. Sorry for not checking that previously.

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
```diff
@@ -48,17 +48,11 @@ abstract class AbstractPlayerFragment<T : ViewBinding>(
...
         }
 
-    var subView: SubtitleView?
-        get() = playerHostView?.subView
-        set(value) { playerHostView?.subView = value }
-
-    var playerPausePlay: ImageView?
-        get() = playerHostView?.playerPausePlay
-        set(value) { playerHostView?.playerPausePlay = value }
+    val subView: SubtitleView? = playerHostView?.subView
```
fire-light42: If you remove the `get() =` then it will always be null. Add `get() =` otherwise this will be bugged.

fire-light42 (changes requested):
Do not forget the getter. 
```kotlin
class X {
    var x : Int? = null
    val x_default : Int? = x
    val x_get : Int? get() = x
}


fun main() {
    val x = X()
    x.x = 123
    println(x.x_default) // <-- null
    println(x.x_get)     // <-- 123
}
```

app/src/main/java/com/lagradost/cloudstream3/ui/player/AbstractPlayerFragment.kt
Luna712: I like learning new things about this language. Thank you! I thought get() was the same thing as just = and only used when using set(). That makes a lot of sense though. Thank you for the great explanation here! This is now fixed.

Although I am now understanding why when making this original code, I had to use get() even without set() to work in some places originally also. That helped clear that up so thanks a lot!

Luna712:
@fire-light42 I was wondering your thoughts about my future plans (as in future PR(s)) for this as well:

* Move all player helper classes into a new player/helper directory for organization
* Factor FullScreenPlayer out fully into helper methods
* Remove AbstractPlayerFragment entirely
* Rename GeneratorPlayer to GeneratorPlayerFragment and make it the main fragment, extending BaseFragment directly and using the helper methods factored out of FullScreenPlayer to reduce inheritance tree
* Seperate all trailer player logic from ResultFragmentPhone directly into ResultTrailerPlayer, and move ResultTrailerPlayer to a new class called PhoneTrailerPlayerFragment (or something similar that is not final) in the player directory
* Add a new callback implementation to ResultFragmentPhone so that the new trailer fragment can manipulate the parent UI visibility etc... without so much inheritance and fully separated fragments just using BaseFragment directly
* Remove everything not actually used in trailer from `trailer_custom_layout` (as it is no longer necessary to exactly match the other layouts), and likewise remove everything (if anything) *only* used in trailer from other player layouts to reduce some complexity if possible

My goal with these current plans is to make things clearer and easier to manage and find where things come from then going down the rabbit hole of so much inheritance, unnecessarily, give the player code where it is actually used (especially for the trailer), and make it more consistent with how all other fragments work. I originally planned a lot of this as part of this PR as well, but decided to reduce it as it was already large enough and these could be better split out into other PRs.

I just wanted some thoughts in this future plan I was currently thinking though before doing it after this is merged, so I don't do it if the architecture design is not good with you.

Also noting if I do this I will try not to do so while large PRs are open touching these areas, as I dont want to create harder to resolve merge conflicts with deleted/renamed classes for others or myself in those PRs doing it also.

fire-light42 (approved):

PR #2690, opened 2026-04-18
Support resuming fragmented MP4s

### LINKED ISSUE #2628: Support resuming playback on fragmented MP4 files

Currently on a fragmented MP4 will automatically merge since ba74a90, however it doesn't merge immediately as soon as it loads, it takes a few seconds then merges. That means even if say you last left the video in the middle of the stream, when you resume it, it starts only at the end of the first part, then merges so can never actually properly resume to your last position. This could *probably* be solved on the CS3-side by waiting until it is fully loaded/merged and then resuming.

### Other details

I am not certain if this is possible to fix on our side or not, I tried looking into this a lot before and couldn't figure this part out, but was just reminded of it because of #2627 for some reason. If we determine we can't solve this on our side, that is understandable as well and this can be closed.

### Acknowledgements

Luna712:
Fixes #2628

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -1414,6 +1414,21 @@ class CS3IPlayer : IPlayer {
             event(PlayerAttachedEvent(exoPlayer))
             exoPlayer?.prepare()
 
+            // For offline fragmented MP4s, FLAG_MERGE_FRAGMENTED_SIDX builds the SIDX seek map
+            // incrementally as data is buffered. The initial seek resolves to the nearest merged
+            // entry (~first fragment, 3 s). On STATE_READY, re-seek to the actual saved position.
+            if (onlineSource == null && playbackPosition > (exoPlayer?.duration ?: 0L)) {
+                exoPlayer?.addListener(object : Player.Listener {
+                    private var seekApplied = false
+                    override fun onPlaybackStateChanged(playbackState: Int) {
+                        if (seekApplied || playbackState != Player.STATE_READY) return
+                        seekApplied = true
+                        exoPlayer?.seekTo(currentWindow, playbackPosition)
+                        exoPlayer?.removeListener(this)
+                    }
+                })
+            }
+
             exoPlayer?.let { exo ->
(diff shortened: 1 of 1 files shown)


fire-light42 (approved):
Code looks good, and will be merged after testing ❤️

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1414,6 +1414,21 @@ class CS3IPlayer : IPlayer {
             event(PlayerAttachedEvent(exoPlayer))
             exoPlayer?.prepare()
 
+            // For offline fragmented MP4s, FLAG_MERGE_FRAGMENTED_SIDX builds the SIDX seek map
+            // incrementally as data is buffered. The initial seek resolves to the nearest merged
+            // entry (~first fragment, 3 s). On STATE_READY, re-seek to the actual saved position.
+            if (onlineSource == null && playbackPosition > (exoPlayer?.duration ?: 0L)) {
```
fire-light42, line 1422: I am having a hard time replicating the core issue. I have tried various media [demos](https://github.com/androidx/media/blob/release/demos/main/src/main/assets/media.exolist.json) but it seems to work without this patch?

Do you have any good example video we can include as a comment, otherwise we may remove this in the future thinking it's solved.

Luna712: Only ones I have locally. To large to upload (~5G) with the first fragment being 3s, so it resumes to 3s, then looses the actual resume time stamp.

Luna712: I am only aware of one demo file this *may* be reproducible on: https://github.com/androidx/media/blob/release/libraries/test_data/src/test/assets/media/mp4/sample_fragmented_seekable_multiple_sidx.mp4 however, that it to short to resume on at all, and also it is so small that merging the fragments is super fast it may in fact not be reproducible on it.

Luna712: I updated the comment to be a little bit more clear.

fire-light42 (approved):

PR #2691, opened 2026-04-18
Fix STATE_IDLE issues in player

### LINKED ISSUE #2175: Closing media notification breaks ability to resume player
### Steps to reproduce

1. Play a video
2. Pause the video
3. Swipe the media notification in device status bar to clear it
4. Try to resume in player

### Expected behavior

Video should resume

### Actual behavior

The play button doesn't do anything anymore.

4.6.1 52100dc

### Android version

Android 16

### Logcat

```java

```

### Other details

_No response_

### Acknowledgements

Luna712:
Fixes #2175  and also other issues

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -961,6 +961,21 @@ class CS3IPlayer : IPlayer {
                 when (event) {
                     CSPlayerEvent.Play -> {
                         event(PlayEvent(source))
+                        // If the player was stopped (e.g. notification dismissed) it lands in
+                        // STATE_IDLE. A bare play() call is a no-op in that state, re-prepare and
+                        // then resume to the saved position once we are in STATE_READY again.
+                        if (playbackState == Player.STATE_IDLE) {
+                            prepare()
+                            exoPlayer?.addListener(object : Player.Listener {
+                                private var seekApplied = false
+                                override fun onPlaybackStateChanged(playbackState: Int) {
+                                    if (seekApplied || playbackState != Player.STATE_READY) return
+                                    seekApplied = true
+                                    exoPlayer?.seekTo(currentWindow, playbackPosition)
+                                    exoPlayer?.removeListener(this)
+                                }
+                            })
+                        }
                         play()
                     }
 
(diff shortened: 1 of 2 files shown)


Luna712:
Would like to also note that while it works it is kinda also a PoC, if anyone elese has a better idea here would love to hear it, I spent a long time debugging a few of these issues and this is just what solution I landed up on to fix it. But may not be the best way.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -961,6 +961,21 @@ class CS3IPlayer : IPlayer {
...
+                        // STATE_IDLE. A bare play() call is a no-op in that state, re-prepare and
+                        // then resume to the saved position once we are in STATE_READY again.
+                        if (playbackState == Player.STATE_IDLE) {
+                            prepare()
+                            exoPlayer?.addListener(object : Player.Listener {
+                                private var seekApplied = false
+                                override fun onPlaybackStateChanged(playbackState: Int) {
+                                    if (seekApplied || playbackState != Player.STATE_READY) return
+                                    seekApplied = true
+                                    exoPlayer?.seekTo(currentWindow, playbackPosition)
```
Luna712: This is broken logic right now I just realized also. It will re seek to whatever it was when it started since playbackPosition doesn't change. Will look into that later.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -393,6 +393,16 @@ class GeneratorPlayer : FullScreenPlayer() {
                     }
                 }
             })
+            .setNotificationListener(object : PlayerNotificationManager.NotificationListener {
+                override fun onNotificationCancelled(notificationId: Int, dismissedByUser: Boolean) {
+                    if (dismissedByUser) {
```
Luna712: Hmm... maybe this is actually worse then it being inconsistent and coming back. This could also mean if the notification is swiped while paused then when resumed it doesnt come back. I have noticed it randomly come back 10 minutes later without ever resuming it which is what this was for but maybe that is a better outcome. Not sure.

At the minimum we should fix STATE_IDLE issues anyway which causes other breakage of the resume button beyond just #2175 as well.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
Luna712: Fixed.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
Luna712: Removed this part now.

fire-light42 (approved):

PR #2701, opened 2026-04-19
Add new URL for Voe

Luna712:


Luna712:
Whoops. I forgot this needed adding to ExtractorApi also lol. Thanks!

PR #2710, opened 2026-04-20
Feat: Anime Skip

fire-light42:
This adds "https://anime-skip.com" as a skip service. However the API is straight up a pain to work with, and really stupid. It requires an account with a API Client to make sure we do not overload the service. However because every request requires a client id, even login requests, it is a catch 22. We have to bootstrap with a rate-limited API, and then use the user client-id. 

The data is also garbage, given that we have to match with name, and not malId or similar. Moreover, the timestamps are only related to each other, not a duration.


Fuzzythegoatq:
Vaild

PR #2712, opened 2026-04-20
Add new TVType for generic Video

Luna712:
Similar to generic Audio. Using the Movie type for some things such as DailyMotion or YouTube videos does not make much sense, in that case a generic Video type makes more sense.


fire-light42 (approved):
Looks good, but maybe add `@Prerelease`?

Luna712:
> Looks good, but maybe add `@Prerelease`?

I wasn't sure if `@Prerelease` would work inside the enum or not?

Luna712:
> Looks good, but maybe add `@Prerelease`?

@fire-light42

I did this now btw, since it seemed to work.

fire-light42 (approved):

PR #2714, opened 2026-04-22
Added Library Poster Size Slider

AryanMir15:
Implemented library poster scaling via RecyclerView span count adjustment. Leveraged AI assistance for code review and optimization of the layout logic to ensure the slider transitions are smooth and performant.

Also adds gitignore rules for app/prerelease/ and app/stable/ folders to prevent committing release APK files.
Screenshot_2026-04-22-17-15-16-71_bba4627ecec1346e77f282610a5b88e8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -225,8 +225,4 @@ fabric.properties
 
 !/gradle/wrapper/gradle-wrapper.jar
 
-# Ignore release APK folders
-app/prerelease/
-app/stable/
-
 # End of https://www.toptal.com/developers/gitignore/api/kotlin,java,android,androidstudio,visualstudiocode
--- a/.gitignore
+++ b/.gitignore
@@ -225,4 +225,8 @@ fabric.properties
 
 !/gradle/wrapper/gradle-wrapper.jar
 
+# Ignore release APK folders
+app/prerelease/
+app/stable/
+
 # End of https://www.toptal.com/developers/gitignore/api/kotlin,java,android,androidstudio,visualstudiocode
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
@@ -1,9 +1,12 @@
 package com.lagradost.cloudstream3.ui.library
 
+import android.content.Context
+import android.preference.PreferenceManager
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import android.widget.FrameLayout
 import androidx.core.view.isVisible
+import com.lagradost.cloudstream3.R
 import com.lagradost.cloudstream3.databinding.SearchResultGridExpandedBinding
 import com.lagradost.cloudstream3.syncproviders.SyncAPI
 import com.lagradost.cloudstream3.ui.AutofitRecyclerView
@@ -12,6 +15,7 @@ import com.lagradost.cloudstream3.ui.NoStateAdapter
 import com.lagradost.cloudstream3.ui.ViewHolderState
 import com.lagradost.cloudstream3.ui.search.SearchClickCallback
 import com.lagradost.cloudstream3.ui.search.SearchResultBuilder
+import com.lagradost.cloudstream3.utils.UIHelper.getSpanCount
 import kotlin.math.roundToInt
 
 class PageAdapter(
@@ -26,6 +30,19 @@ class PageAdapter(
         }
     })) {
     private val coverHeight: Int get() = (resView.itemWidth / 0.68).roundToInt()
+
+    companion object {
+        fun updatePosterSize(resView: AutofitRecyclerView, context: Context, value: Int? = null) {
+            val scale = value ?: PreferenceManager.getDefaultSharedPreferences(context)
+                ?.getInt(context.getString(R.string.library_poster_size_key), 0) ?: 0
+            // Adjust spanCount to control poster size
+            // Higher scale = fewer columns = larger posters
+            val baseSpanCount = context.getSpanCount()
+            val adjustedSpanCount = maxOf(1, (baseSpanCount / (1.0f + scale * 0.1f)).roundToInt())
+            // Update the RecyclerView's spanCount
+            resView.spanCount = adjustedSpanCount
+        }
+    }
 
     override fun onCreateContent(parent: ViewGroup): ViewHolderState<Any> {
         return ViewHolderState(
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
@@ -32,10 +32,10 @@ class PageAdapter(
     private val coverHeight: Int get() = (resView.itemWidth / 0.68).roundToInt()
 
     companion object {
-        fun updatePosterSize(resView: AutofitRecyclerView, context: Context, value: Int? = null) {
+        fun updatePosterSize(resView: AutofitRecyclerView?, context: Context, value: Int? = null) {
             val spanCount = value ?: PreferenceManager.getDefaultSharedPreferences(context)
                 .getInt(context.getString(R.string.library_poster_size_key), 3)
-            resView.spanCount = spanCount
+            resView?.spanCount = spanCount
         }
     }
 
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
@@ -1,8 +1,8 @@
 package com.lagradost.cloudstream3.ui.library
 
 import android.content.Context
-import android.preference.PreferenceManager
 import android.view.LayoutInflater
+import androidx.preference.PreferenceManager
 import android.view.ViewGroup
 import android.widget.FrameLayout
 import androidx.core.view.isVisible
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
```diff
@@ -1,9 +1,12 @@
 package com.lagradost.cloudstream3.ui.library
 
+import android.content.Context
+import android.preference.PreferenceManager
```
fire-light42: You are using the deprecated PreferenceManager. Use `androidx.preference.PreferenceManager`

app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
```diff
@@ -27,6 +31,19 @@ class PageAdapter(
...
+    companion object {
+        fun updatePosterSize(resView: AutofitRecyclerView, context: Context, value: Int? = null) {
+            val scale = value ?: PreferenceManager.getDefaultSharedPreferences(context)
+                ?.getInt(context.getString(R.string.library_poster_size_key), 0) ?: 0
+            // Adjust spanCount to control poster size
+            // Higher scale = fewer columns = larger posters
+            val baseSpanCount = context.getSpanCount()
+            val adjustedSpanCount = maxOf(1, (baseSpanCount / (1.0f + scale * 0.1f)).roundToInt())
+            // Update the RecyclerView's spanCount
+            resView.spanCount = adjustedSpanCount
```
fire-light42: Just let users adjust the span count instead. It is confusing for users when changing the size from 3 to 5 does nothing.

fire-light42: This breaks when users rotate their phones back and forth.

app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
```diff
@@ -27,6 +31,19 @@ class PageAdapter(
     })) {
     private val coverHeight: Int get() = (resView.itemWidth / 0.68).roundToInt()
 
+    companion object {
+        fun updatePosterSize(resView: AutofitRecyclerView, context: Context, value: Int? = null) {
+            val scale = value ?: PreferenceManager.getDefaultSharedPreferences(context)
+                ?.getInt(context.getString(R.string.library_poster_size_key), 0) ?: 0
```
fire-light42: You should always have the same default value here as in `android:defaultValue`. Otherwise the setting does not reflect the visuals (default value is 5 visually but 0 in practice).

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -65,6 +65,20 @@ class SettingsUI : BasePreferenceFragmentCompat() {
             true
         }
 
+        getPref(R.string.library_poster_size_key)?.setOnPreferenceChangeListener { _, newValue ->
+            context?.let { 
+                // Reload library to apply new poster size
+                (it.getActivity() as? MainActivity)?.let { activity ->
+                    activity.supportFragmentManager.findFragmentById(R.id.navigation_library)?.let { fragment ->
+                        if (fragment is com.lagradost.cloudstream3.ui.library.LibraryFragment) {
+                            activity.recreate()
```
fire-light42: This is never triggered, it works regardless. You can remove this block of code.

app/src/main/res/xml/settings_ui.xml
```diff
@@ -35,7 +35,7 @@
             app:showSeekBarValue="true" />
 
         <SeekBarPreference
-            android:defaultValue="0"
+            android:defaultValue="5"
```
fire-light42, line 38: Why change the default value here?

.gitignore
```diff
@@ -225,4 +225,8 @@ fabric.properties
 
 !/gradle/wrapper/gradle-wrapper.jar
 
+# Ignore release APK folders
+app/prerelease/
+app/stable/
```
fire-light42: We do not use these directories for APKs. You should remove these to prevent confusion if we do use these directories in the future.

fire-light42 (changes requested):
Good idea, you just need to fix many small details.

Luna712 (changes requested):
> e: file:///home/runner/work/cloudstream/cloudstream/app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt:36:54 Unresolved reference 'getInt'.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -65,6 +65,10 @@ class SettingsUI : BasePreferenceFragmentCompat() {
             true
         }
 
+        getPref(R.string.library_poster_size_key)?.setOnPreferenceChangeListener { _, newValue ->
```
Luna712: This entire block is unnecessary

app/src/main/res/values/strings.xml
```diff
@@ -745,6 +745,8 @@
     <string name="overscan_settings">Overscan</string>
     <string name="poster_size_settings_des">Changes size of posters</string>
     <string name="poster_size_settings">Poster size</string>
+    <string name="library_poster_size_settings_des">Changes size of library posters</string>
+    <string name="library_poster_size_settings">Library poster size</string>
```
fire-light42: These are now inaccurate, please update the accordingly.

fire-light42 (changes requested):
Remember to also fix the previous unaddressed comments.

app/src/main/res/xml/settings_ui.xml
AryanMir15: The duplicate defaultValue was already removed in a previous commit the current file only has one defaultValue per SeekBarPreference. Though im also confused as to why its not flagging this as outdated. Or maybe the change is too small to be flagged as outdated

app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
```diff
@@ -27,6 +31,14 @@ class PageAdapter(
     })) {
     private val coverHeight: Int get() = (resView.itemWidth / 0.68).roundToInt()
 
+    companion object {
+        fun updatePosterSize(resView: AutofitRecyclerView?, context: Context, value: Int? = null) {
```
fire-light42, line 35: This makes no sense as a companion object function since you are affecting the PageAdapter's AutofitRecyclerView.

app/src/main/java/com/lagradost/cloudstream3/ui/settings/SettingsUI.kt
```diff
@@ -65,6 +66,11 @@ class SettingsUI : BasePreferenceFragmentCompat() {
...
+            context?.let { PageAdapter.updatePosterSize(null, it, newValue as? Int) }
```
fire-light42, line 70: This makes no sense as a piece of code, it will only waste CPU cycles.

This will do `null?.spanCount = newValue` which is a no-op.

app/src/main/res/values/strings.xml
```diff
@@ -743,8 +743,10 @@
     </string>
     <string name="overscan_settings_des">Changes the bounds of the screen</string>
     <string name="overscan_settings">Overscan</string>
-    <string name="poster_size_settings_des">Changes size of posters</string>
-    <string name="poster_size_settings">Poster size</string>
+    <string name="poster_size_settings_des">Changes the number of poster columns</string>
+    <string name="poster_size_settings">Poster columns</string>
```
fire-light42, line 747: This is just false, poster size does not change the poster columns.

app/src/main/res/xml/settings_ui.xml
fire-light42: It is not removed? You still have `android:defaultValue="5"`, please change it back to 0. 😕

fire-light42 (changes requested):
Please read what you contribute before requesting a review. 

Your latest changes frankly do not make sense and would have been caught if you reviewed your own code before requesting a review. [Read rule 2 of our AI policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md).

I expect a baseline level of effort from contributors, since reviewing code takes a lot of time and effort.

If this continues I will be forced to close the pull request. It is okay if you make mistakes, new programmers are welcome, but you need to try your best.

app/src/main/java/com/lagradost/cloudstream3/ui/library/PageAdapter.kt
AryanMir15: I am SO sorry for my behavior and thank you so much for being patient with me, the next time i request a review, i wont disappoint, and im very new to open source and i dont wanna force my code into cloudstream, i will fix all of these in the next review.

app/src/main/res/values/strings.xml
AryanMir15: I am sorry but that actually IS what i had the setting do, i swapped from size to column count which automatically resizes the posters, should i revert back to slider up > poster size increase?

fire-light42: You are confusing `poster_size_settings` and `library_poster_size_settings`.

You did swap the **library** poster setting, but you did not swap the **normal** poster setting. The normal poster setting does not change columns, it changes the size. You should therefore revert the changes to `poster_size_settings` and `poster_size_settings_des`.

PR #2720, opened 2026-04-23
Revert "Improve subtitle selection UX: Move "No Subtitles" option to bottom"

Luna712:
Reverts recloudstream/cloudstream#2523

Reason for revert: while it being on the bottom may be more consistent, it is very hard to get used to and you have to scroll a lot more, especially on TV, to turn them off, which is poor UX IMO.

This can be closed if you decide you do not want to revert though.


fire-light42:
I suppose that **E**nglish is always a lot closer to the top than bottom, which would minimize the scrolling for an average user (assuming a primarily English speaking user base). It is truly difficult to decide which option is best. 

I think I like top subtitles more, but I need to do more testing on different devices before deciding to revert the change.

Luna712:
Totally understandable, and yeah I dont really have an issue with it on Phone as scrolling is easier but it is inconvenient on TV.

PR #2731, opened 2026-04-28
Fix(TV): Color on "Skip Chapter" button

fire-light42:
Changes the style of the TV skip chapter button, to make it easier to see when it is selected. Also moves it down, by removing the previewFrameLayout on TV.


mooncrown04:
// --- OPEN EPISODE LIST WITH OPTIONS / MENU KEY ---
KeyEvent.KEYCODE_MENU, 
KeyEvent.KEYCODE_SETTINGS -> {
    if (isLocked != true) { 
        // This is the actual function that opens the series list:
        toggleEpisodesOverlay(true)
        return true
    }
}                          
"I have updated the comments in the code to English for better clarity. Please note that the path for this file is cloudstream/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt. Could you please add this code snippet to the project?"

PR #2733, opened 2026-04-29
Always go to next source in player and add thread safety

Luna712:
We don't want to ever kick out of the player if the first source fails, we can just show the unknown error toast and still go to the next source for more stable error handling. In the future we can still expand the different types of toasts shown to be clear about what the error is, but in reality we should never just kick out of the player if the first source is an error and there are other sources.

This also adds thread safety, and fixes issues with context required on the UI thread. Before lint couldn't follow `ioSafe` or other custom Coroutines methods, but this adds annotations to tell lint to follow properly ensuring proper build failures with `WrongThread` should it be called on the wrong thread.

For example with context:
```
Error: Method getContext must be called from the UI thread, currently inferred thread is worker thread
```
Is now given if called on the wrong thread. Doing it like this also would have caught other issues in the past as well.

Also, `androidx.annotation:annotation` is fully KMP compatible so has no issues with being in `commonMain` in library either. Just for consistency to ensure we always use the same version, it is also added explicitly to the `app` module.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -1770,7 +1770,6 @@ class CS3IPlayer : IPlayer {
         return exoPlayer != null
     }
 
-
     @MainThread
     private fun loadTorrent(context: Context, link: ExtractorLink) {
         ioSafe {
@@ -1791,7 +1790,7 @@ class CS3IPlayer : IPlayer {
                     loadOnlinePlayer(context, newLink)
                 }
             } catch (t: Throwable) {
-                event(ErrorEvent(t))
+                runOnMainThread { event(ErrorEvent(t)) }
             }
         }
     }
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -12,6 +12,7 @@ import android.os.Looper
 import android.util.Log
 import android.util.Rational
 import android.widget.FrameLayout
+import androidx.annotation.AnyThread
 import androidx.annotation.MainThread
 import androidx.annotation.OptIn
 import androidx.appcompat.app.AlertDialog
@@ -206,16 +207,14 @@ class CS3IPlayer : IPlayer {
     private var requestedListeningPercentages: List<Int>? = null
 
     private var eventHandler: ((PlayerEvent) -> Unit)? = null
-    private val mainHandler = Handler(Looper.getMainLooper())
 
+    @AnyThread
     fun event(event: PlayerEvent) {
-        // Ensure that all work is done on the main looper, aka main thread
-        if (Looper.myLooper() == mainHandler.looper) {
+        // Ensure that all work is done on the main thread.
+        if (Looper.getMainLooper().isCurrentThread) {
             eventHandler?.invoke(event)
-        } else {
-            mainHandler.post {
-                eventHandler?.invoke(event)
-            }
+        } else runOnMainThread {
+            eventHandler?.invoke(event)
         }
     }
 
@@ -235,8 +234,9 @@ class CS3IPlayer : IPlayer {
         }
     }
 
+    @AnyThread
     override fun initCallbacks(
-        eventHandler: ((PlayerEvent) -> Unit),
+        @MainThread eventHandler: ((PlayerEvent) -> Unit),
         requestedListeningPercentages: List<Int>?,
     ) {
         this.requestedListeningPercentages = requestedListeningPercentages
@@ -1790,7 +1790,7 @@ class CS3IPlayer : IPlayer {
                     loadOnlinePlayer(context, newLink)
                 }
             } catch (t: Throwable) {
-                runOnMainThread { event(ErrorEvent(t)) }
+                event(ErrorEvent(t))
             }
         }
     }
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerView.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerView.kt
@@ -615,8 +615,8 @@ class PlayerView @JvmOverloads constructor(
     /** Error handling */
 
     fun playerError(exception: Throwable) {
-        fun showErrorToast(message: String, gotoNext: Boolean = false) {
-            if (gotoNext && callbacks?.hasNextMirror() == true) {
+        fun showErrorToast(message: String) {
+            if (callbacks?.hasNextMirror() == true) {
                 showToast(message, Toast.LENGTH_SHORT)
                 callbacks?.nextMirror()
             } else {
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerView.kt
```diff
@@ -680,17 +680,17 @@ class PlayerView @JvmOverloads constructor(
                  * prevent switching to the next source.
                  */
                 (context as? Activity)?.runOnUiThread {
```
Luna712: Hmm. This could also be another issue. Maybe we want to move this to `showErrorToast` as well? Reason being so that no matter what error is given that runs on the UI thread, if it doesn't I found 4 things happening with torrents before I did it like this the first time.
1. Sometimes the app just totally crashed
2. Sometimes it froze and become unresponsive
3. Sometimes all buttons in player became unresponsive and no longer did anything
4. Regardless of `gotoNext`, it just kicked out of the player

It was very inconsistent though, other than the fact one of those issues *always* happened just which one(s) was inconsistent, but it *never* worked properly, but if anything other than torrents can do this on a non-UI thread, that could cause issues as well, but I am not sure if any others do.

fire-light42: The playerError function is expected to run on the main thread, as that is the thread exoplayer throws exceptions at. However, this system is hijacked by SocketTimeoutException on torrents using ErrorEvent, as it uses a local server on another thread to host a http stream. We simply reused the function for ease of use. Therefore this system is bug free, even if it feels weird. However, given that you refactor the non specificed exceptions to `"gotoNext = true"`, this `else ->` clause also has to have a `runOnUiThread`. 

What exoplayer does when not on the main thread is weird, as you have seen, and is why all interactions with exoplayer has to be done on the main thread. You can wrap `runOnUiThread` on the entire playerError to make this behavior explicit, or mark it as `@MainThread` and let the caller ensure it. I am more of a fan of the `@MainThread` approach as it does not hide the race condition possible when launching threads/runnables/jobs. 

Moreover, InvalidFileException is also unused, so it can be removed.

fire-light42 (commented):
The idea of this is to handle irrevocable and unknown errors by simply exiting the player, and can be seen as a soft app crash. If we do not know what happen, the system may have been left in a bad state causing worse problems than if we did not exit. 

This is the general java philosophy, but I can see why it may cause user friction. Especially given that no such problem is realistically expected to happen.

app/src/main/java/com/lagradost/cloudstream3/ui/player/PlayerView.kt
Luna712: I have added thread safety now also. Used `@UiThread` for playerError and added resolution for the custom Coroutine helpers, which should ensure things use the right thread for future as well, otherwise the annotations don't actually do much. It did catch the loadTorrent running `ErrorEvent` on the wrong thread before I wrapped that in `runOnMainThread` as well.

I also removed InvalidFileException.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1791,7 +1790,7 @@ class CS3IPlayer : IPlayer {
                     loadOnlinePlayer(context, newLink)
                 }
             } catch (t: Throwable) {
-                event(ErrorEvent(t))
+                runOnMainThread { event(ErrorEvent(t)) }
```
fire-light42: No, I mean that runOnMainThread for the caller of playerError, not each and every event. "Events" are supposed to be asynchronous message passing like, where playerError is not.

library/src/jvmMain/kotlin/com/lagradost/cloudstream3/utils/Coroutines.jvm.kt
```diff
@@ -1,5 +1,9 @@
 package com.lagradost.cloudstream3.utils
 
-actual fun runOnMainThreadNative(work: () -> Unit) {
+import androidx.annotation.AnyThread
```
fire-light42, line 3: This feels so wrong...

fire-light42 (changes requested):

library/src/jvmMain/kotlin/com/lagradost/cloudstream3/utils/Coroutines.jvm.kt
Luna712: Indeed, but androidx.annotation is fully cross platform support

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
Luna712: But ErrorEvent is only used by that, all ErrorEvents should always be passed on the main thread since media3 expects its operations on the main thread, unless I am missing something? Another thing I wanted to do is avoid runOnMainThread unnecessarily if it is already on the main thread, and this just enforces that in the IDE better, we don't do each and every `event` just ErrorEvent, as others can be more asynchronous then that one which needs UI operations more.

The IDE just can't follow "is ErrorEvent" but it can follow if ErrorEvent itself is enforced like this.

Another reason I did it like this is so callers are always enforced to pass it like that if we ever extracted the player logic into a stable API. Not against changing it though, this just felt more explicit to me but if this doesn't work right for it it can absolutely be changed.

Either way, I may just be talking nonsense with this. I only just started very extensively looking into how the thread annotations really resolve and why they weren't working for our own methods to know what threads are running, and this was just what I came up with. absolutely no issues changing though.

library/src/jvmMain/kotlin/com/lagradost/cloudstream3/utils/Coroutines.jvm.kt
Luna712: Note it will resolve on JVM, but it doesnt actually do anything though. But the signatures including all annotations need to still match exactly the same in the `actual` as the `expect` function.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
fire-light42: Hmm, I looked around a bit, and it already looks like event is doing the equivalent of runOnMainThread. 

```kotlin
 fun event(event: PlayerEvent) {
        // Ensure that all work is done on the main looper, aka main thread
        if (Looper.myLooper() == mainHandler.looper) {
            eventHandler?.invoke(event)
        } else {
            mainHandler.post {
                eventHandler?.invoke(event)
            }
        }
    }
```

Luna712: Oh. I will change this and add `@MainThread` annotation to that instead for IDE resolution and remove runOnMainThread from that, if it works for you?
This changed in 1fd561f which is after I had to make the work around for SocketTimeoutException in ed6646c so thats no longer needed anyway.

fire-light42: `event` can have `@AnyThread` while `mainCallback` can have a `@MainThread`.

Luna712: Done. Thanks!

fire-light42 (approved):

PR #2738, opened 2026-04-30
Fix `DownloadedPlayerActivity` not loading new files when activity is already running

Luna712:
When the activity was already open (e.g. after pressing the device home button) and a new file was opened, the new intent was silently dropped and the previous file continued playing. Pressing Back and reopening worked fine because that path hit `onCreate()` fresh.

The root cause is that Android reuses an existing activity instance and delivers the new intent via `onNewIntent()` instead of `onCreate()`. Since `onNewIntent()` was not overridden, the new intent was never handled.

Additionally, with the new `onNewIntent()` handling, opening the same file that is already playing would now cause it to unnecessarily reload as well. Therefore, we add a new `isSameIntent` check to prevent the same currently playing file from reloading unnecessarily.


phisher98:
I have observed similar thing when building extensions it loads in app it doesn't reload and gets changes

fire-light42 (approved):

PR #2742, opened 2026-05-01
Fix no links found bug in DownloadedPlayerActivity

Luna712:
Fixes a bug I accidentally caused in #2738 where you open a local video, press the back button, then reopen it from recents, and will no just immediately exir the player with "No Links Found" and will repeat a few times (4-5) before finally opening it again.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadedPlayerActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadedPlayerActivity.kt
@@ -11,6 +11,7 @@ import com.lagradost.cloudstream3.mvvm.safe
 import com.lagradost.cloudstream3.ui.player.OfflinePlaybackHelper.playLink
 import com.lagradost.cloudstream3.ui.player.OfflinePlaybackHelper.playUri
 import com.lagradost.cloudstream3.utils.BackPressedCallbackHelper.attachBackPressedCallback
+import com.lagradost.cloudstream3.utils.BackPressedCallbackHelper.detachBackPressedCallback
 import com.lagradost.cloudstream3.utils.UIHelper.enableEdgeToEdgeCompat
 
 class DownloadedPlayerActivity : AppCompatActivity() {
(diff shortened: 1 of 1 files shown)


Luna712:
Hmm. This only seems to make it happen less actually it still happens sometimes. Have no idea why. Hoping someone else can take a look if possible and see what I am missing? This bug is less of an annoyance then the previous bug that #2738 fixed was, but it still can get a little annoying. I just can't seem to figure out why it happens. `onCreate` is being called when it does (not `onNewIntent`) and there should be no functional change in what `onCreate` is doing before #2738 with this one.

fire-light42:
I believe the function you are looking for is `finishAndRemoveTask()` instead of `finish()`.

Luna712:
> I believe the function you are looking for is `finishAndRemoveTask()` instead of `finish()`.

@fire-light42 

Doesn't work either. Also doesnt prevent the issue without finish() at all.

fire-light42:
> > I believe the function you are looking for is `finishAndRemoveTask()` instead of `finish()`.
> 
> @fire-light42
> 
> Doesn't work either. Also doesnt prevent the issue without finish() at all.

Are you sure this is not some android issue? Because I am unable to replicate any bug if I use finishAndRemoveTask with this pull request instead of finish. If possible could you show a video of the bug? `finishAndRemoveTask` will make it impossible to open if from recents after you have called finish, as opposed to recreating the activity with an outdated intent.

Luna712:
> > > I believe the function you are looking for is `finishAndRemoveTask()` instead of `finish()`.
> > 
> > 
> > @fire-light42
> > Doesn't work either. Also doesnt prevent the issue without finish() at all.
> 
> Are you sure this is not some android issue? Because I am unable to replicate any bug if I use finishAndRemoveTask with this pull request instead of finish. If possible could you show a video of the bug? `finishAndRemoveTask` will make it impossible to open if from recents after you have called finish, as opposed to recreating the activity with an outdated intent.

@fire-light42 

Yes even happens if you open the first time again from file manager also.

Luna712:
Here is a recording of it as well:

fire-light42:
> Here is a recording of it as well: https://github.com/user-attachments/assets/ab50af45-def0-4fbb-984d-12e78a70ccb9

Is attachBackPressedCallback never invoked with `finishAndRemoveTask`? Because that is the only thing that should create this problem. When I press the back button to close the app, the lambda in `attachBackPressedCallback` is invoked and finishes the app. 

There should be no difference between the first and second launch. I am unable to replicate it with the same setup, with the exception of using a different file manager. Have you tried a different file explorer like https://github.com/zhanghai/MaterialFiles or "Files by Google"?

Luna712:
It happens with MaterialFiles and also Files by Google as well.

Luna712:

Luna712:
Moved to issue #2746

PR #2745, opened 2026-05-01
Refactor: Move all key logic into the player, and added toggleEpisodesOverlay keybind

fire-light42:
I assume the old system was expected to supply more generic events in the case of future peripherals, however this was never utilized. This moves it into the player itself, using dispatchKeyEvent instead of onKeyDown. This is slightly different, but should not give any issues, and at most it might override the system keybindings. 

Moreover, a user also requested a keybinding for toggleEpisodesOverlay, using the menu keybind that was previously used for ToggleHide.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
@@ -117,7 +117,6 @@ object CommonActivity {
     val onColorSelectedEvent = Event<Pair<Int, Int>>()
     val onDialogDismissedEvent = Event<Int>()
 
-    var playerEventListener: ((PlayerEventType) -> Unit)? = null
     var keyEventListener: ((Pair<KeyEvent?, Boolean>) -> Boolean)? = null
     var appliedTheme: Int = 0
     var appliedColor: Int = 0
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt
```diff
@@ -534,87 +532,7 @@ object CommonActivity {
 
 
     fun onKeyDown(act: Activity?, keyCode: Int, event: KeyEvent?): Boolean? {
```
Luna712, line 534: Simply out of curiosity, any reason we don't remove this entire method and callers now?

fire-light42: I do not want to remove good abstractions that can help in the future. For example, someone might want to override onKeyDown, however the current implementation forces them to change the commonactivity onKeyDown, preventing the bug of only having it on MainActivity, when DownloadPlayerActivity should also override any keypress.

Luna712: I appreciate the explanation, but just one question, and I am not questioning your reason, just for my own clarity. Doesn't DownloadedPlayerActivity use GeneratorPlayer and by extension FullScreenPlayer, and ResultTrailerPlayer uses MainActivity, so couldn't we remove the abstraction layer from CommonActivity, and add something to PlayerView that can be called from ResultTrailerPlayer and GeneratorPlayer/FullScreenPlayer directly, and the activities both have a callback that gets it from the fragments and overrides their activity level `onKeyDown` methods, in order to keep all future player logic in the actual player? That is my long term goal also in order to allow the player to be used externally more as well, if we ever were to split it to another module making it more stand alone or something.

Honestly, I may not even be making sense here though. I don't think I am explaining this well either.

fire-light42: No, GeneratorPlayer is a fragment that inherits from FullScreenPlayer. It does not use ResultTrailerPlayer or MainActivity at all. CommonActivity exists to separate app wide logic (like themes) from specific logic like AccountSelectActivity for accounts or DownloadedPlayerActivity for downloaded files. 

What I have done is exactly what you describe, I moved the onKeyDown to the local dispatchKeyEvent (slightly different semantics but still the same) inside the player. This refactor is precisely to move the more general functionality to where it belongs.

Luna712: That makes sense. I just reread what this was actually doing and I got a little confused before. Apologies for that. This is is a nice thing to do and also aligns with my long term goals for the player as well.

Luna712 (approved):
Works well for me as well. This aligns exactly with my own vision/idea for the player to keep all player code actually with the player itself.

PR #2750, opened 2026-05-02
Fix race condition in player

### LINKED ISSUE #2746: No Links Found bug in DownloadedPlayerActivity

Luna712:
Fixes #2746

I was unable to reproduce the issue after this, but also it doesn't happen every single time anyway so I can not be 100% certain this fully fixes it, though seemed to.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
@@ -2125,6 +2125,24 @@ class GeneratorPlayer : FullScreenPlayer() {
         }
     }
 
+    private var langFilterList = listOf<String>()
+    private var filterSubByLang = false
+
+    private fun updateCurrentSubs(set: Set<SubtitleData>) {
+        val setOfSub = mutableSetOf<SubtitleData>()
+        if (langFilterList.isNotEmpty() && filterSubByLang) {
+            Log.i("subfilter", "Filtering subtitle")
+            langFilterList.forEach { lang ->
+                Log.i("subfilter", "Lang: $lang")
+                setOfSub += set.filter {
+                    it.originalName.contains(lang, ignoreCase = true) ||
+                        it.origin != SubtitleOrigin.URL
+                }
+            }
+            currentSubs = setOfSub
+        } else currentSubs = set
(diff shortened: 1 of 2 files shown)


fire-light42 (changes requested):
1. Please change currentLinks+currentSubs to ConsistentLiveData if you want to do this.
2. Please see filterSubByLang.

Luna712:
> 1. Please change currentLinks+currentSubs to ConsistentLiveData if you want to do this.
> 2. Please see filterSubByLang.

@fire-light42

I *think* I did what you mean?

fire-light42 (approved):

Luna712:
@fire-light42 I wonder if you could also perhaps explain to me a little bit what ConsistentLiveData does and when it should be used? I'm having a hard time understanding it's code, but am curious at least when it is something that should be used.

fire-light42:
> @fire-light42 I wonder if you could also perhaps explain to me a little bit what ConsistentLiveData does and when it should be used? I'm having a hard time understanding it's code, but am curious at least when it is something that should be used.

ConsistentLiveData is used to avoid the threading issue with LiveData. To be more precise, LiveData is "thread-safe", but only makes its value "visible" after it writes the value in the main thread. This means that:
```
live.postValue(x)
println(live.value) // Not necessarily "x" when on a non-main thread
```
ConsistentLiveData is simply a LiveData with a backing volatile variable, so it becomes visible immediately when using `.value`. See more in the doc comment on ConsistentLiveData. 

In general you should never have to use .value given that observe should only be used for UI. However, to solve this problem we use ConsistentLiveData to solve thread issues whenever we need to access .value.

Luna712:
That makes a ton of sense and I really appreciate you taking the time for the explanation, thanks a ton for that and it will be helpful for me in the future I think.

Luna712:
I think I may have a better way to do thia soon that fixes the underlying issue in the view model rather than relying on this before startPlayer. I am testing and will let you know if this method should be merged or not once I see if it is going to work well. My way fixes a bunch of race conditions and allows full process death restoration and fixes numerous other bugs also. I am just mentioning here so this shouldnt be merged yet if this works out. I will update here within a couple days.

Luna712:
@fire-light42 there is a very easy much better fix for this I will push later.

PR #2756, opened 2026-05-03
Minor fixes to player

Luna712:
Fixes:

* Shadow overlay not properly hidden (except in trailer player)
* ~~Sometimes the play pause or fast forward/rewind buttons would be clicked again even after they had already hidden (usually only for about 0.2 seconds though)~~ (not doing in this for now as I don't think it is a proper fix)
* Duplicate `setOnClickListener` for `playerPausePlay` (one in `PlayerView` and one in `FullScreenPlayer`) - the one in `PlayerView` accidentally omitted the `isLayout(PHONE)` check, causing phone-only behavior to also apply on TV and Emulator
* Buffer icon sometimes overlaps play/pause in `FullScreenPlayer` (not trailer)
* 2x speed on trailer not hiding controls, an inconsistent behavior from `FullScreenPlayer`
* UI not showing after gesture end if it was visible before while paused (trailer-only)

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/FullScreenPlayer.kt
@@ -763,12 +763,10 @@ open class FullScreenPlayer : AbstractPlayerFragment<FragmentPlayerBinding>(
             }
         }
         playerBinding?.apply {
-
             playerLockHolder.isGone = isGone
             playerVideoBar.isGone = isGone
 
-            playerPausePlay.isGone = isGone
-            // player_buffering?.isGone = isGone
+            playerPausePlayHolderHolder.isGone = isGone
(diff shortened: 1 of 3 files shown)


fire-light42:
Please give a brief summery of what this is doing and why, so I can better review and test!

Luna712:
> Please give a brief summery of what this is doing and why, so I can better review and test!

@fire-light42

Done though would also like to note the isClickable fixes here I am unsure if that is the right way to fix it as while it stops it from play pausing etc... when controls are not showing, for about 0.2 seconds, clicking in the center were doesn't show the controls again, so that tells me the isClickable fix is only going to mask the issue and there is likely some other underlying issue. But I am not quite sure what.

**UPDATE**: I decided to remove that part for now, we can look into that issue more later.

fire-light42 (approved):

KenFangidae:
I found that this issue appears when seekbar preview is enabled.

PR #2758, opened 2026-05-04
Fix some bugs in DownloadedPlayerActivity

Luna712:


app/src/main/java/com/lagradost/cloudstream3/ui/player/DownloadedPlayerActivity.kt
```diff
@@ -59,8 +59,26 @@ class DownloadedPlayerActivity : AppCompatActivity() {
         setContentView(R.layout.empty_layout)
         Log.i(TAG, "onCreate")
 
-        handleIntent(intent)
-        attachBackPressedCallback("DownloadedPlayerActivity") { finish() }
+        // When savedInstanceState != null the system saved state before killing the
+        // backgrounded process the NavController restores the correct player fragment
+        // automatically. Do not replay the intent as it would push a duplicate player.
+        if (savedInstanceState == null) {
```
fire-light42: I am unable to fully test this until #2750 is fixed, as introducing a state where savedInstanceState != null causes that bug.

Luna712: I also could have done this part wrong. I was testing a lot of things in a massive local change, this 1) may depend on other things or 2) actually be totally unnecessary. Anyway #2761 fixes that bug. But that might still not let it work here. If not that part can be reverted.

Luna712: If the issue still happens with a null savedInstanceState I will revert that part. I was building this on top of change I was testing to use SavedStateHandle also, so may have had some other effect also. The other bugs this fixes are more important regardless.

Luna712: I removed the savedInstanceState check. That other issues are more important anyway and that has the potential to cause other issues anyway.

Luna712:
Could this potentially be merged @fire-light42 ?

Issue #2650, opened 2026-04-12
Full Name isn't showing due to logo


TamimZaman99:
### Describe your suggested feature

Problem: It's quite inconvenient for shows with long names and many prequel and sequel, Sometimes i have to go back and enter again just to check the full name of the Season

Request: Clicking the Logo should display full name/Option in settings to turn it off completely.

![Image](https://github.com/user-attachments/assets/454dc6f9-3340-406a-8a7b-f3b38facf2d0)

### Other details

Provider said to create the issue here.

### Acknowledgements

- [x] My suggestion is **NOT** about adding a new provider
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.

fire-light42:
A click toggle would be nice!

TamimZaman99:
Yeah.. in some cases it's hard to Read the text on logo

Before creating this issue, I had basically downloaded 'Danmachi.. Season 2" Thinking it's Season 1 due to Weird pattern design for "2"

It's not possible to see the full name at search because it's really long name

VyomVyas4765:
Hey @fire-light42, I am a new contributor and I would love to tackle this issue. Here’s my plan:

- I’ll add an on-click listener to the logo/banner image in the details screen.
- When the logo is clicked, it will toggle the visibility between the logo and the plain text name (full title) of the show.
- If the plain text name is showing, clicking it will bring the logo back.
- This will make it easy for users to quickly view the full name without navigating away, and then return to the original view.

Could you please assign this issue to me? I’m excited to contribute!

fire-light42:
@VyomVyas4765 Sounds like a great idea! Please also remember that we have a TV Layout as well, and it requires properties like `android:nextFocusRight` to navigate between elements. Also check out our policy https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md

PR #2760, opened 2026-05-04
Add toggle between logo and full title on click in details screen

VyomVyas4765:
Implements the feature for #2650 by adding a toggle on the logo to show the full title. Tapping the logo switches to the text title of the media and then tapping the title again brings the logo back. Nothing changes when there is no logo and the title itself is there. Tested and working as expected.

Screenshot_20260504-225722
Screenshot_20260504-225725


fire-light42 (changes requested):
Great first commit, however do not add additional logic that conflicts with existing code. We already have a `result_title` textview, and a `bindLogo` function. Modify those instead of adding new code. Moreover, while your `result_full_title_text` might have `nextFocus` to navigate from, it can not be navigated to.

VyomVyas4765:
Thanks for the feedback! That makes sense, I’ll refactor it to use result_title and bindLogo, and also fix the navigation issue.

app/src/main/java/com/lagradost/cloudstream3/ui/result/ResultFragment.kt
```diff
@@ -248,14 +248,70 @@ object ResultFragment {
...
         }
 
+        // Reset to the default state: logo visible, title hidden
+        logoView.alpha = 1f
         logoView.isVisible = true
+        titleView.alpha = 0f
         titleView.isVisible = false
+        titleView.isClickable = true
+        titleView.isFocusable = true
+        titleView.isFocusableInTouchMode = true
```
fire-light42: We normally do `isFocusableInTouchMode = isLayout(TV)` in CloudStream.

fire-light42 (changes requested):
Good improvements, however there are some visual errors that make this feature look weird.
 
1. The UI is bugged on TV, the focus looks really weird. It should be a clear outline, not a white background. 
2. The UI is not clickable on TV with touch in the results view.
3. As both views fight over the same layout location, the "animation" makes the UI move a lot. It is not seamless and has an affect on the entire layout tree.
4. It is not serialized, saved or is in any shape persistent. Normally this would be fine, but bindLogo may get run several times on the same view, and the view will get recreated when switching e.g. tabs or recommendations. This makes the UI feel forgetful, inconsistent and even buggy when using it on the main page on TV. A local session cache on the url would fix this.  
5. Using `.isEnabled` to store state will lead to inconsistent state, especially when using it in a recycleview.

VyomVyas4765:
The details you gave actually helped a bunch, appreciate it.  I've reworked the logic to fix the layout judder by dropping the animations, and hooked it up to a LruCache so the toggle state persists across the RecyclerView. I also dropped the .isEnabled hack to prevent state bleeding and cleaned up the TV focus outlines. Let me know if this looks better

app/src/main/res/layout/fragment_result_tv.xml
```diff
@@ -1104,4 +1097,32 @@ https://developer.android.com/design/ui/tv/samples/jet-fit
         </LinearLayout>
 
     </androidx.core.widget.NestedScrollView>
+
+    <!-- Logo overlay: must be AFTER the NestedScrollView so it sits above it
+         in the Z-order and can receive touch events for the toggle -->
+    <FrameLayout
```
fire-light43, line 1103: Now it does not scroll with the rest of the layout.

fire-light43 (changes requested):

PR #2764, opened 2026-05-05
Refactor: Player, Generator and ViewModel

### LINKED ISSUE #2746: No Links Found bug in DownloadedPlayerActivity

fire-light42:
Closes #2761 #2750 #2751
Fixes #2746

This pull requests refactors the following:
1. GeneratorPlayer to use the ViewModel state
    - Then the GeneratorPlayer does not have to worry about outdated state.
2. ViewModel to use an immutable state
    - No concurrency issues, and should be more efficient.
3. IGenerator to be stateless
    - The old behavior stored the index of what episode was being played inside the Generator, this was a bad design decision because we accessed it in a random order. 

Fixes:
1. "No Links Found" due to a mismatch of state between model and view
2. Possible race conditions, however there is still some in RepoLinkGenerator
3. View history for links not visited from the main player, e.g. external videos
4. Some UI issues that used invalid state

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -218,6 +218,7 @@ dependencies {
     implementation(libs.fragment.ktx)
     implementation(libs.bundles.lifecycle)
     implementation(libs.bundles.navigation)
+    implementation(libs.kotlinx.collections.immutable)
 
     // Design & UI
     implementation(libs.preference.ktx)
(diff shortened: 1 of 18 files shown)


fire-light42:
@Luna712 you might want to test this pull request, it should solve most of your issues ❤️

Luna712 (approved):
Nice! Seems to work. Thank you for doing this!

fire-light42:
Fixed concurrency issues in RepoLinkGenerator and currentQualityProfile cache for better UX.

Luna712:
@fire-light42 I think I found a major issue that seems to be related to this. It plays the wrong episode sometimes. The player said it was playing S0E1 even though it was like S20E36 or something. It happened after returning to the app after a couple hours and the player reloaded (probably process death) but previously when the player reloaded it would just load all links again on the right episode. I don't know how to reproduce exactly, but to me that indicates some potential major issue with this.

PR #2766, opened 2026-05-06
Fixobserveorder

fire-light42:
The problem with the old system was that observe might say that the link loading is finished when the viewModel.state disagrees, and this fixes that by dropping livedata

1. This should fix some issues with #2764 by dropping observe calls made with the wrong instance number. 
2. I also moved the loadLinks state change to not race the observe by having the state change immediately. 
3. The player also quits if it is fully recreated without showing a "no links found" if the generator does not exists. We need to serialize the generator if we want full relaunch capability, do note that this is not possible inside a bundle due to size limit.
4. And finally when the fragment is recreated it restarts the player from the same link to properly resync the UI.

I tested with:
```
adb shell am kill com.lagradost.cloudstream3.prerelease.debug
```
```
adb shell am force-stop com.lagradost.cloudstream3.prerelease.debug
```

```kotlin
playerGoBack.setOnClickListener {
    requireFragmentManager()
        .beginTransaction()
        .detach(this@FullScreenPlayer).commit()
    requireFragmentManager()
        .beginTransaction()
        .attach(this@FullScreenPlayer).commit()
}
```
```kotlin
playerGoBack.setOnClickListener {
    activity?.recreate()
}
```
``` Background process limit -> No background processes ```.

@Luna712

This should give:
Fragment recreated -> Reloads player
Activity recreated -> Reloads links (may use link cache)
Process killed -> Exists player

There might be some other possibility I am unaware of given some weird recreation order, but now the view-model should be in-sync at all times.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
@@ -2298,7 +2298,7 @@ class GeneratorPlayer : FullScreenPlayer() {
             }
 
             safe {
-                if (viewModel.state.links.any { link ->
+                if (!isPlayerActive.get() && viewModel.state.links.any { link ->
                         getLinkPriority(currentQualityProfile, link.first) >=
                                 QualityDataHelper.AUTO_SKIP_PRIORITY
                     }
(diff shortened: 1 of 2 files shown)


Luna712:
Nice! Thanks for this!

PR #2769, opened 2026-05-07
Add support for ksoup and migrate core extractors to it

Luna712:
This is just a test right now and is not ready for merging at all. At the minimum, before merging, `ksoupDocument` needs to be moved to NiceHttp directly and an alternative for `documentLarge` with Ksoup added also.


CranberrySoup:
I have already privately migrated NiceHttp to use Ksoup and Ktor but it is not backwards compatible. I was thinking of making a pull request to NiceHttp to keep the old NiceRequest and adding a generic networking client. After that we can create a new `app` with Ksoup and deprecate the old `app`.

PR #2785, opened 2026-05-12
fix: external player visibility on Android 11+

asheeshsahu7300:
Here is a detailed description you can use for the **fix: external player visibility on Android 11+** Pull Request:

---

# Description: Fix External Player Visibility on Android 11+ (Google TV 14)

This PR resolves the "App not found" error that occurs when users attempt to play content using external players (like VLC, MPV, etc.) on devices running Android 11 or higher, specifically including Google TV 14.

## The Problem
Starting with Android 11, the system introduced **Package Visibility** restrictions. Apps can no longer see other installed apps by default. While CloudStream requests the `QUERY_ALL_PACKAGES` permission, some device implementations (notably Google TV 14 and newer Android TV builds) are more restrictive or have specific intent-resolution behaviors that cause `ActivityNotFoundException` when trying to launch external apps via implicit intents.

This resulted in the "App not found" error even when players like VLC were correctly installed.

## The Solution
To ensure reliable intent resolution, this PR adds a `<queries>` block to the `AndroidManifest.xml`. This explicitly tells the Android system which intent patterns and packages CloudStream needs to interact with.

### Key Changes:
1.  **Intent Pattern Queries**: Added explicit queries for `VIEW` actions with `video/*`, `application/x-mpegURL` (m3u8), and `magnet` schemes. This allows the system to resolve any generic player that handles these types.
2.  **Explicit Package Queries**: Added specific package names for all external players currently supported by the app's `actions/temp` directory. This ensures that even if generic intent resolution is throttled, the app can still "see" and launch its known supported players.
3.  **Filtered List**: The list of queried packages is strictly limited to those with active implementations in the codebase to keep the manifest clean:
    *   **VLC**: `org.videolan.vlc`, `org.videolan.vlc.debug`
    *   **MPV Family**: `is.xyz.mpv`, `is.xyz.mpv.ext`, `is.xyz.mpv.ytdl`, `app.marlboroadvance.mpvex`
    *   **MPV-Kt**: `live.mehiz.mpvkt`, `live.mehiz.mpvkt.preview`
    *   **Others**: Just Player, Next Player, Web Video Cast
    *   **Torrent Clients**: LibreTorrent, BiglyBT

## Documentation
As per the AI usage policy, this fix was developed with AI assistance to identify the specific package names from the `actions/temp` directory and verify the correct intent filters required for Android 14 compatibility.

---

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -22,6 +22,47 @@
         android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="QueryAllPackagesPermission" />
 
+    <queries>
+        <!-- 
+            QUERY_ALL_PACKAGES does not work on some devices running Android 11+ (like Google TV 14), 
+            so we must explicitly specify the packages and intent patterns we query to ensure visibility.
+        -->
+        <!-- For external video players -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:mimeType="video/*" />
+        </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:mimeType="application/x-mpegURL" />
+        </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:mimeType="application/vnd.apple.mpegurl" />
+        </intent>
+        <intent>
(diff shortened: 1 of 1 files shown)


fire-light43 (approved):

fire-light43:
Thank you for the contribution!

PR #2791, opened 2026-05-14
Add support for using kotlinx-serialization rather than Jackson

Luna712:
If accepted, I will slowly migrate the app and library to kotlinx-serialization, standardizing all the different JSON parsers we use as well. kotlinx-serialization is fully KMP and also more performant, because unlike Jackson, it doesn't use runtime reflection, it uses validation at compiler time.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -5,42 +5,68 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
     val mapper: ObjectMapper = jacksonObjectMapper().configure(
         DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES,
         false
     )
 
+    val json = Json { ignoreUnknownKeys = true }
+
     override fun <T : Any> parse(text: String, kClass: KClass<T>): T {
-        return mapper.readValue(text, kClass.java)
+        val serializer = json.serializersModule.getContextual(kClass)
```
fire-light43: This always returns null right now, even for serializable classes. Try `kClass.serializerOrNull()`

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -5,42 +5,68 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
 // Short name for requests client to make it nicer to use
-private val jacksonResponseParser = object : ResponseParser {
+@OptIn(ExperimentalSerializationApi::class)
+private val jsonResponseParser = object : ResponseParser {
     val mapper: ObjectMapper = jacksonObjectMapper().configure(
         DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES,
         false
     )
 
+    val json = Json { ignoreUnknownKeys = true }
```
fire-light43: This should probably be global, to make the same mapper available everywhere in the app.

fire-light43 (commented):
A good start, but I want more insurance that this works when we migrate. Therefore I have created a comprehensive test to validate the move, please see if something like this can be included in androidTest in this pull request.

```kt
package com.lagradost.cloudstream3

import androidx.test.ext.junit.runners.AndroidJUnit4
import androidx.test.platform.app.InstrumentationRegistry
import com.lagradost.cloudstream3.utils.AppUtils.toJson
import dalvik.system.DexFile
import kotlinx.serialization.ExperimentalSerializationApi
import kotlinx.serialization.InternalSerializationApi
import kotlinx.serialization.KSerializer
import kotlinx.serialization.Serializable
import kotlinx.serialization.json.Json
import kotlinx.serialization.serializer
import kotlinx.serialization.serializerOrNull
import org.instancio.Instancio
import org.junit.Test
import org.junit.runner.RunWith
import kotlin.reflect.KClass
import kotlin.reflect.jvm.jvmName
import kotlin.test.assertEquals
import kotlin.test.assertNotNull

@RunWith(AndroidJUnit4::class)
class SerializationClassTester {
    // Same as app, or using app reference
    val jacksonMapper = mapper
    val kotlinxMapper = Json {
        ignoreUnknownKeys = true
    }

    @Test
    fun isIdenticalSerialization() {
        val serializableClasses = findSerializableClasses("com.lagradost")
        println("Number of serializable classes: ${serializableClasses.size}")

        serializableClasses.forEach { kClass ->
            val instance = Instancio.create(kClass.java)

            val jacksonJson = jacksonMapper.writeValueAsString(instance)
            val kotlinxJson = serializeWithKotlinx(kClass, instance)

            assertEquals(
                jacksonJson,
                kotlinxJson,
                """
                    Serialization mismatch for:
                    ${kClass.qualifiedName}

                    Jackson:
                    $jacksonJson

                    Kotlinx:
                    $kotlinxJson
                    
                    """.trimIndent()
            )
            println("Identical serialization for: ${kClass.jvmName}")
        }
    }

    @OptIn(InternalSerializationApi::class, ExperimentalSerializationApi::class)
    @Test
    fun isIdenticalDeserialization() {
        val serializableClasses = findSerializableClasses("com.lagradost")
        println("Number of serializable classes: ${serializableClasses.size}")

        serializableClasses.forEach { kClass ->
            val instance = Instancio.create(kClass.java)
            // Convert to JSON to get example JSON object
            // We prefer jackson here because the app may have many jackson JSON strings in local storage
            val originalJson = jacksonMapper.writeValueAsString(instance)

            // Create an object from the JSON using kotlinx
            val serializer =
                kClass.serializerOrNull() ?: kotlinxMapper.serializersModule.getContextual(kClass)
            assertNotNull(serializer, "The class: ${kClass.jvmName} must be serializable!")
            val kotlinxDecoded = kotlinxMapper.decodeFromString(serializer, originalJson)

            // Create an object from the JSON using jackson
            val mapperDecoded = jacksonMapper.readValue(originalJson, kClass.java)


            // Deep inspect both object using the mapper toJson function.
            // This deep equality check can be performed using other methods, but this just works.
            val jacksonJson = mapperDecoded.toJson()
            val kotlinxJson = kotlinxDecoded.toJson()

            assertEquals(
                jacksonJson,
                kotlinxJson,
                """
                    Serialization mismatch for:
                    ${kClass.qualifiedName}

                    Jackson:
                    $jacksonJson

                    Kotlinx:
                    $kotlinxJson
                    
                    """.trimIndent()
            )
            println("Identical deserialization for: ${kClass.jvmName}")
        }
    }

    private fun findSerializableClasses(packageName: String): List<KClass<*>> {
        val context = InstrumentationRegistry
            .getInstrumentation()
            .targetContext

        val dexFile = DexFile(context.packageCodePath)

        return dexFile.entries()
            .toList()
            .filter { it.startsWith(packageName) }
            .mapNotNull {
                runCatching { Class.forName(it).kotlin }
                    .onFailure { e ->
                        // Log.e("DEX", "Failed loading $it", e)
                    }
                    .getOrNull()
            }.filter { kClass ->
                kClass.java.annotations.any {
                    it is Serializable
                }
            }
    }

    @OptIn(InternalSerializationApi::class)
    @Suppress("UNCHECKED_CAST")
    private fun serializeWithKotlinx(
        kClass: KClass<*>,
        value: Any
    ): String {
        val serializer = kClass.serializer() as KSerializer<Any>
        return kotlinxMapper.encodeToString(serializer, value)
    }
}
``` 

I use 
```kt
    androidTestImplementation(kotlin("test"))
    androidTestImplementation("org.instancio:instancio-core:5.5.1")
    androidTestImplementation(libs.kotlinx.serialization.json)
```

Luna712:
The two suggestions are done. I also preemptively fixed issues that could have been caused during migration including bugs that broke backup and other stuff. I also added support to the rest of the app by using helper methods rather than mapper directly etc... I will add the test later also. For the record, for testing this I literally converted every single aspect of the app to kotlinx serialization and nothing broke now, but I will do smaller PRs later to convert sections of the app incrementally to make things a little easier to test later on.

Luna712:
@fire-light43 I added the test now with some minor changes to what you had above to fix some deprecation warnings and a couple other things. I did #2792 so these tests run when making PRs also, since I plan to do those to start migrating a lot of things I thought that would make things easier also.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStore.kt
```diff
@@ -87,9 +92,38 @@ data class Editor(
...
+            }
+        } else {
+            mapper.readValue(this, kClass.java)
+        }
+    }
+
+    private fun anyToJsonString(obj: Any): String {
+        // @Serializable generates a serializer at compile time; contextual serializers are
+        // registered manually in serializersModule, we need both to support all cases
+        val serializer = obj::class.serializerOrNull() ?: json.serializersModule.getContextual(obj::class)
```
fire-light43: This logic is now replicated 3 times in different files. We should have one singular internal implementation all other methods use.

app/src/main/java/com/lagradost/cloudstream3/utils/serializer/UriSerializer.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import android.net.Uri
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+object UriSerializer : KSerializer<Uri> {
```
fire-light43: Shouldn't we register this serializer globally?

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -74,6 +75,8 @@ const val USER_AGENT =
 class ErrorLoadingException(message: String? = null) : Exception(message)
 
 //val baseHeader = mapOf("User-Agent" to USER_AGENT)
+
+val json = Json { ignoreUnknownKeys = true }
```
fire-light43: Probably good to have a `@Prerelease` annotation here

fire-light43: I think also using `encodeDefaults = true` better matches our current jackson

fire-light43 (commented):
Looks good, just some minor things. It looks ready to merge very soon.

Good choice going with ClassGraph 👍

app/src/main/java/com/lagradost/cloudstream3/utils/serializer/UriSerializer.kt
Luna712: Would rather not. We only need it in two places. But if you think we should no issues from me.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStore.kt
Luna712: Yeah I thought about that but it needed done a little differently here so I just replicated it. I will see about combining it though. I was mostly just copying how mapper was done before which we had it copied way to many times as well I think.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: Good idea, thanks for that.

Luna712: Good idea, thanks for that.

Luna712:
I will do changes tomorrow more than likely rather than tonight. Thanks for the review again!

app/src/main/java/com/lagradost/cloudstream3/utils/serializer/UriSerializer.kt
fire-light43: Do what you think is best, but an explanatory comment might be good if we limit the usage of this.

fire-light43:
No problem, this is already very fast progress

app/src/main/java/com/lagradost/cloudstream3/utils/serializer/UriSerializer.kt
Luna712: Absolutely fair. I will add a comment. Very easy to use though for example:
```kt
    @Serializable
    data class MinimalVideoLink(
        @SerialName("uri")
        @Serializable(with = UriSerializer::class)
        val uri: Uri?,
        ...
    )
```
is one of the only two places we will need it. I will describe this in a comment though.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt

app/src/main/java/com/lagradost/cloudstream3/utils/serializer/UriSerializer.kt
Luna712: Added comment.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStore.kt
Luna712: Unified logic

Luna712:
@fire-light43 done with all that now, I also added two more serializers to do things we currently do with Jackson, and also added another test to ensure all the serializers are properly working.

P.S. I may be writing more tests in the future for things now. You kinda got me in to liking to write those lol.

app/src/androidTest/java/com/lagradost/cloudstream3/SerializationClassTester.kt
```diff
@@ -0,0 +1,127 @@
...
+import kotlin.test.assertNotNull
+
+@RunWith(AndroidJUnit4::class)
+class SerializationClassTester {
+    // Same as app, or using app reference
+    val jacksonMapper = mapper
+    val kotlinxMapper = json
+
+    @Test
+    fun isIdenticalSerialization() {
```
Luna712, line 28: hmm... I just thought of something... unless I am missing something.... this may not actually match a lot of times because things will be different when we swap `@JsonProperty` with `@SerialName` and a couple of other changes.

Luna712: I guess we could keep `@JsonProperty` for now until migration is complete and then remove them all so we would have both `@JsonProperty` with `@SerialName` in the meantime, which this may be better anyway to ensure the fallback also properly works if something goes wrong with kotlinx serialization during migration as well...

fire-light43: Yes, it will flag all issues with `@JsonProperty`. I was also thinking we keep all  `@JsonProperty` until we fully remove jackson, and when we do we can change/remove this test. It was made to catch any faulty or missing`@JsonProperty` migration.

Luna712: Yep that makes sense. We can have both `@JsonProperty` and `@SerialName` in the meantime.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStore.kt
```diff
@@ -165,17 +173,17 @@ object DataStore {
     fun <T> Context.setKey(path: String, value: T) {
         try {
             getSharedPrefs().edit {
-                putString(path, mapper.writeValueAsString(value))
+                putString(path, value?.let { it.toJson() })
```
fire-light42: This causes a bug when saving strings, toJson and writeValueAsString are not identical.

```kt
data class TestObj(
    val value: String
)
val obj = TestObj("test")
println("mapper: ${mapper.writeValueAsString(obj.toJson())}")
println("toJson: ${obj.toJson().toJson()}")
```
Gives:
```
mapper: "{\"value\":\"test\"}"
toJson: {"value":"test"}
```

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/utils/DataStore.kt
Luna712: This *should* be fixed as it is now functionally the exact same thing by using a new toJsonLiteral that removes the is String check for it. It seems this is the only place that matters though as its the only place that it might already be a String rather than an object that hsed to use writeValueAsString directly. writeValueAsString in the NiceHttp responseParser doesn't matter either since NiceHttp has its own built in check and it checks if its a string and just returns String before running writeValueAsString at all anyway. Hopefully that all works well now though.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1,46 +1,40 @@
 package com.lagradost.cloudstream3
 
-import com.fasterxml.jackson.databind.DeserializationFeature
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.lagradost.cloudstream3.utils.AppUtils.parseJson
+import com.lagradost.cloudstream3.utils.AppUtils.toJson
```
fire-light42, line 4:
```suggestion
import com.lagradost.cloudstream3.utils.AppUtils.toJsonLiteral
```

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1,46 +1,40 @@
...
-        } catch (e: Exception) {
+            parse(text, kClass)
+        } catch (_: Exception) {
             null
         }
     }
 
     override fun writeValueAsString(obj: Any): String {
-        return mapper.writeValueAsString(obj)
+        return obj.toJson()
```
fire-light42, line 24:
```suggestion
        return obj.toJsonLiteral()
```
To match previous behavior with Strings.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -1,21 +1,49 @@
...
+            try {
+                json.decodeFromString(serializer, value)
+            } catch (e: SerializationException) {
+                logError(e)
+                mapper.readValue(value)
+            }
+        } else {
+            mapper.readValue(value)
+        }
     }
```
fire-light42, line 25:
```suggestion
    inline fun <reified T : Any> parseJson(value: String): T {
        return parseJson(value, T::class)
    }
```
I dislike having the parsing logic duplicated, better to just call the other internal function.

app/src/androidTest/java/com/lagradost/cloudstream3/SerializationClassTester.kt
```diff
@@ -0,0 +1,127 @@
...
+                    Kotlinx:
+                    $kotlinxJson
+                    
+                    """.trimIndent()
+            )
+            println("Identical deserialization for: ${kClass.jvmName}")
+        }
+    }
+
+    private fun findSerializableClasses(packageName: String): List<KClass<*>> {
```
fire-light42, line 103: I have no idea why or how but this function no longer returns anything, even if I have Serializable classes. I am sure it worked before. Do you get classes here? Even allClasses is empty.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/serializers/NonEmptySerializer.kt
```diff
@@ -0,0 +1,45 @@
...
+ *   @Serializable(with = MyData.Serializer::class)
+ *   data class MyData(
+ *       val tags: List<String> = emptyList(),
+ *       val title: String = "",
+ *       val meta: Map<String, String> = emptyMap(),
+ *   ) {
+ *       object Serializer : NonEmptySerializer<MyData>(MyData.generatedSerializer())
+ *   }
+ */
+abstract class NonEmptySerializer<T : Any>(tSerializer: KSerializer<T>) :
```
fire-light42, line 32: Good to have `@Prerelease` here

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/serializers/WriteOnlySerializer.kt
```diff
@@ -0,0 +1,36 @@
...
+ *       val fieldA: String = "",
+ *       val fieldB: String = "",
+ *   ) {
+ *       object Serializer : WriteOnlySerializer<MyData>(
+ *           MyData.generatedSerializer(),
+ *           setOf("fieldB"),
+ *       )
+ *   }
+ */
+abstract class WriteOnlySerializer<T : Any>(
```
fire-light42, line 29: Good to have `@Prerelease` here too

fire-light42 (changes requested):
Everything looks good, just some small changes before merge.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainActivity.kt
Luna712: That shouldn't be necessary here, as NiceHttp returns a string without calling this method if it is already a string, but I can make the change if you want?

fire-light42: Good catch! In that case it actually makes more sense to keep it as `toJson()`. 
Looks like I missed your previous comment about this :disappointed:

fire-light42:
Everything looks good, I will merge tomorrow after testing one last time.
Not sure why the tests are suddenly not working for me, but they are not blockers for this pull request.

Luna712:
Sounds great thank you! I will see what I can do with tests in the meantime though

app/src/androidTest/java/com/lagradost/cloudstream3/SerializationClassTester.kt
Luna712: The issue is because of classgraph it seems. My bad about that. Seems it doesn't support searching in dex real great. I didn't want to use DexFile because of deprecation warnings but that may realistically be the only way to fix this I am not certain though.

fire-light42 (approved):
Very good job on finally migrating to kotlinx serialization!

fire-light42 (approved):
Very good job on finally migrating to kotlinx serialization!

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
Luna712: This caused #2851 by the way, we could probably revert this part but it was faster for me to just fix the updater itself for now and look more into this later.

PR #2792, opened 2026-05-15
Support running instrumented tests with a comment in pull requests

Luna712:
I was originally going to do this in #2791 when adding the test for that but decided it would be better as a separate PR.


app/src/androidTest/java/com/lagradost/cloudstream3/ExampleInstrumentedTest.kt
```diff
@@ -55,8 +56,18 @@ class ExampleInstrumentedTest {
...
+    private fun skipIfNoProviders() {
+        Assume.assumeTrue(
+            "Skipping: no providers loaded",
+            getAllProviders().isNotEmpty()
+        )
+    }
+
     @Test
     fun providersExist() {
+        skipIfNoProviders()
```
fire-light42: Better to just remove `providersExist` to prevent confusion in the future.

Luna712: Done

.github/workflows/pull_request.yml
```diff
@@ -34,3 +34,50 @@ jobs:
       with:
         name: pull-request-build
         path: "app/build/outputs/apk/prerelease/debug/*.apk"
+
+  instrumented-tests:
```
fire-light42: This looks to be heavy CPU/time-wise, but the tests are still extremely valuable in some specific cases (like all of your recent pull requests). Would it be possible to limit these tests to manual-run/tagged/reviewed or something else to prevent running these tests for every single commit in every single pull request? It would simply be a waste otherwise.

Luna712: Apparently running the other serializer tests in the emulator fails also, so I need to do some more work to fix that now. I will do so later on. This isn't fully urgent in the meantime, but I will get it done in the next day or two.

Luna712: #2902 fixed the test failure here also btw.

Luna712: @fire-light42 also the reason I made instrumentation tests a separate task is to allow ignoring it and not delay actual archive build. I figured it might also be useful like for the layout binding tests for now also, and other things. If you want though I can change its trigger though yeah.

Luna712: I have created a new workflow for this by adding a comment `/run-tests`. It won't do anything until merged though. It is security gated also. I intentionally allowed myself along with members or owners of the org, but I can remove that part if necessary.

PR #2794, opened 2026-05-17
fix not auto selecting subtitle

IstarVin:
player auto selecting subtitle is not working after the refactor commit [a45f1d9ab1edc165941d01996cda92954f99852e](https://github.com/recloudstream/cloudstream/tree/a45f1d9ab1edc165941d01996cda92954f99852e)


LagradOst (approved):
Nice catch!

PR #2795, opened 2026-05-17
Add selectable cross-provider source aggregation

onurcvnoglu:
## Summary

Adds an optional provider setting to aggregate playable sources from selected installed extensions when loading a movie or episode.

## Changes

- Added a new provider setting: "Search sources in all extensions"
- Added a provider selection preference for choosing which extensions can be merged
- Limited cross-provider source aggregation to:
  - max 5 selected providers
  - max 5 video sources per provider
- Reused the existing `RepoLinkGenerator` cache/deduplication flow
- Added matching logic by title, media type, year, season, and episode
- Skips WebView-based providers for alternate source aggregation to avoid native Chromium crashes
- Stops provider source loading after the source limit is reached
- Added English and Turkish strings
- Added local implementation notes under `docs/local/` excluded from Git

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/APIRepository.kt
@@ -20,9 +20,12 @@ import com.lagradost.cloudstream3.newSearchResponseList
 import com.lagradost.cloudstream3.utils.Coroutines.threadSafeListOf
 import com.lagradost.cloudstream3.utils.ExtractorLink
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CancellationException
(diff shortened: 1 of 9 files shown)


LagradOst:
1. Read our AI policy https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md
2. This can and should be implemented on the provider level, not on the entire app. A good cross provider helper is much better than this, as that is opt in for the extension developer. 
3. This breaks so many design decisions and will result in very weird errors, making the entire system unreliable.
4. This makes loading time worse, and makes more get requests. This will make the CPU work more for less.
5. The code is very wet, and moves the responsibility of things to the wrong place. 
6. Limits link loading count with arbitrary counts and timeouts.
7. No documentation or motivation. 

Please make an extension instead.

PR #2796, opened 2026-05-17
Add shared link routing

IstarVin:
- Add ShareLinkActivity to receive shared text and HTTP(S) link intents.
- Route supported provider links directly to Result View.
- Route supported extractor links directly to the player.
- Wait for plugin loading before showing unsupported-link feedback.
- Harden shared URL parsing, matching, and duplicate intent handling.

used AI


LagradOst (changes requested):
This functionallity already exists in https://github.com/recloudstream/cloudstream/blob/03eb6ccd45fe84baf0cf31cd3a2c42fea13ad70a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt#L279

Only "extractor links" is not implemented, so implement that here instead, it should be a 10line change.

IstarVin:
the current implementation only handles links with "csshare:" which is from the built in share feature from the app.
this pull request would add an entry in the android share link which would accept direct https link which the app would match to its currently registered providers and redirect to the correct result view.

app/src/main/AndroidManifest.xml
```diff
@@ -98,6 +98,29 @@
             android:shell="true"
             tools:targetApi="q" />
 
+        <activity
```
fire-light43: Why do we need another activity? This will make everything more complicated, for example logging in to accounts.

Please use the existing link handler instead of creating another activity

fire-light43 (changes requested):

PR #2798, opened 2026-05-17
Migrate Java date utils in library to kotlinx-datetime

Luna712:
I wrote unit tests for this as well as it was hard for me to manually test every possible supported format to ensure it works properly supporting the same formats this supported before. The tests caught a couple issues in my initial version before them where some dates didnt work properly and were silently ignored but it should all work now supporting same formats that were supported before.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -2512,15 +2521,41 @@ constructor(
...
+@OptIn(FormatStringsInDatetimeFormats::class)
 fun Episode.addDate(date: String?, format: String = "yyyy-MM-dd") {
-    try {
-        this.date = SimpleDateFormat(format, Locale.getDefault()).parse(date ?: return)?.time
-    } catch (e: Exception) {
-        logError(e)
-    }
+    if (date == null) return
+    this.date = runCatching {
+        val fmt = DateTimeComponents.Format { byUnicodePattern(format) }
```
fire-light42: It looks like the new formatter is more strict than `SimpleDateFormat`.

```kt
SimpleDateFormat("yyyy-MM-dd", Locale.getDefault()).parse("2026-01-01T12:30:00.000Z")
```

Gives a valid date whereas `DateTimeComponents` fails. Would it be possible to make the new formatter less strict, even if the format is incorrect?

Luna712: That could be achieved by passing the addDate with the Instant format, but I will support it here also.

Luna712: I also covered that exact case in tests.

fire-light42 (approved):
Looks good! Great tests!

PR #2800, opened 2026-05-18
Rewrite TmdbProvider to remove external lib dependency

Luna712:
~~As a side-effect, this also removes all gson usage from extractors and uses AppUtils methods, because gson was only an implicit dependency from the external tmdb library.~~

**Update**: I ended up splitting the gson changes out to a seperate PR.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/TmdbProvider.kt
```diff
@@ -67,219 +54,363 @@ open class TmdbProvider : MainAPI() {
...
+    // As some sites don't support s0
     open val disableSeasonZero = true
 
     override val hasMainPage = true
     override val providerType = ProviderType.MetaProvider
 
-    // Fuck it, public private api key because github actions won't co-operate.
-    // Please no stealy.
-    private val tmdb = Tmdb("e6333b32409e02a4a6eba6fb7ff866bb")
+    private val tmdbApiKey = "e6333b32409e02a4a6eba6fb7ff866bb"
```
Luna712, line 67: This should work in GH actions as a secret now as well (done just like MDL). We should probably move it there.

fire-light43:
Will this remove all instances of gson? Because I think many extensions are using gson.

Luna712:
> Will this remove all instances of gson? Because I think many extensions are using gson.

gson was never officially supported, extensions should never have been using it. There is also no way to actually *deprecate* this so there is no way to fully communicate this. In order to facilitate eventual jackson -> kotlinx serialization, extensions should also first be using Jackson/AppUtils methods to make it easier. There really isnt a better way to communicate it. I dont know if this fully breaks extensions that uses gson or not, but IMO even if so the outcome would eventually be the same, they either break and have to migrate now or do so later. No official example or documentation said to use gson ever, so there was always the *correct* alternative.

Will do what you want to do with this though. Again not certain if it breaks existing by removing the implicit dependency here though.

fire-light43:
> > Will this remove all instances of gson? Because I think many extensions are using gson.
> 
> gson was never officially supported, extensions should never have been using it. There is also no way to actually _deprecate_ this so there is no way to fully communicate this. In order to facilitate eventual jackson -> kotlinx serialization, extensions should also first be using Jackson/AppUtils methods to make it easier. There really isnt a better way to communicate it. I dont know if this fully breaks extensions that uses gson or not, but IMO even if so the outcome would eventually be the same, they either break and have to migrate now or do so later. No official example or documentation said to use gson ever, so there was always the _correct_ alternative.
> 
> Will do what you want to do with this though. Again not certain if it breaks existing by removing the implicit dependency here though.

"gson was never officially supported", Yes but extension developers sure do love using it. I am uncomfortable breaking the CI/CD of like half the extensions overnight without warning. 

https://github.com/search?q=repo%3Aphisher98%2Fcloudstream-extensions-phisher%20gson&type=code
https://github.com/search?q=repo%3ASaurabhKaperwan%2FCSX%20gson&type=code
https://github.com/search?q=repo%3AAbodabodd%2Fre-3arabi%20gson&type=code
https://github.com/search?q=repo%3ATeKuma25%2FIndoStream%20gson&type=code
https://github.com/search?q=repo%3AAsm0d3usX%2FCloudX-V2%20gson&type=code
https://github.com/search?q=repo%3ANivinCNC%2FCNCVerse-Cloud-Stream-Extension%20gson&type=code
https://github.com/search?q=repo%3AGian-Fr%2FItalianProvider%20gson&type=code
https://github.com/search?q=repo%3ASushan64%2FNetMirror-Extension%20gson&type=code

Luna712:
> > > Will this remove all instances of gson? Because I think many extensions are using gson.
> > 
> > 
> > gson was never officially supported, extensions should never have been using it. There is also no way to actually _deprecate_ this so there is no way to fully communicate this. In order to facilitate eventual jackson -> kotlinx serialization, extensions should also first be using Jackson/AppUtils methods to make it easier. There really isnt a better way to communicate it. I dont know if this fully breaks extensions that uses gson or not, but IMO even if so the outcome would eventually be the same, they either break and have to migrate now or do so later. No official example or documentation said to use gson ever, so there was always the _correct_ alternative.
> > Will do what you want to do with this though. Again not certain if it breaks existing by removing the implicit dependency here though.
> 
> "gson was never officially supported", Yes but extension developers sure do love using it. I am uncomfortable breaking the CI/CD of like half the extensions overnight without warning.
> 
> https://github.com/search?q=repo%3Aphisher98%2Fcloudstream-extensions-phisher%20gson&type=code https://github.com/search?q=repo%3ASaurabhKaperwan%2FCSX%20gson&type=code https://github.com/search?q=repo%3AAbodabodd%2Fre-3arabi%20gson&type=code https://github.com/search?q=repo%3ATeKuma25%2FIndoStream%20gson&type=code https://github.com/search?q=repo%3AAsm0d3usX%2FCloudX-V2%20gson&type=code https://github.com/search?q=repo%3ANivinCNC%2FCNCVerse-Cloud-Stream-Extension%20gson&type=code https://github.com/search?q=repo%3AGian-Fr%2FItalianProvider%20gson&type=code https://github.com/search?q=repo%3ASushan64%2FNetMirror-Extension%20gson&type=code

So then how do we even communicate it will be removed? If we keep it more and more usage will keep being added as well. We cant exactly deprecate it either. But I can add it back as an explicit dependency if you want? Still don't really like that by using this lib we accidentally had an implicit dependency causing a diversion in what should be used. But your right we probably should do this a little different then sudden removal. Not sure exactly best route though.

Luna712:
Why dont we just communicate the best we can and remove it in 1-2 months? I plan to have the majority of my current efforts finished within the time frame, so want to push it fairly quickly if possible. But I dont think we need to keep it for a long time. We can't really follow a "deprecation" schedule for this and alternative has always existed.

fire-light43:
> So then how do we even communicate it will be removed? If we keep it more and more usage will keep being added as well. We cant exactly deprecate it either. But I can add it back as an explicit dependency if you want? Still don't really like that by using this lib we accidentally had an implicit dependency causing a diversion in what should be used. But your right we probably should do this a little different then sudden removal. Not sure exactly best route though.

For this pull request, an explicit dependency would be best. However, for future development I think an issue on each repository should at least be a start. We can not communicate this over the deprecated API, but we can try to figure out the affected developers. 

Otherwise we might have to make a custom lint to warn about this. However, I am unsure if this would be possible for this usecase.

Luna712:
Absolutely fair. A lint rule is actually what I was also thinking to facilitate jackson migration as well.

Luna712:
@fire-light43 I added gson without adding to catalog so that removal is just removing once since we don't intend to keep it to long. I will also add issues on each repo either tonight or tomorrow.

Luna712:
I will fix this in a bit. This is the only one I hadn't ran the tests on it before, but it exposed some issues with it.

Luna712:
> I will fix this in a bit. This is the only one I hadn't ran the tests on it before, but it exposed some issues with it.

Now fixed.

app/build.gradle.kts
```diff
@@ -276,6 +276,8 @@ dependencies {
     implementation(libs.jackson.module.kotlin) // JSON Parser
     implementation(libs.zipline)
 
+    // Temp/deprecated; will be removed once extensions have time to migrate from using it
+    implementation("com.google.code.gson:gson:2.14.0")
```
fire-light42: This should be 2.11.0 to match the TMDB dependency. Since we are removing it later we might as well have an older version to maximize compatibility.

Luna712: Done, thanks!

fire-light42 (approved):
Work perfectly!

PR #2801, opened 2026-05-18
Add support for custom app fonts, resolves #2747

### LINKED ISSUE #2747: using custom font for the app itself

I mean, it should contain a dropdown or a field in which we can type whatever font we want and set all the text of app to that font (except subtitles). Honestly, I would prefer both field and dropdown

### Other details

We can use [coolLabs](https://fonts.coollabs.io) for the Font API (Privacy friendly wrapper over Google Font)
I can work on it if maintainers approve this type of feature

### Acknowledgements

itsmeadarsh2008:
## Summary (ok to go)
- add runtime support for downloading and applying custom app fonts from the coolLabs font API
- add a UI setting with both suggestions and a free-text font family field
- keep subtitle rendering untouched while applying the selected font across the rest of the app UI

## Notes
- fonts are cached locally after download
- the branch history is split into runtime support and settings UI commits

**Note**: Codex is used to configure woff2-android and its unstability issues and AppFontManager.kt


itsmeadarsh2008:
There are some UI bugs, fixing them rn. Setting priority order for fonts. Requires app restart to change the font.

itsmeadarsh2008:
Update: doesn't need a restart for suggestive fonts, while if you bring your own font name from Google Fonts, then you will need a restart for it. Also, I have made some UI changes because the name was overlapping the border.

fire-light43 (changes requested):
I like your idea, and your code likely works. However, constantly observing the view tree to replace all typefaces is unacceptable. It can (and will) cause performance issues, edge cases and all sorts of weird bugs.

If you __must__ do runtime fonts then look into using a custom layout inflater. It would apply the font before it's rendered, which would bypass many of the issues caused by runtime layout changes. However, any such change would need to be maintainable and with a minimal performance overhead.

I would much prefer if you just use custom styles with a number of preset fonts. You can still use the font manager to let the user download subtitle fonts.

fire-light43:
cat-work-in-progress

itsmeadarsh2008:
> I like your idea, and your code likely works. However, constantly observing the view tree to replace all typefaces is unacceptable. It can (and will) cause performance issues, edge cases and all sorts of weird bugs.
> 
> If you **must** do runtime fonts, then look into using a custom layout inflater. It would apply the font before it's rendered, which would bypass many of the issues caused by runtime layout changes. However, any such change would need to be maintainable and with a minimal performance overhead.
> 
> I would much prefer it if you just use custom styles with a number of preset fonts. You can still use the font manager to let the user download subtitle fonts.

Bro, I tried a few solutions locally, all of them broke. I tried removing "no restart app font change", then that also broke. When I restarted, it didn't change the font. 2nd solution worked for some time, but only while using the app, didn't save to disk. So, restarting the app resets to the default font. 3rd solution, didn't even work lmao.

This is the only solution that I have found to work.

While testing, I didn't face any issues with performance. What are the actual constraints for the minimum supported systems? I think it will also work seamlessly on a 2GB RAM Android TV, too.
image

fire-light43:
> 2nd solution worked for some time, but only while using the app, didn't save to disk.

That just sounds like a matter of fixing the saving mechanism

> 3rd solution, didn't even work lmao.

Styles can be confusing, but I am certain they can work. If you need help implementing any specific solution just ask here or in the discord.

> While testing, I didn't face any issues with performance. What are the actual constraints for the minimum supported systems? I think it will also work seamlessly on a 2GB RAM Android TV, too.

We are aiming to support old hardware running non-standard Android 6, but text rendering is an issue even on recent devices. A big part of rendering the home page consists of just rendering the text, which means any extra overhead will noticeably impact scrolling smoothness of the home page.

fire-light42:
We are now [moving to compose](https://github.com/recloudstream/cloudstream/pull/2829), which likely has much greater support for custom fonts. All of this is very relevant after our migration.

PR #2802, opened 2026-05-18
Don't use any external library for Levenshtein ratio matching

Luna712:


app/build.gradle.kts
```diff
@@ -255,7 +255,6 @@ dependencies {
     // Extensions & Other Libs
     implementation(libs.jsoup) // HTML Parser
     implementation(libs.rhino) // Run JavaScript
-    implementation(libs.fuzzywuzzy) // Library/Ext Searching with Levenshtein Distance
```
fire-light42, line 258: Some extensions are using fuzzywuzzy, we need to give them time to migrate to `Levenshtein`. I was thinking we keep fuzzywuzzy in the app for some time, but remove it from the library when we do the final push for KMP.

Luna712: Extensions use things I never thought they really would. Anyway, readded.

library/build.gradle.kts
```diff
@@ -57,7 +57,6 @@ kotlin {
             implementation(libs.nicehttp) // HTTP Lib
             implementation(libs.jackson.module.kotlin) // JSON Parser
             implementation(libs.kotlinx.coroutines.core)
-            implementation(libs.fuzzywuzzy) // Match Extractors
```
fire-light42, line 60: Sorry, I was too vauge in my latest comment. I meant that we keep fuzzywuzzy as a dependency in both app and library for now. However, we remove fuzzywuzzy from the library first, and we do that as one of the last pushes before a fully KMP library. This way extension makers can have the maximum time to migrate.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/Levenshtein.kt
```diff
@@ -0,0 +1,513 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.lagradost.cloudstream3.utils
+
+import kotlin.math.round
+
+// Taken from https://github.com/terrakok/FuzzyKot/blob/f794d43/fuzzykot/src/commonMain/kotlin/com/github/terrakok/fuzzykot/Levenshtein.kt
+object Levenshtein {
```
fire-light42, line 32: Don't forget the `@Prerelease` annotations

library/build.gradle.kts
Luna712: Oh my bad. Done. Though I didnt think that was necessary since the dependencies in library dont go down to extensions when they only use tbe library, they have to explicitly add it also, and it being in app is enough to prevent runtime errors if they use it.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/Levenshtein.kt
Luna712: Done. Keep forgetting that sorry thanks!

fire-light42 (approved):
Looks great!

PR #2807, opened 2026-05-19
Replace charset string conversions with Kotlin native equivalents

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/ByseSX.kt
```diff
@@ -94,7 +93,7 @@ open class ByseSX : ExtractorApi() {
         cipher.init(Cipher.DECRYPT_MODE, secretKey, spec)
 
         val plainBytes = cipher.doFinal(cipherBytes)
-        var jsonStr = String(plainBytes, StandardCharsets.UTF_8)
+        var jsonStr = String(plainBytes, Charsets.UTF_8)
```
fire-light43: Would it not make more sense to use `ByteArray.decodeToString()` since that is multiplatform unlike `Charsets`?

https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-byte-array#2071674355%2FFunctions%2F1935454986

Luna712: This is now fixed, here and everywhere else in library. Or at least should be.

fire-light42 (approved):

PR #2808, opened 2026-05-19
Replace synchronized usage in library with kotlinx-atomicfu

Luna712:


app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
```diff
@@ -689,22 +689,13 @@ object PluginManager {
...
         }
 
-        synchronized(extractorApis) {
-            extractorApis.removeIf { provider: ExtractorApi -> provider.sourcePlugin == plugin.filename }
-        }
-
-        synchronized(VideoClickActionHolder.allVideoClickActions) {
-            VideoClickActionHolder.allVideoClickActions.removeIf { action: VideoClickAction -> action.sourcePlugin == plugin.filename }
-        }
+        APIHolder.allProviders.removeAll { provider -> provider.sourcePlugin == plugin.filename }
```
fire-light42: A `ConcurrentModificationException` can happen in all of these `removeAll` due to it iterating over the list.

fire-light42 (changes requested):

app/src/main/java/com/lagradost/cloudstream3/plugins/PluginManager.kt
Luna712: Shouldn't anymore because we have removeall synchronized via AtomicMutableList, unless I am missing something?

fire-light42: The `removeAll` is pointing into the Kotlin library than your function :shrug: 

```kt
@IgnorableReturnValue
public fun <T> MutableList<T>.removeAll(predicate: (T) -> Boolean): Boolean = filterInPlace(predicate, true)
```

Luna712: My goal was to make it not need withLock or synchronized as much to make callers that forget not have ConcurrentModificationException as it could be handled by AtomicMutableList itself. At least that was the idea. Perhaps I totally failed with that goal if it doesnt actually work to prevent that...

Luna712: What should I do here then? Change it to withLock and remove that part from AtomicMutableList?

fire-light42: The issue here is that the AtomicMutableList method is `removeAll(elements: Collection<T>)` while the kotlin method is `.removeAll(predicate: (T) -> Boolean)`. 

Your design is correct and I like that synchronized is not required for basic operations. For this I think it is okay with `withLock { }` since we intentionally do not synchronize iterators.

Luna712: Right... that makes a lot of sense... my bad. Thanks a lot. Done.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AtomicList.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    operator fun plus(element: T): AtomicList<T> = synchronized(this) { AtomicList(delegate + element) }
+    operator fun plus(elements: Collection<T>): AtomicList<T> = synchronized(this) { AtomicList(delegate + elements) }
+}
+
+class AtomicMutableList<T>(
+    delegate: MutableList<T> = mutableListOf()
+) : AtomicList<T>(delegate), MutableList<T> {
+
+    private val mutableDelegate get() = delegate as MutableList<T>
```
fire-light42: Not a fan of this cast, it causes unexpected exceptions when used like this:

```kt
val list = AtomicMutableList(mutableListOf(1))
list.set(emptyList())
list.add(1)
```

I think we should just remove the set method, make the AtomicList delegate a val and change the constructor to something like this in AtomicMutableList:

```kt
class AtomicMutableList<T>(
    private var mutableDelegate: MutableList<T> = mutableListOf()
) : AtomicList<T>(mutableDelegate), MutableList<T> {
```  

That makes it easier to reason about the code and avoids weird exceptions.

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AtomicList.kt
Luna712: This is now done. Thank you!

fire-light42 (approved):

PR #2813, opened 2026-05-20
Migrate crypto from javax to cryptography-kotlin in library

Luna712:


fire-light42 (commented):
These are a lot of changes which are difficult to properly test and would give silent errors if incorrectly ported. 

Would it be possible to do some sorts of tests, or anything like that to ensure the migration won't cause regressions?

Luna712:
> These are a lot of changes which are difficult to properly test and would give silent errors if incorrectly ported.
> 
> Would it be possible to do some sorts of tests, or anything like that to ensure the migration won't cause regressions?

@fire-light42 

I'm not sure how to really write reliable tests for this. In order to really write something that would ensure things are proper it would have to be given the right data, and test way more than I really know how to write them for myself. Trying to write tests for some other migrations the best I could but for this one I just didn't really know how unfortunately.

fire-light42:
It is outside the scope of this pull request, but the extractors themselves could be tested with mock data :thinking:

fire-light42:
Do not worry about the tests, I will try to test them manually before merging.

Luna712:
Sounds great thanks a lot!

fire-light42 (approved):
Very difficult to test all possible cases, but the code looks good and it passed all tests I could do.

PR #2815, opened 2026-05-20
Remove back-compat constructor for AnimeSearchResponse

Luna712:
This is in preparation to remove the use of EnumSet for `dubStatus`. Those that use the builder, use `addDubStatus` which means we can easily do this without breaking bytecode compatibility, but first we need to remove the back-compat constructor, as once we do that it wouldn't work anymore anyway.


fire-light42:
This should be ok to merge, given that I did not find any repository that is not 3years out of date that did not switch already.

PR #2817, opened 2026-05-20
SubtitleHelper: replace usage of java.lang.Character

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/SubtitleHelper.kt
```diff
@@ -315,10 +315,17 @@ object SubtitleHelper {
         val flagOffset = 0x1F1E6  // regional indicator "[A]"
         val offset = flagOffset - asciiOffset
 
-        val firstChar: Int = Character.codePointAt(countryLetters, 0) + offset
-        val secondChar: Int = Character.codePointAt(countryLetters, 1) + offset
+        fun toSurrogatePair(codePoint: Int): String {
+            val high = ((codePoint - 0x10000) shr 10) + 0xD800
```
fire-light43, line 325: Just a short comment explaining where these magic values originate from, or what this does would go a long way

Luna712: Comment added

fire-light43 (approved):

PR #2819, opened 2026-05-20
Replace usage of String.format in library

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Gofile.kt
```diff
@@ -68,9 +69,11 @@ open class Gofile : ExtractorApi() {
     }
 
     private fun formatBytes(bytes: Long): String {
```
fire-light43, line 78: This does not have an identical return value:

https://play.kotlinlang.org/embed?short=X4ZTRxKOB

fire-light43 (dismissed):
Just one small change, I want to avoid behavioral changes in the migration

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Gofile.kt
Luna712: Will look in the next couple days thanks for this!

Luna712: This should now be fixed.

fire-light42 (approved):

PR #2820, opened 2026-05-21
Remove usage of junit within the app itself

Luna712:
When I was testing compose, I realized that this causes issues where junit was wiring up instrumentation within the app, which overrode and conflicted with compose resource context. We don't need it as a dependency for just TestingUtils, so this refactors it to just use AssertionError directly.


fire-light43 (approved):
Thank you for the contribution

PR #2828, opened 2026-05-23
HlsPlaylistParser: replace usage of Java APIs with kotlin equivalents

Luna712:
This does add kotlinx-io as a dependency, it is more light weight then okio, and when we migrate to ktor for example it will become an implicit dependency anyway so it should be fine to just add. Allows for more reliable and readable usage then doing some of this manually.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/HlsPlaylistParser.kt
```diff
@@ -534,29 +535,29 @@ object HlsPlaylistParser {
 
     object C {
```
fire-light42: This object ought to be private to prevent accidental imports when trying to import from ExtractorApi.

fire-light42 (approved):
Looks good and works well from testing

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/HlsPlaylistParser.kt
Luna712: Done now. That maybe could've been a seperate PR however I figured it was a small enough change to just do here.

fire-light42 (approved):

PR #2829, opened 2026-05-23
Setup compose multiplatform and migrate first screen to it

Luna712:
I explained most in code comments the best I could as well. Icons are automatically generated from https://fonts.google.com/icons which provides automatically generated code for Compose as well. In the future most icons will use this if a relevant one is found, otherwise we can use SVG icons. Compose multiplatform also supports android XML versions in a KMP context, but I actually found this to be very inconsistent and we shouldn't use that directly if we can avoid it.

I am also going to go through a lot of other changes here and some reasoning as well:

This adds reusable components that will be useful elsewhere as well:
* `.cloudStreamRipple`: replacement for `selectableItemBackground`
* `.tvFocusable`: handle manual focus for TV to ensure it is correctly following it and handles the border outline for it also, will probably need improved/changed more later though as we need different versions for it in some places
* ProfilePicture: to reuse in all the different profile picture containers, accepts parameters where things will need to be different later
* SettingsItem: Used for all settings category headers for now, will later also be used in all settings subpages for consistency and avoid code duplication. This may need some changes later though also

This handles Monet/Material You differently. This is a dynamic color scheme only available on Android, therefore to define it it uses expect/actual declarations. Other platforms do have their own concept of dynamic themes though, so in the commonMain API it  just uses the `DYNAMIC` terminology for better support for these later on and a more consistent cross platform API. This applies to both the theme and primary colors for it.

Since things are not XML based (and when we do use XML based stuff like strings in composeResources, there is still no `@string/` usage), this also adds new `PreferenceKeys` and `PreferenceDefaults` objects with constants for preferences. `PreferenceKeys` will replace all the string keys values in the current `donottranslate-strings.xml` and `PreferenceDefaults` will just include the default values for the preferences. Both are there to ensure consistency (not different defaults in different places) and to avoid accidental typos or something. For now it keeps what is added there only to what we actually use in `composeApp` as more are needed there these will be populated more as well.

For now, this also duplicates the default profile background images to `composeResources` so we can use them in `:composeApp` directly as well.

This also avoids refactoring to much of the app intentionally like removing/moving the companion object from the old SettingsFragment to avoid this PR growing massively. This adds a KMP version for device layout detection though as I needed to get if its a TV in `:composeApp`. 

Eventually we will have to split `:app` into `:androidApp` which just includes the signing stuff and entry points (AndroidManifest.xml, MainActivity (but a much slimmer version) and CloudStreamApp and other entry points like services (things defined in AndroidManifest).

AGP 9 prevents the ability to have both `com.android.application`  and `com.android.kotlin.multiplatform.library` at the same time, which means android entry points/APK generation can't be in a KMP module. `:app` will slowly be drained into fully KMP APIs into `:composeApp`. Stuff in there should not be stuff extensions will typically need though, otherwise it should go in `:library` instead.

Also, it is not immediately necessarily, but eventually we should also add the `composeResources` directory to weblate. This also sets up the locales python script to support that directory if and/or when we add it.

Happy to make any changes necessary though.


fire-light42 (commented):
Please try to add `@Preview` annotations to components and screens. To make compose accessible people need to see the layouts.

Luna712:
> Please try to add `@Preview` annotations to components and screens. To make compose accessible people need to see the layouts.

That was a good idea and is now done. Thank you!

composeApp/src/commonMain/kotlin/com/lagradost/cloudstream4/utils/DeviceInfo.kt
```diff
@@ -0,0 +1,11 @@
+package com.lagradost.cloudstream4.utils
+
+internal expect object DeviceInfo {
+    fun getDeviceType(): DeviceType
+    fun isLandscape(): Boolean
+    fun getLayoutPreference(): Int
+}
+
+enum class DeviceType {
```
fire-light42: While having a enum for this sort of thing is good, it will make it a pain with imports. E.g. if you write EMULATOR is it a DeviceType or DeviceLayout? getDeviceType or the enum will never be used in any real code given that we already have the isLayout function that checks an entire set. 

```kotlin
when(DeviceInfo.getDeviceType()) {
PHONE, TV -> {...}
EMULATOR,COMPUTER -> {...}
}
```
Is more typesafe, but in all our code we use 

```kotlin
if (isLayout(PHONE or TV) {...} else {...}
```
It is both shorter and more readable. 

So I suggest we simply remove this enum `DeviceType` to save anyone the confusion on why `isLayout(PHONE)` is giving an error because the wrong type was used.

fire-light42 (changes requested):

composeApp/src/commonMain/kotlin/com/lagradost/cloudstream4/utils/DeviceInfo.kt
Luna712: I am little confused. The reason it wss like that is so that the expect/actual can return real types for KMP and future platforms we may want to add. The usage would still be isLayout(PHONE OR TV) just using DeviceLayout imports. Once that is normalized in other parts of the app, I dont see where confusion would come from? Also DeviceInfo is internal, so wouldnt be a public API beyond composeApp (could mark with `@InternalAPI` also). But the expect/actual needs a type safe option to return also. Thats why the enum exists for DeviceType.

fire-light42: Ok, let me just explain it for a developer experience, I do not want this happening: 

Image

There should only be one import with this name, especially when both look nearly identical in the import path. It is really easy to import the wrong one with the current setup. The DeviceType enum should not be exposed to compose developers.

Luna712: Oh wait I did forget to mark this enum Internal it seems. But maybe I can just move the enum to the same class and make it private so that it doesn't work outside where we need it for at all?

Luna712: Apologies, I just now understood what you meant. I will remove the enum and just let the expect/actual use Int

Luna712: I created a method that combines both ways using a type safe method without ambiguous imports using a value class.
IE like
```kt
    @JvmInline // This still works but has no affect on non-JVM targets
    value class Layout(val value: Int) {
        infix fun or(other: Layout) = Layout(value or other.value)
    }

    val PHONE = Layout(0b00001)
    val TV = Layout(0b00010)
    val EMULATOR = Layout(0b00100)
```

Then can be typed as `DeviceLayout.Layout` and still be used as `isLayout(PHONE or TV)` with zero ambiguity.

I will probably push this in a bit but could revert if you don't like this approach.

fire-light42: Yes, that is one solution but makes it a bit worse due to no "const" on value classes, and forces the user to also import "or". However it is a goodish solution for this case, but I would also make the "value" private as it does not make sense to expose it.

Luna712: > Yes, that is one solution but makes it a bit worse due to no "const" on value classes, and forces the user to also import "or". However it is a goodish solution for this case, but I would also make the "value" private as it does not make sense to expose it.

"or" doesnt need imported it seems? Also it can't be private because it needs to use for types elsewhere also unless I am misunderstanding something? I could make it internal though.

**UPDATE**: Yes I misunderstood, will make value private. I have to add another internal and though to do it so it can check layoutid also.

Luna712: I pushed what I had. I am open to changing anything you feel necessary with this implementation though. Hopefully I did what you meant by make value private not 100% certain.

kazemcodes:
Thanks for your efforts, just a question Google material contains the vector images, is it not possible to use them instead of using pure vector designin in code?

Luna712:
> Thanks for your efforts, just a question Google material contains the vector images, is it not possible to use them instead of using pure vector designin in code?

Google material extended that includes the actual icons is deprecated and no longer maintained, and increases the app size by about 20MB as it loads everything. We don't need a library with thousands of icons we will never use.

kazemcodes:
​Based on my experience migrating several apps to Compose multiplatform, here is how to save yourself time and headaches:
​1. Wrap Your Core Components
​Always create wrapper functions for common Compose code.
The framework evolves quickly, and APIs can change. If you wrap your components in shared code, a future update means you only have to fix the code in one place, instead of rewriting it across the entire app.
​2. Use SQLDelight for Databases
​I highly recommend SQLDelight. It is efficient, reliable, and built perfectly for multiplatform development. (However, feel free to use whatever database fits your project best!)
​3. Choose Koin for Dependency Injection
​For DI, use Koin. It is lightweight, easy to set up, and works seamlessly with Kotlin Multiplatform to keep your architecture clean

Luna712:
2 and 3 are planned. Koin is still under consideration. If it will end up helpful it will be used. SQLDelight I have already started writing a lib and migration code for SharedPreferences for it.

kazemcodes:
> > Thanks for your efforts, just a question Google material contains the vector images, is it not possible to use them instead of using pure vector designin in code?
> 
> Google material extended that includes the actual icons is deprecated and no longer maintained, and increases the app size by about 20MB as it loads everything. We don't need a library with thousands of icons we will never use.

Thanks I didn't know that it increases the size that much but R8 will remove unnecessary items from code and resources

fire-light42 (changes requested):
I think we can make the layout system a bit more compose friendly. 

By having a mutable state, we do not need to restart the entire activity. 
Instead we simply have to call the update function and every "by isLayoutState" will cause a recomposition.

```kotlin
private val layoutState = mutableStateOf(layoutId)
fun isLayoutState(layoutFlags : Layout) = derivedStateOf { layoutState.value.and(layoutFlags) }
fun update() {
	layoutId = resolveLayout()
	layoutState.value = layoutId
}
```

```kotlin
val isTV by isLayoutState(TV)
```

This is easier to read, and reason about. 

I also have some nitpicks that I want to clear up early. 

1. More things can be marked as `@Immutable`, this includes "Layout", "SettingsProfileState" and "SettingsVersionState"

2. No need to specify the structuralEqualityPolicy, it is the default

3. I would rather you use LocalContext.current and pass it as a variable than use globals

```kotlin
val version = buildVersionState(LocalContext.current)
val profile = buildProfileState(LocalContext.current)
```

Where the functions use `context.currentCommitHash()` and `DataStoreHelper.getDefaultAccount(context)`

This is because we want to avoid all requireContext/requireActivity, use inflater.context for ComposeView and LocalContext.current for `@Composable`.

fire-light42:
Now after we have released a new stable, I think this should be merged as soon as possible!

Luna712:
I will work on what you mentioned later today and push in the next few days. I am a little confused about some of what you said though I will work on it and might ask here for some clarity in a bit if I can't figure out what you mean.

Luna712:
I pushed some of the fixes now. Will do the rest later.

Also just to note, I have tests written for compose (both UI and logic tests) but I will push that in a seperate PR. I also have detekt setup to ensure code standards while things are migrated that I will push in a seperate PR. It would ensure things like that bracket newline if-block thing you wanted in my other PRs and other standards. I am not sure what all you will want but I can push it later, and it can be reviewed. I mention that now because some of what I also just pushed was found from that.

Luna712:
@fire-light42 I *believe* I did everything you were wanting. Hopefully correctly. If not I apologize and can adjust.

LagradOst (changes requested):
Using this PR I started porting QN to compose. And I found out that 

1. Missing AmoledLight, Amoled is not the same as AmoledLight https://github.com/recloudstream/cloudstream/blob/505722c1b5c1a79d8295e984e55d3c72457b2e44/app/src/main/res/values/styles.xml#L87
2. MaterialTheme typography? I think we use google_sans.xml for all fonts

Luna712:
> Using this PR I started porting QN to compose. And I found out that
> 
> 1. Missing AmoledLight, Amoled is not the same as AmoledLight https://github.com/recloudstream/cloudstream/blob/505722c1b5c1a79d8295e984e55d3c72457b2e44/app/src/main/res/values/styles.xml#L87
> 2. MaterialTheme typography? I think we use google_sans.xml for all fonts

@LagradOst 

1. Whoops nice catch. That was originally intentional while I first set it up before splitting that but I forgot to finish it.
2. I will take a look again. I was aiming to use material stuff mostly to ensure more consistent cross platform but I will look into that.

LagradOst:
> > Using this PR I started porting QN to compose. And I found out that
> > 
> > 1. Missing AmoledLight, Amoled is not the same as AmoledLight https://github.com/recloudstream/cloudstream/blob/505722c1b5c1a79d8295e984e55d3c72457b2e44/app/src/main/res/values/styles.xml#L87
> > 2. MaterialTheme typography? I think we use google_sans.xml for all fonts
> 
> @LagradOst
> 
>     1. Whoops nice catch. That was originally intentional while I first set it up before splitting that but I forgot to finish it.
> 
>     2. I will take a look again. I was aiming to use material stuff mostly to ensure more consistent cross platform but I will look into that.

Please also include comments directly on the colors, as it makes it a lot easier to convert. 

```kotlin
    /** primaryBlackBackground */
    var background by mutableStateOf(background)
    /** primaryGrayBackground */
    var surfaceVariant by mutableStateOf(surfaceVariant)
    /** iconGrayBackground */
    var surface by mutableStateOf(surface)
    /** boxItemBackground */
    var surfaceContainer by mutableStateOf(surfaceContainer)
    /** textColor */
    var onBackground by mutableStateOf(onBackground)
    /** grayTextColor */
    var onSurfaceVariant by mutableStateOf(onSurfaceVariant)
    /** iconColor */
    var icon by mutableStateOf(icon)
    /** colorPrimary */
    var primary by mutableStateOf(primary)
    /** colorOngoing */
    var ongoing by mutableStateOf(ongoing)
```

LagradOst:
Also, is there a good reason to use https://fonts.google.com/icons imagevectors directly? I found `painterResource(R.drawable....)` to be very easy to use without bloating the code directory. It should be the same as `painterResource(Res.drawable....)` on KMP.

Luna712:
> Also, is there a good reason to use https://fonts.google.com/icons imagevectors directly? I found `painterResource(R.drawable....)` to be very easy to use without bloating the code directory. It should be the same as `painterResource(Res.drawable....)` on KMP.

This was a personal preference, because it was easier for me to build and maintain it in pure Kotlin where possible, faster build by reducing the stuff in composeResources a bit, and while mostly negligible reduced the app size just a tiny bit. I did first use drawables before I realized fonts.google.com gave the Kotlin version also. Not 100% against changing if you really want though.

Luna712:
> MaterialTheme typography? I think we use google_sans.xml for all fonts

I have a finished version that readds this but the more I think about it the more I don't really like it. By default MaterialTheme uses `FontFamily.Default` which, at least on Android, becomes the actual system default font, selected via own system settings and typically expected throughout. I kinda would rather us keep that rather than our custom app font. Although that is just my opinion.

The way I have it written would make it easy to eventually add a setting to use system default font, fonts from composeResources, or a user custom selected font via local file path (though it only adds the initial API and doesn't set anything up there, was just to make it easier later on). Bur the more I was thinking of it, I don't think this was the right approach. I think what we should do is system default font, or a user selected font (eventually). Again, just my opinion.

I was hoping for opinions on this before I push anything else with it. My main goals with this PR is to make it easily expandable, readable, and maintainable in the future so I just want to get this right the first time. At the very least I do thing we should in some way support using the system default font rather than the app font itself as it looks more universal across other aspects of the system that way also.

LagradOst:
> > MaterialTheme typography? I think we use google_sans.xml for all fonts
> 
> I have a finished version that readds this but the more I think about it the more I don't really like it. By default MaterialTheme uses `FontFamily.Default` which, at least on Android, becomes the actual system default font, selected via own system settings and typically expected throughout. I kinda would rather us keep that rather than our custom app font. Although that is just my opinion.
> 
> The way I have it written would make it easy to eventually add a setting to use system default font, fonts from composeResources, or a user custom selected font via local file path (though it only adds the initial API and doesn't set anything up there, was just to make it easier later on). Bur the more I was thinking of it, I don't think this was the right approach. I think what we should do is system default font, or a user selected font (eventually). Again, just my opinion.
> 
> I was hoping for opinions on this before I push anything else with it. My main goals with this PR is to make it easily expandable, readable, and maintainable in the future so I just want to get this right the first time. At the very least I do thing we should in some way support using the system default font rather than the app font itself as it looks more universal across other aspects of the system that way also.

The problem is that our "custom app font" is used everywhere in the app, and the migration will take longer than 1 day. Therefore we will have a lot of time in-between the full port, where half the app uses another font compared to the other. I could agree that FontFamily.Default might be good in the end, but while porting I need the font to match the rest of the app exactly. We can make it customizable later, but now we have to hardcode it as otherwise it will look very weird.

Luna712:
I know I may just be being stubborn right now and I apologize and I am not 100% against what you said, I just don't think it is necessary, what I would rather is make the existing app also use the system font by default. This could be done by simply removing the custom app font, as android will use the system font by default as well. Again if you really want me to hard code it I will but the point of this is that composeApp won't need so many changes later on, only a few temp-only stuff, not much at all... so that as things are migrated it will be easy to use. If I do it manually that is a much larger change because the way I did it use MaterialTheme typography that controls the size as well, as I would have to override for each typography style the font only, upload the all the ttfs to composeResources, all for a temporary solution that would eventually be changed again, and that is kinda what I wanted to avoid.

LagradOst:
> I know I may just be being stubborn right now and I apologize and I am not 100% against what you said, I just don't think it is necessary, what I would rather is make the existing app also use the system font by default. This could be done by simply removing the custom app font, as android will use the system font by default as well. Again if you really want me to hard code it I will but the point of this is that composeApp won't need so many changes later on, only a few temp-only stuff, not much at all... so that as things are migrated it will be easy to use. If I do it manually that is a much larger change because the way I did it use MaterialTheme typography that controls the size as well, as I would have to override for each typography style the font only, upload the all the ttfs to composeResources, all for a temporary solution that would eventually be changed again, and that is kinda what I wanted to avoid.

I think we need the built in font. I choose it because it fits the app. and changing it randomly like this to system is a no-go for me right now. And no, adding a font is really easy, check out https://github.com/LagradOst/QuickNovel/blob/master/app/src/main/java/com/lagradost/quicknovel/compose/theme.kt#L464-L496

Luna712:
I had a very similar version though not exact that works but I felt it was more boilerplate than I want. I guess I can just later push what I have and can get a review if you want it like that or not.

LagradOst:
Closed as of #3151 
This PR really laid the groundwork for compose even if it was never merged, so I really have to commend you for the work on this pull request.

Luna712:
I just want to say that, now that I’ve closed all my PRs, I will no longer be contributing to this repo whatsoever.

I do not appreciate putting in all the work I did for the Compose migration and everything else I have done, only for this PR to just be closed like that and the work I put into it to then get lumped into other diffs. I spent a significant amount of time working on those changes and putting everything together, and it is extremely frustrating to see that work effectively get folded into other changes rather than being handled as the contribution I actually made.

When I contribute to a project, I expect the time and effort I put into the work to actually be respected. Having a substantial amount of work that I personally spent certainly a lot of time implementing, testing, and preparing ultimately get absorbed into other diffs makes me feel like the work I put into this was simply disrespected.

Because of this, I no longer have any interest in continuing to contribute to this repo. I’ve closed my PRs, and I will not be contributing here again.

LagradOst:
This is a severe miscommunication! 

I greatly respect you, your work and the time you put into every single pull request, but this appreciation is difficult to get across with text. 

This pull request is so foundational and important that we could not keep waiting for it to get merged. For a multitude of reasons. It showed that Compose was possible, and such a change would enable a lot of opportunities. In fact, it made me so excited I ported the entire quicknovel app to compose based on this PR, and learned a lot while doing it. 

Since then, we have gotten multiple comprehensive rewrites and forks aiming to replace or create new compose UIs, which I do not want to encourage when you have already made a much better foundation. This influx of (AI-assisted) compose rewrites really put pressure on getting an initial compose version up and running. Since I absolutely do not want to rewrite the entire codebase from a subpar starting point or block their pull requests.   

That is why (besides my excitement) I asked you a month ago if I could take over the pull request to get it merged as soon as possible. Otherwise even more people would get started on compose too early.  With no answer and an indeterminate time until this pull request was ready, I needed to make a decision in order to allow other people to also start working on Compose. 

I closed this pull request because I respect your time, and did not want you to do any double work. At the time it did not cross my mind that this would be disrespectful. I am very sorry. It was not at all my intention to dismiss or diminish your contributions and hard work. 

I understand your decision, but I want to make it clear that I never meant you any ill.

Luna712:
I appreciate the response, though that doesn't change my opinion of the matter.

I was busy when you first messaged me so I didn't respond because I didn't care if you took *this* PR over to finish it, what I didn't like was that the entire work was practically erased by opening a new PR, with an entirely different structure which made everything else I had already worked on and had already ready to go PRs for the future completely irrelevant and huge waste of time now.

I had dozens of UI and logic unit tests, standardized checks to ensure standards are maintained of this repository has been trying to attain via detekt rules, plus an entire KMP DataStore that allowed the backup sync stuff better I all had done, but they relied on the fact that the structure in this PR was used, and by changing that and erasing all my own contribution here, that ended up making all the work I had planned or in progress for this undermined, and is why I no longer want to contribute to a project where it doesn't seem my contributions are taken seriously.

PR #2833, opened 2026-05-24
Add KMP locale API for SubtitleHelper

Luna712:


library/api/jvm/library.api
```diff
@@ -7255,8 +7255,9 @@ public final class com/lagradost/cloudstream3/utils/SubtitleHelper$LanguageMetad
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class com/lagradost/cloudstream3/utils/SubtitleHelperKt {
+public final class com/lagradost/cloudstream3/utils/SubtitleHelperPlatform_jvmKt {
```
fire-light42, line 7258: A breaking change, but I could thankfully not find existing plugins using `SubtitleHelper.getCurrentLocale()`

fire-light42 (approved):

PR #2836, opened 2026-05-24
Coroutines KMP API

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/mvvm/ArchComponentExt.kt
```diff
@@ -129,16 +124,18 @@ suspend fun <T> suspendSafeApiCall(apiCall: suspend () -> T): T? {
...
-        separator = "\n"
-    ) {
-        "${it.fileName} ${it.lineNumber}"
-    }
+    val prefix = if (showMessage) this.message?.let { "\n$it" } ?: "" else ""
+    return prefix + this.stackTraceToString()
+        .lines()
+        .mapNotNull { line ->
+            val trimmed = line.trim()
+            if (trimmed.startsWith("at ")) trimmed.removePrefix("at ") else null
```
fire-light42: This shows the full package names as opposed to just the file names. It does not look good in the provider test to have full package names, can it be solved?

fire-light42: If not it's fine, expect actual is overkill for this. I could not find a good solution from my cursory check.

Luna712: I have a way to fix this. I wrote a test for it also. I am just testing it on a JS compiled version also before pushing it.

Luna712: Appears to work on JS also. I will push it and you can tell me what you think also, we can revert it back if you don't like this way.

fire-light42 (approved):
Good solution to the pretty stack trace!

PR #2845, opened 2026-05-26
Replace EnumSet for dubStatus

Luna712:
Extensions always use addDubStatus and don't use the constructor directly, only the builder with the helper method, so this should still maintain backwards compatibility.

Needs #2815 first


fire-light42:
I looked around, and it looks like the only actively maintained repository that is using this is https://github.com/lawlietbr/lietrepo/tree/main so it might be worth it to make an issue before this merge.

Luna712:
> I looked around, and it looks like the only actively maintained repository that is using this is https://github.com/lawlietbr/lietrepo/tree/main so it might be worth it to make an issue before this merge.

issue created on that repo. Thank you! I missed that one.

PR #2852, opened 2026-05-29
Fix T::class in parseJson causing type erasure in some cases

Luna712:


Luna712:
Was trying to get this update to test it but it wasn't building in my fork or locally so opened here so I can test this for a moment.

**UPDATE**: Tested with both kotlinx serialization and Jackson and it worked now

app/src/main/java/com/lagradost/cloudstream3/utils/InAppUpdater.kt
```diff
@@ -93,19 +93,17 @@ object InAppUpdater {
     private suspend fun Activity.getReleaseUpdate(): Update {
         val url = "https://api.github.com/repos/$GITHUB_USER_NAME/$GITHUB_REPO/releases"
         val headers = mapOf("Accept" to "application/vnd.github.v3+json")
-        val response = parseJson<Array<GithubRelease>>(
```
fire-light42, line 96: To be honest I would rather keep these as Arrays to prevent any potential type erasure issues in the future.

Luna712: Absolutely fair actually. Will change back now.

Luna712: And done.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -63,19 +68,26 @@ object AppUtils {
     }
 
     @InternalAPI
```
fire-light42, line 70: I would prefer to split the logic like this to make it more readable (GitHub suggestions did not work):
```kt
    @InternalAPI
    fun <T : Any> parseJson(
        value: String,
        kClass: KClass<T>,
        serializer: KSerializer<T>? = null,
        typeReference: TypeReference<T>? = null,
    ): T {
        // @Serializable generates a serializer at compile time; contextual serializers are
        // registered manually in serializersModule, we need both to support all cases
        val s =
            serializer ?: kClass.serializerOrNull() ?: json.serializersModule.getContextual(kClass)

        // Prefer Kotlin Serialization over Jackson
        if (s != null) {
            try {
                return json.decodeFromString(s, value)
            } catch (e: SerializationException) {
                logError(e)
            }
        }

        return if (typeReference != null) {
            mapper.readValue(value, typeReference)
        } else {
            mapper.readValue(value, kClass.java)
        }
    }
```

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -21,7 +23,10 @@ object AppUtils {
     }
 
     inline fun <reified T : Any> parseJson(value: String): T {
-        return parseJson(value, T::class)
+        // serializer<T>() preserves full generic type info (e.g. List<DataClass>)
+        // and must be resolved here while T is still reified, same for TypeReference
+        val serializer = try { serializer<T>() } catch (_: SerializationException) { null }
```
fire-light42: This can also throw an IllegalArgumentException. Perhaps it's better to just swallow all exceptions here.

fire-light42 (commented):
Looks good and seems to work well, just some minor nitpicks

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
Luna712: Thanks done.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
Luna712: Changed to Exception. We can change to be more specific maybe when we rework this again when removing Jackson, but all Exceptions are fine for now yeah, and may just leave forever, the reason I want to do minimum is that's why we have `parseJson` and `tryParseJson` as separate. `parseJson` is not supposed to swallow as many exceptions. That's the only reason why.

fire-light42: I do agree on the `parseJson/tryParseJson` point, but I do not want to exit early if kotlin serialization fails when Jackson may succeed. When we remove Jackson we should fail here :+1:

Luna712: Yep good point.

fire-light42 (approved):

PR #2861, opened 2026-05-31
Add TRAKT_CLIENT_ID to Archive build as well

Luna712:


fire-light42:
You are 20s faster than me at https://github.com/recloudstream/cloudstream/tree/fire-light42-patch-2 😆

PR #2863, opened 2026-06-01
Add a new jvmCommonMain source set to library

Luna712:


Luna712:
@fire-light42 could this be merged so we can deduplicate the extractor so if there are changes it doesn't need to be done twice? No worries if just waiting though, but just thought it would be best to deduplicate the extractor.

fire-light42:
Yes, I will merge this as soon as I have time to properly test.

PR #2868, opened 2026-06-03
Update Kotlin to 2.4.0

Luna712:
* Uuid is now stable, so remove opt-in
* `annotation-default-target=param-property` is now default, and compiler warns about it being redundant, so remove
* Compiler now warns about a redundant expression in HomeFragment, so just remove redundant `true`; has no other affect


Luna712:
@fire-light42 this could be merged also to make extensions not to have to do any opt-in magic for UUIDs when we deprecate the old Java UUID back-compat after next stable also. Of course they will also need to update to Kotlin 2.4 to avoid needing opt-in, but this would also set that up. Would also more guarantee stable library binary compat for Uuid though.

fire-light42:
Looks good, I will try to merge soon. It just needs some testing before I can merge it :+1:

Luna712:
@fire-light42 could this be merged before next stable please? If not I guess that is fine but would be a better more stable API that way.

fire-light42 (approved):
Works well when testing backwards compatibility.

I missed merging this before stable :(

PR #2870, opened 2026-06-03
MyDramaList: migrate to kotlinx serialization

Luna712:
**Note**: MyDramaList has other preexisting issues that need fixed to make it fully work.


fire-light42 (commented):
I am unsure if this might cause problems with extensions targeting the "new" version breaking things for the stable. Do you think this should be merged before or after the next stable?

Luna712:
> I am unsure if this might cause problems with extensions targeting the "new" version breaking things for the stable. Do you think this should be merged before or after the next stable?

@fire-light42 

This one in particular doesn't matter and can be merged whenever as MDL doesn't work right now. Hopefully we can fix it soon though. The others I think should be merged just before next stable especially #2874 so that extensions can use those as serializable classes.

#2800 should also probably be merged just before next stable if you're happy with it after testing etc... as it does the only other metaprovider not in my very recent PRs to kotlinx serialization also. Then only extractors will remain in library anyway.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
```diff
@@ -77,7 +80,7 @@ abstract class MyDramaListAPI : MainAPI() {
         val list = app.get(
             url = "${request.data}&limit=20&page=$page&lang=en-US",
             interceptor = headerInterceptor
-        ).parsed<SearchResult>().map { element ->
+        ).parsed<List<MediaSummary>>().map { element ->
```
fire-light42: Please use arraylist as before

fire-light42 (changes requested):
Please use ArrayList as before, to prevent problems with jackson.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/metaproviders/MyDramaList.kt
Luna712: Done.

fire-light42 (approved):

PR #2872, opened 2026-06-03
TraktProvider: migrate to kotlinx serialization

Luna712:


Luna712:
I have now also fixed a circular reference with this. Jackson would handle it, but it doesn't work the same in kotlinx. Running the test caught that.

fire-light42 (approved):

PR #2874, opened 2026-06-03
Make some classes serializable

Luna712:
Replaces JsonProperty with SerialName only if it does not match the variable name, because in the case it would be the default and just makes it easier to migrate later on by removing what we can now.


fire-light42 (approved):

Luna712:
@fire-light42 I caught some issues when further testing the ControllerActivity change. This now does some changes to Jackson as well  That shouldn't cause an issue and makes the output fully match. We shouldn't need to serialize those methods like that. And also fixed some things that were being serialized in kotlinx but not Jackson. I can change this if necessary though.

PR #2877, opened 2026-06-04
InAppUpdater: migrate to kotlinx serialization

Luna712:
This one shouldn't need to wait for next stable like the library ones might. I also tested this when fixing the updater the first time to ensure my fix worked with both versions. So this should still work unless something changed since then.


fire-light42 (approved):
I also tested and it worked well 👍

PR #2878, opened 2026-06-04
CloudStreamPackage: migrate to kotlinx serialization

Luna712:


app/src/main/java/com/lagradost/cloudstream3/actions/temp/CloudStreamPackage.kt
```diff
@@ -49,19 +52,16 @@ class CloudStreamPackage : OpenInAppAction(
         const val DURATION_EXTRA: String = "dur" // Duration time in MS (Long)
     }
 
+    @Serializable
     data class MinimalVideoLink(
-        @JsonProperty("uri")
+        @JsonProperty("uri") @SerialName("uri")
+        @Serializable(with = UriSerializer::class)
         val uri: Uri?,
```
fire-light42, line 61: I think we've run into the first null issue from https://github.com/recloudstream/cloudstream/pull/2897

```
Java.lang.AssertionError: Serialization mismatch for:
com.lagradost.cloudstream3.actions.temp.CloudStreamPackage.MinimalVideoLink

Jackson:
{"uri":null,"url":"YRZOHJ","mimeType":"QSIUJMXEP","name":"NHOIYX","headers":{"UDHQNR":"CZDV","CQTCZKYKWN":"KGRYBRD","DXAOWNP":"XDQXGLZAOC"},"quality":1779}

Kotlinx:
{"url":"YRZOHJ","mimeType":"QSIUJMXEP","name":"NHOIYX","headers":{"UDHQNR":"CZDV","CQTCZKYKWN":"KGRYBRD","DXAOWNP":"XDQXGLZAOC"},"quality":1779}
``` 
However, this can be sidestepped by making Instancio create a valid Uri-type (we should really have this for all troublesome types):
```kt
val instance = Instancio.of(kClass.java).lenient().supply(all(Uri::class.java), Supplier {
    Uri.parse("https://example.com")
}).create()
```
But this gives 
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Direct self-reference leading to cycle (through reference chain: com.lagradost.cloudstream3.actions.temp.CloudStreamPackage$MinimalVideoLink["uri"]->android.net.Uri$StringUri["canonicalUri"])
```
Meaning that Jackson actually cannot serialize it properly 😬 
Perhaps we also need custom JsonDeserializer for jackson? What do you think is best?

Luna712: Tbh I think the best option here is to add a temporary annotation that can skip certain classes in tests and drop worrying about Jackson for things like this. I feel focusing on fixing Jackson to match kotlinx exactly is a waste of time for things like this when Jackson is being phased out. But it is up to you.

fire-light42: I do agree that fixing Jackson is not something we should waste time on. An annotation to skip the test is probably the best solution.

Luna712: #2947

Luna712: Added the annotation to skip it.

PR #2879, opened 2026-06-04
SearchSuggestionApi: migrate to kotlinx serialization

Luna712:


fire-light42 (approved):
Not sure about the suspend, but it is private and works.

Luna712:
> Not sure about the suspend, but it is private and works.

Whoops sorry, that was an unrelated change that will later be necessary caused by me splitting a bunch of different changes on a single branch to different PRs.

fire-light42:
I suspected that was the case and kept it, no problem :+1:

PR #2881, opened 2026-06-04
videoskip: migrate to kotlinx serialization

Luna712:


fire-light42 (approved):
There were some issues in testing with Instancio recursion depth, but I fixed them 👍

PR #2887, opened 2026-06-05
Bump media3 & nextlibmedia3

KingLucius:
- Bump `media3` to `1.10.0` to keep the full compatibility with `nextlibMedia3`
- Bump `nextlibMedia3` to `1.10.0-0.12.1`

I have tried many media files with different formats & DRM


Luna712:
Just an FYI we reverted this before because it breaks tracks.

KingLucius:
> Just an FYI we reverted this before because it breaks tracks.

After check, you are right. It seems Media3 changed behavior of how/when tracks are loaded, I will investigate it

PR #2888, opened 2026-06-05
AuthAPI: migrate to kotlinx serialization

Luna712:


app/src/main/java/com/lagradost/cloudstream3/syncproviders/AuthAPI.kt
```diff
@@ -51,20 +56,25 @@ data class AuthToken(
...
     val id: Int,
     /** Profile picture URL */
-    @JsonProperty("profilePicture")
+    @JsonProperty("profilePicture") @SerialName("profilePicture")
     val profilePicture: String? = null,
     /** Profile picture Headers of the URL */
-    @JsonProperty("profilePictureHeader")
-    val profilePictureHeaders: Map<String, String>? = null
+    @JsonProperty("profilePictureHeaders") @JsonAlias("profilePictureHeader")
+    @SerialName("profilePictureHeaders") @JsonNames("profilePictureHeader")
```
fire-light42, line 76: Why JsonNames and JsonAlias? Is `@SerialName("profilePictureHeader")` not sufficent?

Luna712: Because it seemed like a typo and this was meant to keep back compat. using profilePictureHeaders instead of header, which was more correct here. Not absolutely necessary but it just bothered me and wanted to use this to fix it.

fire-light42: Good thinking

PR #2890, opened 2026-06-05
Subdl: migrate to kotlinx serialization

Luna712:


app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/Subdl.kt
```diff
@@ -122,69 +123,77 @@ class SubDlApi : SubtitleAPI() {
         }
     }
 
+    @Serializable
     data class SubtitleOAuthEntity(
-        @JsonProperty("userEmail") var userEmail: String,
```
fire-light42, line 126: Please keep all `@JsonProperty` to ensure everything still works as expected with Jackson. Keeping `@JsonProperty` also makes testing easier.

Luna712: Yes I am working through my open PRs to do this already, although I was only adding them back where they are different then the variable name since that is the default anyway. In future PRs was planning to keep all though.

Luna712: This one is done now.

fire-light42: No, I want to keep *all* `JsonProperty` until we fully migrate.

Removing `JsonProperty` before we fully migrate will cause unforeseen issues. 

Removing it assumes identical names will retain identical behavior. However, that assumption is not always true, the simplest example is private properties. 

```kt
data class TestClass(
    @SerialName("name") val name: String = "default"
)

data class TestClassPrivate(
    @SerialName("name") private val name: String = "default"
)

@Test
fun jsonPropertyTest() {
    val json = """{"name": "real value"}"""
    val normalClass = jacksonMapper.readValue<TestClass>(json)
    val privateClass = jacksonMapper.readValue<TestClassPrivate>(json)
    println("result $normalClass $privateClass")
}
```
This only affects few properties, but renaming properties before we remove Jackson can also lead to obscure bugs. We just don't know what other cases also fail when we remove `JsonProperty`. 

It is simply safer to keep all annotations while Jackson is still in use.

Luna712: Done for this one

Luna712: Now done for all my open PRs unless I missed something.

fire-light42 (approved):

PR #2891, opened 2026-06-05
SubSource: migrate to kotlinx serialization

Luna712:


fire-light42 (approved):
One step closer to KMP

PR #2896, opened 2026-06-06
Fix serialization testing

fire-light42:
The only other option I could find was moving this to the library JVM tests and keeping ClassGraph, but we lose out on app data class testing in that case.

@Luna712


Luna712:
Just a note this won't work for testing everything anyway, some things like Any in data classes parsed by Json must be changed to JsonElement so the tests will fail for things like that anyway. There is not really any way around it for that as far as I know either. Other than that this looks fine to me though.

fire-light42:
I tested with the MyDramaList `JsonElement` example and it worked :partying_face: 

If we find a class which is impossible to test we can always add an annotation to bypass the serialization tests.

Luna712:
Wow nice that really surprises me...

fire-light42:
I was also surprised, Android development is rarely convenient like that :smile:

PR #2897, opened 2026-06-06
Set explicitNulls = false for kotlinx serialization

Luna712:
This is more inline with Jackson's behavior otherwise for nullable types with non default values, and they don't exist, gives a Missingfield exception whereas it worked on Jackson. We may need coerceInputValues = true also but I am unsure of that right now.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -90,6 +90,7 @@ class ErrorLoadingException(message: String? = null) : Exception(message)
 @Prerelease
 val json = Json {
     encodeDefaults = true
+    explicitNulls = false
     ignoreUnknownKeys = true
 }
 


Luna712:
Actually this also prevents the encoding of explicit null values as well so it is a little different but it is the closest I could fine. I am actually unsure of this...

Luna712:
~~Actually docs are just confusing...~~
~~https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-builder/explicit-nulls.html~~

~~says null values are omitted but the example shows it encoding null values if they are explicit.~~ yeah I am just getting confused the example uses two separate Json instances. Perhaps we just want to force default values for nullable fields, unsure. Even then it would be a bit different behavior than Jackson.

Luna712:
@fire-light42 any thoughts here? (to note I was testing with a fully migrated everything and removed Jackson fallback to uncover a few issues with our current Json setup)

fire-light42:
We do aim to simulate exact Jackson behavior however possible, if this helps with that I see no issues. However, I would see these edge cases before I can make a proper decision.

Luna712:
For example in RepositoryManager SitePlugin no plugins can load on kotlinx serialization if things like description is not passed. I got Missingfield exception on description in that case.

fire-light42:
In that case we absolutely need `explicitNulls = false`, but `coerceInputValues` is not what we want. Turning nulls to default values is very unexpected behavior.

fire-light42:
I dislike that encoding also strips the null. It should always encode null values, but allow missing null values too.

Luna712:
As far as I can tell, it can't do both.

As for coerceInputValues from my understanding, that turns invalid values to default values in some cases like nulls on non nullable types or an invalid enum.

fire-light42:
Unfortunate, but the explicitNulls behavior will still work with Jackson because missing fields become null. It just prevents us from turning it on in the future.

Luna712:
Okay this should not be done. It corrupts all data. I just did some more tests and restoring from previous backups causes a lot of data to be corrupted or just plain missing

Luna712:
Nevermind, that was my fault. Ironically the cause that broke it was that this fixed it in way more places then I thought, it was catching back to Jackson but this made it actually work to use kotlinx serialization but I had it wrong in my full migration, after fixing that if fully works. So this is still needed.

Luna712:
@fire-light42 do you think I should also add `.serializationInclusion(JsonInclude.Include.NON_NULL)` to the mapper for Jackson to ensure the exact same behavior also?

fire-light42:
No I would not want to change established Jackson behavior, it needs to be a stable fallback for now.

PR #2898, opened 2026-06-06
AppUtils: handle generic collection types for toJsonLiteral

Luna712:
This is primarily for set key which calls toJsonLiteral, to avoid having to manually change every set key to determine proper serializers manually. This might have some edge cases where it doesn't fully work but under most circumstances it should.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
@@ -9,6 +9,10 @@ import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.InternalSerializationApi
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerializationException
+import kotlinx.serialization.builtins.ListSerializer
+import kotlinx.serialization.builtins.MapSerializer
+import kotlinx.serialization.builtins.SetSerializer
+import kotlinx.serialization.builtins.serializer
 import kotlinx.serialization.serializer
 import kotlinx.serialization.serializerOrNull
 import kotlin.reflect.KClass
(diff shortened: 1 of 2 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/AppUtils.kt
```diff
@@ -28,19 +32,56 @@ object AppUtils {
...
-                json.encodeToString(serializer as KSerializer<Any>, this)
+                return json.encodeToString(serializer as KSerializer<Any>, this)
             } catch (e: SerializationException) {
                 logError(e)
-                mapper.writeValueAsString(this)
+                return mapper.writeValueAsString(this)
+            }
+        }
+        // Handle generic collection/map types where type params are erased at runtime
+        // and no serializer can be found via reflection alone
```
fire-light42, line 45: Please write test cases for this type of behavior, it looks very easy to break things here.

Luna712: Not 100% sure what to write for tests here. I have another method for this to recursively convert things to a JsonElement, more reliable maybe but a least a tiny bit less performant. The issue is that for things like setKey, the type ends up getting erased, which means we have a few other options also, make setKey always use reified for serialized classes, or pass the type directly. I am not 100% sure on the best path here.

fire-light42: Well, what objects would be improperly serialized in the `toJson()` function without these helpers? Those are some very good test cases. With those in mind it's much easier to make a good judgement on how to handle this.

Luna712: Will look more into this later and provide more information on why once I test again so I can remember what one actually prompted this PR as I can't remember which one now...

PR #2914, opened 2026-06-13
ParCollections: remove deprecated methods

Luna712:
runBlocking is not available on some platforms, instead of handling that with a temporary expect/actual, we can just remove these. They have been deprecated for awhile anyway.


fire-light42 (approved):
Looks good, but we will hold off on merging this one until it is necessary.

PR #2916, opened 2026-06-14
fix: resolve error 2004/2001 on Android TV (IPv6, timeouts)

fgmitesh:
Android TV gets ExoPlayer errors 2004/2001 on sources that work fine on 
mobile with the same Wi-Fi. Three fixes:

1. DohProviders: add .includeIPv6(false) — TV firmware often has a broken 
   IPv6 stack. DoH returns both IPv4+IPv6, OkHttp tries IPv6 first and 
   silently fails. Mobile handles the fallback; TV doesn't.

2. RequestsHelper: set explicit 30s timeouts + retryOnConnectionFailure(true)
   — OkHttp's default 10s is too short for slower TV network stacks.

3. CS3IPlayer: increase Cronet timeout 15s → 30s for the same reason.


fire-light42:
Did you test this?

mehdigm4life (approved):

PR #2923, opened 2026-06-15
ui: Hide pinned NSFW-only extensions from non-NSFW categories

Kraptor123:
https://github.com/user-attachments/assets/dd683e3e-ddf6-48be-a31b-6b2ce46cb35c

Currently, pinning an NSFW extension makes it visible in all categories (including the default "Movies" and "TV Series" view), which can clutter the UI with adult content when it's not desired. With this change, pinned NSFW-only extensions will only be visible if the user explicitly selects the NSFW chip.

Modified HomeFragment.selectHomepage to filter out pinned, strictly NSFW extensions if TvType.NSFW is not in the selected categories.


fire-light42 (approved):
LGTM 👍

PR #2925, opened 2026-06-15
feat(extractors): add flyfile.app extractor

Bnyro:
- Example URL to test with: https://flyfile.app/v/vsSkhe3XMBZ0yN8
- From my tests, the player timed out 90% of times. I'm not sure if that's only a temporary issue or if the service is generally so slow (it's the same on the website). The API always assigned me to `https://streaming-de-2.flyfile.app`, so maybe in other regions, it's faster.

I've already used KotlinX for this, I'm not really 100% sure of the state of it in Cloudstream, but it seems to work.

Great work on KotlinX btw @Luna712!


fire-light42 (approved):
Code looks good, (included test) works well, and even annotated with Prerelease. Great pull request!

PR #2931, opened 2026-06-16
Add new splitUrlParameters API accepting Ktor Url or string

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -755,18 +755,64 @@ fun MainAPI.fixUrl(url: String): String {
...
+ *
+ * @param url the [Url] whose query parameters should be extracted.
+ * @return a map of decoded query parameter names to their first decoded value.
+ *
+ * @sample
+ * splitQuery(Url("https://example.com/path?foo=bar&baz=qux"))
+ * // returns {"foo": "bar", "baz": "qux"}
+ */
+@Prerelease
+fun splitQuery(url: Url): Map<String, String> {
```
fire-light42: I would prefer to keep the API-surface somewhat dependency-agnostic. There is no need to make the same mistake as `splitQuery(url: java.net.URL)` again, when we can just expose a string argument instead.

fire-light42 (approved):
It works well from testing and the code looks good 👍 

I just have a thought regarding the new API surface. This is not blocking. I just want to hear what you think before merging this.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
Luna712: What I was aiming for was to expose a way to accept a direct URL as well as String. By passing a URL it can sometimes be a bit more type safe as well, and also as we shift towards ktor URL more, sometimes a URL may be wanted directly, in which case, I just thought it was better and more straightforward to accept a URL with a separate convenience method accepting a string only as well. I am open to changes though if you think it should be done differently?

Luna712: I am kinda considering whether this should instead be an extension function on String in StringUtils though, in which case it would make a little more sense to simply drop the URL overload and just accept that extension method. I am honestly not 100% sure on the best option here...

Luna712: Hmm. Maybe better not in StringUtils (or is it, still haven't decided that) but perhaps it also being called something like `splitUrlParameters` or `getUrlParameters` would make more sense and be more clear than `splitQuery`?

Luna712: I am thinking maybe just name `splitQuery`, `parseUrlParameters` instead. It's a more clear naming scheme, and is more consistent with other Url-based naming.

Luna712: Okay I just pushed a little bit more clear method name now anyway and used the String method as the default recommended version in ReplaceWith, I still think we should keep allowing to pass a ktor URL directly if wanted though.

PR #2932, opened 2026-06-18
SyncApi Search query fix

KingLucius:
Passing real query instead of fixed string name & code formatting


fire-light42 (approved):
Very good catch! ❤️  Looks like the name shadowing from the old API made the bug slip through.

PR #2940, opened 2026-06-21
Add debug logging for if Jackson or kotlinx serialization is used

Luna712:


Luna712:
@fire-light42 could this be merged at some point if you're okay with it btw? It helps to debug primitive types at least and other things that the class tester doesn't pick up. It shouldn't spam logs on pre or stable, only logs on debug version but helps me to debug things without making a branch mess for now. If you don't want it though that is fine too, I can figure something else out.

fire-light42:
I am not a huge fan of further touching the inline code :grimacing:

Luna712:
> I am not a huge fan of further touching the inline code 😬

I don't really see a way that would break anything at all, however I also didn't think it would before. but debugPrint exists on all versions so binary should be compatible. Up to you though.

fire-light42:
Yes, it *should* not break anything, but with `debugPrint` also being inline it really just makes things too tricky. This would silently give us issues in pretty much all extensions if we refactor either `debugPrint` or `AppDebug` (which is `InternalAPI`). 

I am sorry, but I cannot merge this :disappointed:

PR #2947, opened 2026-06-23
Add a temporary SkipSerializationTest annotation

Luna712:
For issues such as https://github.com/recloudstream/cloudstream/pull/2878#discussion_r3452107676

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
@@ -50,7 +50,7 @@ import kotlin.time.Instant
  * API available only on prerelease builds.
  * Using it will cause stable to crash with `NoSuchMethodException`.
  */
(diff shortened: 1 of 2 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -50,7 +50,7 @@ import kotlin.time.Instant
  * API available only on prerelease builds.
  * Using it will cause stable to crash with `NoSuchMethodException`.
  */
-@MustBeDocumented // Same as java.lang.annotation.Documented
+@MustBeDocumented
```
Luna712, line 53: Note: I removed this comment because I didn't want to do I separate PR just for that. This comment actually confused me a bit. Also we don't mention same as for every other Java -> Kotlin replacement, and as we push more towards KMP this wouldn't be the case in other platforms. I could readd it if absolutely wanted though.

fire-light42 (approved):
Perfect!

PR #2949, opened 2026-06-23
SerializationClassTester: show all failures at once

Luna712:
This is helpful for batch testing rather than testing and fixing one at a time. This is what I was using to test many at once.


fire-light42 (approved):
Nice!

PR #2960, opened 2026-06-26
SerializationClassTester: skip abstract classes

Luna712:
Necessary for some such as LibrarySearchResponse and DownloadCached. Using the skip annotation causes it to also skip child classes which we don't want so just skip abstract classes since Instancio can't create an instance of an abstract class: "It is an abstract class and no subtype was provided".


fire-light42:
I think we can sidestep the issue by using interfaces on some of these classes, such as DownloadCached.

Luna712:
Interfaces has the same issue. And personally I would also prefer to not change things to much during the migration if possible but up to you.

Luna712:
Sorry I just realized I wasn't to clear. An interface would probably work for DownloadCached but I don't think it would for LibrarySearchResponse and thus this would still be needed. I also wanted to avoid changing the actual inheritance during this migration but could if you wanted.

PR #2961, opened 2026-06-26
fix: disable QUIC in Cronet to fix error 2004/2001 on Android TV

kesarpatidar19-spec:
Cronet has enableQuic(true) which uses UDP port 443. Many TV  routers/ISPs block or throttle UDP. Phones recover gracefully;  Cronet on TV gets stuck and fails with error 2004/2001 on sources  that work fine on mobile on the same Wi-Fi.

Fix: enableQuic(false) forces Cronet to always use TCP.


kesarpatidar19-spec:
Don't accept it I'll download the apk then test it

kesarpatidar19-spec:
Accidently closed pull
I'm testing soon

PR #2963, opened 2026-06-26
extractors: migrate to kotlinx serialization

Luna712:


fire-light42:
I will try to test and merge soon to prevent conflicts.

fire-light42 (approved):
Looks good and the var to var change does not seem to break any existing extensions

PR #2958, opened 2026-06-25
ControllerActivity: migrate to kotlinx serialization

Luna712:

PR #2952, opened 2026-06-24
AniListApi: migrate to kotlinx serialization

Luna712:
This also fixes some circular reference issues that technically worked (albeit poorly), with Jackson but don't in kotlinx serialization. The tests caught that for me.

PR #2964, opened 2026-06-26
DataStore: deprecate toKotlinObject

Luna712:
1. It can sometimes cause type erasure
2. It is less clear what it is doing
3. It is easy to replace
4. It is very rarely used, all app usage has already been replaced, and I haven't found any extensions actually using it but still following normal deprecation cycle as some still might be using it.


Luna712:
Still needs #2952 and #2958 first

app/src/main/java/com/lagradost/cloudstream3/utils/DataStore.kt
```diff
@@ -181,33 +181,49 @@ object DataStore {
...
 
+    @Deprecated(
+        message = "Use parseJson<T>(this) directly instead.",
+        level = DeprecationLevel.WARNING,
+        replaceWith = ReplaceWith(
+            expression = "parseJson<T>(this)",
+            imports = ["com.lagradost.cloudstream3.utils.AppUtils.parseJson"],
+        ),
+    )
     inline fun <reified T : Any> String.toKotlinObject(): T {
```
fire-light42, line 204: Can't we keep this extension function? It should not cause type erasure and it could be renamed for clarity.

Having extension functions helps with chained calls such as:

```kt
text.lowercase()
    .substringAfter("[")
    .substringBefore("]")
    .trim()
    .parseJson<…>()
```
And it would be a good complement to `toJson()`

Luna712: We should still deprecate this and add the new version to AppUtils then I think. I am a little busy IRL now to work on things though but should be back more next week again.

PR #2967, opened 2026-06-27
fix: skip broken CF bypass on Android TV + recover from mid-stream ti…

kesarpatidar19-spec:
…meouts

Fix for Android TV playback errors:

1. CloudflareKiller: detect WebView availability at startup. On many Android TV / FireStick devices WebView is absent — CookieManager throws (the code already had a comment about this), CF challenge is never solved, request returns 403 → error 2004. With WebView unavailable we skip the bypass and fall through to a direct request instead of silently failing. No behaviour change on phones (isWebViewAvailable always true).

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/network/CloudflareKiller.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/network/CloudflareKiller.kt
@@ -1,7 +1,9 @@
 package com.lagradost.cloudstream3.network
 
+import android.os.Build
 import android.util.Log
 import android.webkit.CookieManager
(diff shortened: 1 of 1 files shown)


Luna712:
There is no evidence this will actually fix anything. And also please read the [AI Policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md). The description sounds like something AI would hallucinate true or not, actually human testing should be what confirms it not what AI says.

fire-light42:
I am converting your pull requests to drafts until you demonstrate that the pull requests are making TV playback better, and it is not an AI hallucination.

PR #2968, opened 2026-06-27
fix: skip broken CF bypass on Android TV + recover from mid-stream ti…

kesarpatidar19-spec:
…meouts

Fix:

1. CS3IPlayer: also recover from ERROR_CODE_IO_NETWORK_CONNECTION_TIMEOUT mid-stream the same way CONNECTION_FAILED is already handled — call prepare() instead of giving up. TV network stacks produce transient timeouts during buffering even when the stream was already playing. The duration != TIME_UNSET guard ensures we still switch mirrors if the stream never started.


Luna712:
There is no evidence this will actually fix anything. And also please read the [AI Policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md). The description sounds like something AI would hallucinate true or not, actually human testing should be what confirms it not what AI says.

PR #2976, opened 2026-06-29
Replace javaClass with Kotlin ::class for consistency

Luna712:
I was originally just going to replace it in CustomSite, but figured I would do it all for consistency. When retrieving the data from CustomSite they would both resolve to the same thing so is backwards compatible. However, `::class.simpleName` can return null if something really goes wrong so we do handle that a little different for PlayerSettings, and bail out rather than setting to an empty string for CustomSite just to ensure consistency with current behavior also.

Eventually this will be needed anyway as we migrate more to KMP in library and Compose Multiplatform for UI, which as I am building compose, I am avoiding JVM specific stuff also so if we later expand to other platforms besides Android and JVM Desktop we can more easily do that as well.

I used Jackson's JsonAlias and Kotlinx Serialization's JsonNames annotations for CustomSite to maintain backwards compatibility for deserialization but future serialization would use the new name, so `parentClassName` rather than `parentJavaClass`.

I also added some extra comments to SettingsPlayer as the code kinda confused me, like what `add("")` was for, while I figured it out fairly quickly, I figured it didn't hurt to add comments to make it more clear in the future.


fire-light42 (approved):
Good changes

PR #2977, opened 2026-06-29
EmturbovidExtractor: replace selectXpath

Luna712:
This will be necessary when we finally fully migrate to ksoup, which currently doesn't have selectXpath, but for our usages that is not necessary. This way is also more consistent with how we do it everywhere else.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/EmturbovidExtractor.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/EmturbovidExtractor.kt
@@ -14,7 +14,7 @@ open class EmturbovidExtractor : ExtractorApi() {
 
     override suspend fun getUrl(url: String, referer: String?): List<ExtractorLink>? {
         val response = app.get(url, referer = referer ?: "$mainUrl/")
-        val playerScript2 = response.document
+        val playerScript = response.document
             .select("script")
             .first { it.data().contains("var urlPlay") }
             .html()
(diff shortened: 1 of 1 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/EmturbovidExtractor.kt
```diff
@@ -8,35 +8,32 @@ import com.lagradost.cloudstream3.utils.Qualities
...
-        val response = app.get(
-            url, referer = referer ?: "$mainUrl/"
-        )
-        val playerScript =
-            response.document.selectXpath("//script[contains(text(),'var urlPlay')]")
-                .html()
+        val response = app.get(url, referer = referer ?: "$mainUrl/")
+        val playerScript = response.document
+            .select("script:contains(var urlPlay)")
+            .html()
```
fire-light42: `:contains` does not seem to work when testing. 

Use

```suggestion
        val playerScript2 = response.document
            .select("script")
            .first { it.data().contains("var urlPlay") }
            .html()
```

fire-light42 (changes requested):

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/EmturbovidExtractor.kt
Luna712: oh wait, I think this would have worked using selectFirst instead of select like we do everywhere else for script:contains. Anyway I applied your suggestion, no issues with that instead if you rather but selectFirst might be a little cleaner.

fire-light42: If you can make a selectFirst which works it would be great! Right now it seems like the `:contains` selector does not work for this case.

Luna712: I suppose we can just do it this way since we know it works I may investigate this more in the future still though.

fire-light42 (approved):

PR #2990, opened 2026-07-02
Replace some JVM specific stuff in JsHunter and JsUnpacker

Luna712:


Luna712:
These were things found after attempting to actually compile to JS btw.

fire-light42 (approved):

PR #2991, opened 2026-07-03
Add more space to fix last menu item half-cut

KingLucius:
Fixes last item in the `BottomSheetDialog` half-cut on TV

### DIFF
--- a/app/src/main/res/layout/bottom_selection_dialog.xml
+++ b/app/src/main/res/layout/bottom_selection_dialog.xml
@@ -37,6 +37,7 @@
         android:paddingTop="10dp"
         android:dividerHeight="1dp"
         android:requiresFadingEdge="vertical"
+        android:paddingBottom="32dp"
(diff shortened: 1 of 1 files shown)


Luna712 (approved):
Nice thanks for fixing this issue!

fire-light42:
I cannot replicate the issue :confused: 
Not sure about the looks on phone too

KingLucius:
> I cannot replicate the issue :confused: 
> Not sure about the looks on phone too

I can make it conditional for TVs only, but you can't reproduce it ? 🤔😅

fire-light42:
No I cannot, or maybe I am missing something. An image would help demonstrate the issue.

KingLucius:
Hold middle button on play episode button (show options) and scroll to last menu item, it would be half-cut

fire-light42:
I think the issue is due to overscan, which is exclusive to TV, but I think the padding is okay on phone :thinking:

PR #2992, opened 2026-07-03
feat(extractors): fix vidara and update list of mirrors

Bnyro:
Example URL:
- https://vidaarax.net/e/xB5Cf_RkGVbZ?kk111

Side notes:
- Strmup and Streamix are dead, so I removed them.
- I also migrated to KotlinX Serialization with these changes.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidara.kt
```diff
@@ -52,18 +74,19 @@ open class Streamup() : ExtractorApi() {
         }
     }
 
+    @Serializable
     private data class StreamUpFileInfo(
         val title: String,
         val thumbnail: String,
-        @JsonProperty("streaming_url")
+        @SerialName("streaming_url")
```
Luna712, line 82: Please keep both JsonProperty and SerialName for now, especially when the name here is different then the variable name, as otherwise our migration serialization tests will fail. Thank you!

Bnyro: Thanks for the hint, done :+1:

fire-light42 (approved):
Nice!

PR #2993, opened 2026-07-03
feat(extractors): add firestream.to extractor

Bnyro:
Example URL:
- https://firestream.to/v/54h2ZiYZ

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Firestream.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Firestream.kt
@@ -1,0 +1,52 @@
+package com.lagradost.cloudstream3.extractors
+
+import com.lagradost.cloudstream3.Prerelease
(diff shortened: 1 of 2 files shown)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Firestream.kt
```diff
@@ -0,0 +1,50 @@
...
+        )
+    }
+
+    override fun getExtractorUrl(id: String): String {
+        return "$mainUrl/e/$id"
+    }
+
+    @Serializable
+    private data class VideoResponse(
+        val signedVideoUrl: String,
```
Luna712, line 50: In the past we really haven't, and I can do this later during my mass migrations if you would rather also, but please provide SerialName to *all* properties. That will help to ensure proper expected behavior and make things more obvious from a code maintenance standpoint also. I am mostly just mentioning for future awareness, for now it is not an absolute requirement as I can do this during my mass migrations also.

PR #2997, opened 2026-07-03
Add expect/actual for WorkerThread

Luna712:
This will later be necessary for other targets as it only exists on JVM platforms. Other targets would just define it as a base annoation that doesn't do anything, allows us to keep thread safety on JVM platforms but still be able to compile for others that don't have it.


Luna712:
Also just to note, besides my other currently open PRs and removal of deprecated stuff + NiceHttp and Jsoup migration, and remaining kotlinx serialization migration (which I have all done myself already). I was able to fully compile the library to JS. This was basically the final missing piece of it.

fire-light42:
I will try to get your pull requests merged soon!

I am really looking forward a JS library :rocket:

PR #3001, opened 2026-07-04
shortcode: Add py.md as alternative to cutt.ly

Kraptor123:
cutt.ly is currently blocked in Turkiye, so users are getting connection errors when trying to load repos with shortcodes. I wrapped the request in a try-catch and added py.md as a fallback. Now if cutt.ly fails, it just tries py.md instead.


fire-light42 (changes requested):
The code and motivation are both good, but this option opens up vulnerabilities in the code system. For example what if someone maliciously registers a short code for a popular repository with `py.md`? 

I think the best option is to have a different prefix for `py.md` links, or something to visibly distinguish them from known short codes. Different users should not get different repositories with the same short code.

Kraptor123:
i can add all known repos to my account if thats okey so users can access every repo without any problem.

fire-light42:
I like the idea of "!", which we can easily migrate away from in the future if we decide to remove cutt.ly. 

However, creating all short codes leaves your account as another single point of failure. I recommend that you instead communicate with fellow repository maintainers to let them create their own short codes before I merge this.

fire-light42 (approved):
Code looks good, idea is good, just want some feedback from other developers.

app/src/main/java/com/lagradost/cloudstream3/plugins/RepositoryManager.kt
```diff
@@ -141,13 +141,24 @@ object RepositoryManager {
...
-                        return@safeAsync url
+                if (fixedUrl.startsWith("!")){
+                    app.get("https://py.md/${fixedUrl.removePrefix("!")}", allowRedirects = false).let { it2 ->
+                        it2.headers["Location"]?.let { url ->
+                            if (url.startsWith("https://py.md/404")) return@safeAsync null
+                            if (url.removeSuffix("/") == "https://py.md") return@safeAsync null
+                            return@safeAsync url
+                        }
+                    }
+                }else {
```
Bnyro:
```suggestion
                } else {
```
nit: formatting

app/src/main/java/com/lagradost/cloudstream3/plugins/RepositoryManager.kt
```diff
@@ -141,13 +141,24 @@ object RepositoryManager {
             }
         } else if (fixedUrl.matches("^[a-zA-Z0-9!_-]+$".toRegex())) {
             safeAsync {
-                app.get("https://cutt.ly/${fixedUrl}", allowRedirects = false).let { it2 ->
-                    it2.headers["Location"]?.let { url ->
-                        if (url.startsWith("https://cutt.ly/404")) return@safeAsync null
-                        if (url.removeSuffix("/") == "https://cutt.ly") return@safeAsync null
-                        return@safeAsync url
+                if (fixedUrl.startsWith("!")){
```
Bnyro:
```suggestion
                if (fixedUrl.startsWith("!")) {
```
nit: formatting

Bnyro (commented):
@Kraptor123 fyi, you can auto-format the code with Intellij IDEA / Android Studio using `Ctrl + Alt + L`.

Otherwise somebody else will probably do these formatting changes by accident when changing this file the next time :)

Kraptor123:
@fire-light42 Apologies for the tag. Would it be possible to add support for the megarepo shortcode to pymd?

PR #3002, opened 2026-07-05
Fix regex to match quality case-insensitively for Doodstream

danzybae:
The regex used to extract video quality (`\d{3,4}p`) only matched lowercase `p`, causing it to miss titles that use uppercase `P` (e.g. `1080P` instead of `1080p`). As a result, the quality value would return `null` for those cases.

Updated the quality regex from `\d{3,4}p` to `\d{3,4}[pP]` so it matches the resolution suffix regardless of case (`p` or `P`).


fire-light42 (approved):
Thank you for your first contribution!

PR #3007, opened 2026-07-06
fix(extractors) fix gupload extractor

Bnyro:
They no longer base 64 encode the player config., so the fix is as easy as removing that part

Example URL:
- https://gupload.xyz/data/e/b8df1c423c91


fire-light42:
Looks good, I will test soon :+1:

fire-light42 (approved):

PR #3027, opened 2026-07-07
DataStoreHelper: migrate to kotlinx serialization

Luna712:


Luna712:
This needs #2874 first. This is also the very last migration where classes actually have JsonProperty, so the last obvious migration, still possible there are some that were missed that were not as obvious.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -63,13 +71,11 @@ class UserPreferenceDelegate<T : Any>(
...
         property: KProperty<*>,
-        t: T?
+        t: T?,
     ) {
         if (t == null) {
             removeKey(realKey)
-        } else {
-            setKeyClass(realKey, t)
-        }
+        } else setKeyClass(realKey, t)
```
fire-light42: I would prefer if we keep curly brackets, to prevent bugs such as [this](https://github.com/recloudstream/cloudstream/commit/55b5017068efd32a11088d8dd6f7af9353a6854a), even if they are redundant.

Luna712: Thanks. Though, I don't personally think it is necessary for definitive one line/one call stuff, however it doesn't really matter that much to me either way, so done.

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -232,47 +240,52 @@ object DataStoreHelper {
...
-        @JsonProperty("type") override var type: TvType?,
-        @JsonProperty("posterUrl") override var posterUrl: String?,
-        @JsonProperty("year") open val year: Int?,
-        @JsonProperty("syncData") open val syncData: Map<String, String>?,
-        @JsonProperty("quality") override var quality: SearchQuality?,
-        @JsonProperty("posterHeaders") override var posterHeaders: Map<String, String>?,
-        @JsonProperty("plot") open val plot: String? = null,
-        @JsonProperty("score") override var score: Score? = null,
-        @JsonProperty("tags") open val tags: List<String>? = null,
+        @Transient override var id: Int? = null,
```
fire-light42, line 276: These Transient annotations need a small explanatory comment for those not aware of the abstract issues.

Luna712: Done.

fire-light42 (approved):
Very good job on the migration pull requests!

PR #3034, opened 2026-07-08
test: add an IgnoreOnWeb annotation

Luna712:
For some things like coroutines, when compiled to JS tests will fail because it is single threaded, and it can't cancel a synchronous operation from a single threaded system. That causes a few tests from #2812 to fail on JS only, rather than splitting the location for the tests per platform, which can make it harder to find, just ignoring the tests would be better. With this we could then just add `@IgnoreOnWeb` to the tests we don't care about. The actual in `webTest` would later be:

```kt

import kotlin.test.Ignore

actual typealias IgnoreOnWeb = Ignore
```
and because of `@OptionalExpectation`, we don't need to add actual to any other source sets. I decided to do this as a different PR rather than in #2812 because it is kinda different scope after all, and I don't want to drift that scope to much.

I also used `Ignore.kt` as the file name because `IgnoreOnWeb.kt` would mean for `webTest` it would be `IgnoreOnWeb.web.kt` which just seems a bit odd. Additionally this way if we ever need we could also add more in the same file like `IgnoreOnJvm` or `IgnoreOnAndroid`.

### DIFF
--- a/library/src/commonTest/kotlin/com/lagradost/cloudstream3/utils/Ignore.kt
+++ b/library/src/commonTest/kotlin/com/lagradost/cloudstream3/utils/Ignore.kt
@@ -1,0 +1,6 @@
+package com.lagradost.cloudstream3.utils
+
+@OptIn(ExperimentalMultiplatform::class) // OptionalExpectation is an experimental annotation for now
+@OptionalExpectation
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+expect annotation class IgnoreOnWeb()
--- a/library/src/commonTest/kotlin/com/lagradost/cloudstream3/utils/JsInterpreterTest.kt
+++ b/library/src/commonTest/kotlin/com/lagradost/cloudstream3/utils/JsInterpreterTest.kt
@@ -2129,6 +2129,7 @@ class JsInterpreterTest {
     }
 
     @Test
+    @IgnoreOnWeb
     fun suspendEvalJsWithTimeoutCancelsInfiniteLoop() = runTest {
         /**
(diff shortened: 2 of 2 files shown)


library/src/commonTest/kotlin/com/lagradost/cloudstream3/utils/Ignore.kt
```diff
@@ -0,0 +1,6 @@
+package com.lagradost.cloudstream3.utils
+
+@OptIn(ExperimentalMultiplatform::class) // OptionalExpectation is an experimental annotation for now
+@OptionalExpectation
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+expect annotation class IgnoreOnWeb()
```
fire-light42, line 6: Is the goal of this to be used exclusively in tests or elsewhere too? 
If this is only for tests, it should be named differently (`IgnoreTestOnWeb`) to discourage usage elsewhere.

Luna712: It can't be used elsewhere because it is in commonTest.

Luna712: I also used this name because it seemed to be the convention I found in other public repos. Not really against changing it but I think for something that is only compiled on tests anyway, adding Test to the annotation name reads a bit odd.

fire-light42: Nice! I missed that it was only usable in commonTest, disregard my earlier comment.

fire-light42 (approved):
Good thinking

PR #3043, opened 2026-07-10
feat(Sync): Scrobble Support + SIMKL Scrobble

KingLucius:
Changes:
- Scrobble SyncApi support
- Simkl scrobble

Reason:
That would make sync so much easier, instead of manual sync, we just tell simkl server the event and the progress and it updates the watching now and mark it as watched after 80% progress

Screenshot:
Screenshot_20260710_110534_com_android_chrome_ChromeTabbedActivity_edit_505619097329775

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/syncproviders/AuthAPI.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/syncproviders/AuthAPI.kt
@@ -16,7 +16,9 @@ import java.net.URI
 import java.security.SecureRandom
 
 data class AuthLoginPage(
-    /** The website to open to authenticate */
+    /**
+     * The website to open to authenticate
+     */
     val url: String,
     /**
      * State/control code to verify against the redirectUrl to make sure the request is valid.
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/lagradost/cloudstream3/syncproviders/providers/SimklApi.kt
```diff
@@ -1083,4 +1087,79 @@ class SimklApi : SyncAPI() {
...
+            ).isSuccessful
+        } catch (e: Exception) {
+            logError(e)
+            false
+        }
+    }
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    data class ScrobbleEpisode(
+        @JsonProperty("season") val season: Int,
```
fire-light42: Please make these serializable with a `@Serializable` annotation and corresponding `@SerialName` annotations:

```kt
@JsonInclude(JsonInclude.Include.NON_EMPTY)
@OptIn(ExperimentalSerializationApi::class) // KeepGeneratedSerializer is an experimental annotation for now
@Serializable(with = ScrobbleEpisode.Serializer::class)
@KeepGeneratedSerializer
data class ScrobbleEpisode(
    @JsonProperty("season") @SerialName("season") val season: Int,
    @JsonProperty("episode") @SerialName("episode") val episode: Int,
) {
    object Serializer :
        NonEmptySerializer<ScrobbleEpisode>(ScrobbleEpisode.generatedSerializer())
}
```

app/src/main/java/com/lagradost/cloudstream3/syncproviders/AuthAPI.kt
```diff
@@ -148,6 +148,11 @@ abstract class AuthAPI {
     /** The requirements to login in app */
     open val inAppLoginRequirement: AuthLoginRequirement? = null
 
+    /** Determine if SyncApi supports Playback Scrobbling.
```
Luna712, line 185: These comments should follow the format here: https://kotlinlang.org/docs/kotlin-doc.html so `/**` then newline and ending with `*/` rather than `*   */`. I have been cleaning this up as I come across them but thought I would mention it just for future reference.

rajarshikhatua:
Please add an option to set the percentage of progress at which a movie is automatically marked as watched.

And, add a rewatch feature for multiple viewings of the same movie or episods. All of this is supported by simkl

KingLucius:
> Please add an option to set the percentage of progress at which a movie is automatically marked as watched.
> 
> And, add a rewatch feature for multiple viewings of the same movie or episods. All of this is supported by simkl

Every service determine a certain percentage to mark the media as watched, usually it's 80% like Simkl & trakt.
For scrobbling we only need to report the progress and event (play/pause/stop) and service will change status accordingly.
If you start the media again it will be in rewatching status

PR #3046, opened 2026-07-10
remove prerelease annotations

fire-light42:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/ExtractorApi.kt
```diff
@@ -472,7 +470,6 @@ val WIDEVINE_DRM_UUID = Uuid.fromLongs(-0x121074568629b532L, -0x5c37d8232ae2de13
  * PlayReady is supported on all AndroidTV devices. Note that most other Android devices do not
  * provide PlayReady support.
  */
-@Prerelease
```
Luna712, line 479: I think these should stay prerelease (or maybe we just shouldn't fully deprecate the old ways yet) until the next stable because the Kotlin 2.4 patch wasn't merged before so Kotlin UUIDs are still technically expiremental and compatibility could have changed between 2.3 and 2.4 but up to you.

Luna712: I would also say the same for newDrmExtractorLink with Kotlin Uuid. Again up to you though. I am just thinking to ensure we have a stable API before actively recommending extensions to use it.

PR #3051, opened 2026-07-10
M3u8Helper: add new suspended getDecryptedBytes()

Luna712:


library/src/commonMain/kotlin/com/lagradost/cloudstream3/utils/M3u8Helper.kt
```diff
@@ -100,15 +100,15 @@ object M3u8Helper2 {
     }
 
     @OptIn(DelicateCryptographyApi::class)
-    fun getDecrypted(
+    suspend fun getDecrypted(
```
fire-light42: This breaks the ABI, contrary to expectations.

Luna712: Yes I know and forgot to add a PR description (the whole reason I just opened this PR lol). What I wanted was some opinion on how not to do that while still doing a suspended version of this.

fire-light42: I know of no good way beyond the standard deprecation method :confused:

Luna712: Okay I will figure something out later then lol. Thanks for the comment though.

Luna712: I am marking this as a draft until then.

Luna712: I have added a new suspended getDecryptedBytes to replace the old. The name made sense and allowed it to avoid a conflict and still make a new suspend version and maintain ABI compatibility.

PR #3058, opened 2026-07-12
Migration guide

fire-light42:
@Luna712


Luna712:
Nice! I think it can be expanded more later but good for now.

Luna712 (approved):

fire-light42:
Perfect! Since it is a tracked file everyone is welcome to update the migration guidelines. More details will of course be apparent as we perform the migration.

PR #3059, opened 2026-07-12
Add vpn support natively in cloudstream

NivinCNC:
Added native vpn support. But need some suggestions in cronet since it obeys only system wide proxy and doest support proxy itself

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/CloudStreamApp.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/CloudStreamApp.kt
@@ -84,6 +84,9 @@ class CloudStreamApp : Application(), SingletonImageLoader.Factory {
         }
 
         AppDebug.isDebug = BuildConfig.DEBUG
+
+        // Auto-restore VPN connection in the background if enabled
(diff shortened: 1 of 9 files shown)


fire-light42:
Please see our [AI policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md).

Proxy support could work in CloudStream, but this implementation is very bloated. Ideally the implementation should be about ±100 lines (tests should be separated). Additionally, the references to VPNs are misleading, this feature should be advertised (and programmed) as proxy support, even if users only know what a VPN is.

NivinCNC:
I admit this code is by ai and sorry for not mentioning it, i just wanna outline the vpn implementation and standardize it later, since i wanted to know wheter disabling cronet if vpn is enabled is okai or not. If that's confirmed, i can provide a quality code at my level best.

fire-light42:
Yes, Cronet may be disabled when a proxy is enabled, we already disable it whenever interceptors are used :+1: 

It is preferable if Cronet can be made compatible, but I understand that it may not be possible.

PR #3068, opened 2026-07-14
More robust date parsing

fire-light42:
The date parsing fails at "2026-05-17 12:30:45" when the previous implementation managed to parse it. This was the best solution but it still loses the "12:30:45" part, which is fair considering the format is "yyyy-MM-dd".


Luna712 (approved):
Thank you! Sorry I missed this one...

PR #3072, opened 2026-07-15
Add actual declarations for webMain/webTest

Luna712:


library/src/webMain/kotlin/com/lagradost/api/Log.kt
```diff
@@ -0,0 +1,19 @@
+package com.lagradost.api
+
+actual object Log {
+    actual fun d(tag: String, message: String) {
+        console.log("[$tag] $message")
```
fire-light42, line 7: Not console.debug?

fire-light42 (approved):
Looks good! I assume web target is coming soon

library/src/webMain/kotlin/com/lagradost/api/Log.kt
Luna712: The choice was because console.debug as a lower priority than console.log so thought this was a little closer to JVM methods, though I could do console.debug if you'd rather.

fire-light42: I see, I do agree with the choice, but the motivation should probably be inside a short comment.

Luna712: I am about to go out for awhile. I will add the comment later. This isn't fully urgent as a full web target is still blocked on a couple other things.  But yes it is coming soon, finally.

fire-light42: No stress, you have done a great job!

Luna712: Hmm. I was thinking on this more, perhaps console.debug would be better here, while it may be a bit different than the JVM implimentation, it is a more expected convention in JS environments... was wondering your thoughts though.

fire-light42: I was thinking we can go with the expected console.debug and switch to console.log if we notice visibility programs when we run the web version.

fire-light42: Nonetheless, both are good options.

Luna712: Lol I just realized there is no console.debug in Kotlin/JS for some reason...

fire-light42: Well that settles it 😆 

Looks like it is coming in 2.5 at least https://github.com/JetBrains/kotlin/pull/5474

Luna712: Nice! I added a comment anyway.

PR #3073, opened 2026-07-15
Use cryptography-kotlin in the app itself also

Luna712:
It is more type safe and IMO cleaner code.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
@@ -112,7 +112,6 @@ import kotlinx.coroutines.delay
 import okhttp3.Interceptor
 import org.chromium.net.CronetEngine
 import java.io.File
-import java.security.SecureRandom
 import java.util.UUID
 import java.util.concurrent.Executors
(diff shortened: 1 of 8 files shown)


Luna712:
Note: I have had this patch ready to go for quite awhile, but wanted to wait for the one that added the kotlinx-io dependency so I could use it. I can't fully remember just how much of this was actually fully tested so there may indeed be some potential issue with it.

app/src/main/java/com/lagradost/cloudstream3/ui/player/CS3IPlayer.kt
```diff
@@ -1893,7 +1892,7 @@ class CS3IPlayer : IPlayer {
             if (ignoreSSL) {
                 // Disables ssl check
                 val sslContext: SSLContext = SSLContext.getInstance("TLS")
-                sslContext.init(null, arrayOf(SSLTrustManager()), SecureRandom())
+                sslContext.init(null, arrayOf(SSLTrustManager()), null)
```
Luna712, line 1895: Note: this was just removed because it has the same behavior as far as I could tell, null just uses a default SecureRandom and we don't need to pass it so I thought we should just let it handle it there to remove all usage of SecureRandom. But it can be changed if really wanted.

PR #3076, opened 2026-07-16
Library: use api for some dependencies

Luna712:
I am not 100% sure if we should. This will avoid extensions having to manually add them also allowing us to make sure we standardize versions, but it can also slow the build down according to https://docs.gradle.org/current/userguide/performance.html#use_implementation_for_internal_dependencies

### DIFF
--- a/library/build.gradle.kts
+++ b/library/build.gradle.kts
@@ -53,19 +53,19 @@ kotlin {
         }
 
         commonMain.dependencies {
-            implementation(libs.annotation) // Annotations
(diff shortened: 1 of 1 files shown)


Luna712:
The common convention for libraries does seem to be to use api() so that is what I do think we should do.

Luna712:
Eventually but I am thinking to wait on this for now so marking as draft.

fgmitesh:
Hm

PR #3085, opened 2026-07-19
Fix posterHeaders not passed in home preview carousel

danzybae:
`HomeScrollAdapter.onBindContent()` calls `loadImage(posterUrl)` without passing `item.posterHeaders`, for both `HomeScrollViewBinding` and `HomeScrollViewTvBinding`. This means posters/backgrounds from providers that require custom headers may fail to load in the home preview carousel.


fire-light42 (approved):
Nice find

PR #3086, opened 2026-07-20
Add profile pictures from files

### LINKED ISSUE #3079: Account PFP from file

My suggested feature is to implement a option to use a PNG or jpg on your device as the account profile picture. This will greatly improve the experience on TV as it is a pain to get the image URL

### Other details

In addition to this, the edit icon for the picture cannot be focused on TV, it must use a mouse to access and select it, please look into that issue

### Acknowledgements

arpitagarwal1301:
## What does this change?

Adds a file option to the profile image editor while preserving the existing URL flow.

Selected image documents must decode successfully before preview. Read access is persisted only when the edit is applied, and an inaccessible document falls back to the account's default avatar. Replaced and deleted document grants are released only after no account still references them.

This review revision follows the requested architecture by exposing the profile-image picker through `CommonActivity`. Each initialized activity owns its lifecycle-safe launcher, while stale picker results and superseded validation requests are ignored. Account UI call sites no longer thread picker callbacks through adapters and dialogs.

The edit icon also has a content description and explicit D-pad focus navigation while retaining its existing touch background.

Closes #3079

## Testing

- `:app:compileStableDebugKotlin`
- `:app:testStableDebugUnitTest` (8 tests, 0 failures)
- `:app:lintStableDebug` (0 errors)
- `git diff --check`

## Limitations

- The system picker, activity recreation, and TV focus path were not exercised with UI instrumentation or on physical hardware.
- The optional internal-file copy was not added; unavailable external documents fall back to the default avatar.

## AI disclosure

AI-assisted: implementation and review used OpenAI Codex; the listed checks were run on the submitted code.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
@@ -4,15 +4,14 @@ import android.app.Activity
 import android.content.Context
 import android.content.DialogInterface
 import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
 import android.text.Editable
 import android.view.LayoutInflater
 import android.view.inputmethod.EditorInfo
 import android.widget.TextView
 import android.widget.Toast
-import androidx.activity.ComponentActivity
 import androidx.activity.result.ActivityResultLauncher
-import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.StringRes
 import androidx.appcompat.app.AlertDialog
 import androidx.core.view.isGone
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
```diff
@@ -43,6 +47,38 @@ import com.lagradost.cloudstream3.utils.UIHelper.showInputMethod
 import com.lagradost.cloudstream3.utils.UIHelper.showProgress
 
 object AccountHelper {
+    private fun pickProfileImage(context: Context, callback: (String) -> Unit) {
+        val activity = context.getActivity() as? ComponentActivity ?: return
+        var launcher: ActivityResultLauncher<Array<String>>? = null
+        try {
+            launcher = activity.activityResultRegistry.register(
```
fire-light42: Please use registerForActivityResult instead, to get a safer and lifecyle-aware file picking. It may be more difficult, but it removes the dirty `getActivity()` and having to unregister.

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
```diff
@@ -43,6 +47,38 @@ import com.lagradost.cloudstream3.utils.UIHelper.showInputMethod
...
+                    logError(error)
+                    return@register showToast(R.string.edit_profile_image_error_invalid)
+                }
+                ImageLoader(context).enqueue(ImageRequest.Builder(context).data(uri)
+                    .allowHardware(false).size(512, 512).listener(
+                        onSuccess = { _, _ -> callback(uri.toString()) },
+                        onError = { _, _ -> showToast(R.string.edit_profile_image_error_invalid) }
+                    ).build())
+            }
+            launcher.launch(arrayOf("image/png", "image/jpeg"))
```
fire-light42: This should be `"image/*"` to support all types of images.

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
```diff
@@ -189,8 +225,7 @@ object AccountHelper {
                         .allowHardware(false)
                         .listener(
                             onSuccess = { _, _ ->
-                                currentEditAccount = currentEditAccount.copy(customImage = url)
-                                binding.accountImage.loadImage(url)
+                                callback(url)
```
fire-light42: Elegant solution, very nice!

app/src/main/java/com/lagradost/cloudstream3/utils/DataStoreHelper.kt
```diff
@@ -169,7 +169,10 @@ object DataStoreHelper {
         @JsonProperty("lockPin") @SerialName("lockPin") val lockPin: String? = null,
     ) {
         @get:JsonIgnore
-        val image get() = customImage?.let { UiImage.Image(it) } ?:
+        val image get() = customImage?.takeUnless { image ->
+            image.startsWith("content://") && context?.contentResolver?.persistedUriPermissions
```
fire-light42, line 173: The image should preferably be copied to an internal directory to keep it in the app even if the user renames or moves the file.

However, this feature is scope creep and you may skip this if you feel like it would be too difficult to do in a clean manner.

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
```diff
@@ -216,6 +251,30 @@ object AccountHelper {
                 }
             }
         }
+
+        binding.editProfilePhotoButton.setOnClickListener {
+            AlertDialog.Builder(context)
+                .setTitle(R.string.edit_profile_image_title)
+                .setItems(arrayOf(
+                    context.getString(R.string.player_load_subtitles),
```
fire-light42: Please add another string in strings.xml, which references `player_load_subtitles` (like `action_open_play`). This makes it easier to prevent mistakes in the future, for example if `player_load_subtitles` is changed to something like "Load subtitle from file".

app/src/main/java/com/lagradost/cloudstream3/ui/account/AccountHelper.kt
```diff
@@ -216,6 +251,30 @@ object AccountHelper {
...
+                    context.getString(R.string.edit_profile_image_hint),
+                )) { _, selection ->
+                    if (selection == 0) {
+                        pickProfileImage(context) { image ->
+                            if (!dialog.isShowing) return@pickProfileImage
+                            currentEditAccount = currentEditAccount.copy(customImage = image)
+                            binding.accountImage.loadImage(image)
+                            showToast(R.string.edit_profile_image_success, Toast.LENGTH_SHORT)
```
fire-light42: This toast should be moved to pickProfileImage for consistency.

fire-light42 (changes requested):
Thank you for you contribution! Great job on the new feature!

Overall good code, but there are some minor nitpicks.

arpitagarwal1301:
Addressed the requested Activity Result API, image/* MIME, string alias, and toast placement changes in f59858c. I kept internal file copying out of scope as noted optional; persisted document permissions and the existing fallback remain in place. Ready for re-review when convenient.

fire-light42 (changes requested):
Great job!

Unfortunately passing around a variable to 100 subclasses is a bit too messy, but I realize you did it for a lack of a better solution, there is no good solution after all. 

I would really want you to register one global variable in CommonActivity like this: https://github.com/recloudstream/cloudstream/blob/master/app/src/main/java/com/lagradost/cloudstream3/CommonActivity.kt#L254-L266
This is how we solved a similar problem previously, I failed to send it though :(

It should be a pretty easy change thankfully. This way we do not have to pass around the variable everywhere. After these changes it should be fully done and ready to merge. Sorry for the added difficulties 😢

arpitagarwal1301:
Implemented the global picker in `CommonActivity` on the current head `49fef7a`, as requested. All review threads are resolved and CI is passing. Ready for re-review when convenient.

fire-light42:
You almost had it perfect, but the latest commit simply has too much confusing boilerplate code. Please see https://github.com/arpitagarwal1301/cloudstream/pull/1 for my proposed fixes.

PR #3104, opened 2026-08-02
feat(extractors): add vidara mirrors and fix links with filecodes

Bnyro:
URL to test: https://nameitweb.com/v/xkNOuHSrCNMj?24

(works for all other mirrors to if you just replace the base URL)


fire-light42 (approved):
Excellent!

PR #3107, opened 2026-08-03
FIX: Vidmoly Extractor

okibcn:
### BACKGROUND

Vidmoly wasn't working because the address of the iframe where the JWP script was extracted is now different. It also had a problem with single quotes in the JSON.

### CHANGES
1. newUrl is now generated from the vidmolyId
2. Script now doesn't have single quotes and the JWPlayerHelper can properly extract the m3u8 url.

### VERIFICATION
Tested with multiple Vidmoly links.

### DIFF
--- a/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidmoly.kt
+++ b/library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidmoly.kt
@@ -7,28 +7,26 @@ import com.lagradost.cloudstream3.extractors.helper.JwPlayerHelper
 import com.lagradost.cloudstream3.utils.ExtractorApi
 import com.lagradost.cloudstream3.utils.ExtractorLink
 
-@Prerelease
 class Vidmolyme : Vidmoly() {
     override val mainUrl = "https://vidmoly.me"
 }
 
-@Prerelease
 class Vidmolyto : Vidmoly() {
     override val mainUrl = "https://vidmoly.to"
 }
 
-@Prerelease
 class Vidmolybiz : Vidmoly() {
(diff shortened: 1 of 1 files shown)


okibcn:
Debug Build available at https://github.com/recloudstream/cloudstream/actions/runs/30785954728/artifacts/8845314112

library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/Vidmoly.kt
```diff
@@ -35,14 +35,13 @@ open class Vidmoly : ExtractorApi() {
             "Sec-Fetch-Dest" to "iframe"
         )
         
-        val newUrl = if (url.contains("/w/")) 
-            url.replaceFirst("/w/", "/embed-") + ".html" 
-            else url
+        val vidmolyId=url.removeSuffix("/").substringAfterLast("/")
+        val newUrl ="https://vidmoly.biz/embed-${vidmolyId}.html"
```
fire-light42: Please use mainUrl instead of a hardcoded url.

fire-light42 (changes requested):

okibcn:
@fire-light42 this PR is waiting for your approval. The current Extractor doesn't work. This new one works. VidMoly is used in many sites. I am supplying the extractor in my extensions, but it would be fine having it properly working in the core project.

fire-light42:
The issue of a hardcoded url remains: `val downloadUrl = "https://vidmoly.biz`. Is it neccessary to have specifically .biz to keep it working? Otherwise please use mainUrl to keep the other extractors resilient from domain switches.

duplexspy:
Thnks I would like that

PR #3122, opened 2026-08-14
feat(tv): add plugin settings button to home top bar

RjBiermann:
Adds a plugin settings button to the TV home page top bar (the home_api_holder group with plugin dropdown, reload, quick search, random).

- New `home_preview_settings_button` in `fragment_home_tv.xml`, placed right after the plugin dropdown button, with the same focus-chain pattern as the other icons
- Visibility driven by the active plugin: shown only when the active provider's plugin exposes `openSettings`; hidden via `isGone` (matches sibling buttons, no dead layout slot)
- Click resolves the active plugin from `apiName` via `APIHolder.getApiFromNameNull` and invokes `openSettings`
- `fragment_home.xml` gets a 0dp placeholder so `FragmentHomeBinding` exposes the field (phone stays unaffected)

TV/emulator only: the button is `isLayout(TV or EMULATOR)` gated for visibility; phone layout keeps a hidden 0dp placeholder.

## AI usage

This PR was developed with AI assistance (opencode). AI usage included design brainstorming, implementation, and testing on TV and phone emulators. The code was built, linted, and verified on the emulators before opening this PR. I can explain every line and fix any issues found in review.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeFragment.kt
@@ -811,10 +811,10 @@ class HomeFragment : BaseFragment<FragmentHomeBinding>(
                 homeChangeApi.text = apiName
                 homePreviewReloadProvider.isGone = (apiName == noneApi.name)
                 homePreviewSearchButton.isGone = (apiName == noneApi.name)
-                if (isLayout(TV)) {
+                if (isLayout(TV or EMULATOR)) {
                     val plugin = APIHolder.getApiFromNameNull(apiName)
                         ?.sourcePlugin?.let { PluginManager.plugins[it] } as? Plugin
-                    homePreviewSettingsButton.isVisible = plugin?.openSettings != null
+                    homePreviewSettingsButton.isGone = plugin?.openSettings == null
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


RjBiermann:
image
image
image

fire-light42 (changes requested):
Great pull request!

However, it looks like you are running into a race condition when determining the visibility of homePreviewSettingsButton. Some plugins can be slow to set `openSettings` which leads to an inconsistent settings symbol visibility.

I recommend moving the visibility logic to page load instead. UI element pop-ins are acceptable when the whole UI pops in and it will give extensions more time to set their `openSettings` variable.

RjBiermann:
Addressed the review (race condition): moved the settings button visibility check out of the apiName observer into the page load Success branch, so it's applied when the whole UI pops in and plugins have time to set openSettings. Pushed to the branch as 1d983201.

RjBiermann:
> Addressed the review (race condition): moved the settings button visibility check out of the apiName observer into the page load Success branch, so it's applied when the whole UI pops in and plugins have time to set openSettings. Pushed to the branch as [1d98320](https://github.com/recloudstream/cloudstream/commit/1d9832013ccc91170d1257fc492de1063aa607c4).

tested. now button shows after the homepage loads.

fire-light42 (approved):

PR #3126, opened 2026-08-16
fix(tv): re-enable navigation_home nextFocusRight so D-pad can reach the 'None' provider button

### LINKED ISSUE #2592: Cannot select "none" button with D-pad on google TV
### Steps to reproduce

1. in home page when no server is selected.
2. the cursor cannot touch the "None" button and view the server list
3. already use arrow up and down on D-Pad Botton but no hit the "None" button

Image

### Expected behavior

the D-Pad can reach the "None" button

### Actual behavior

the D-Pad cannot reach the "None" button on the home page

4.7.0

### Android version

Android 11

### Logcat

```java

```

### Other details

_No response_

### Acknowledgements

RjBiermann:
Fixes #2592

## Problem
On TV, the D-pad cannot reach the "None" / provider-selector button (`home_change_api`) on the home screen. After navigating away from Home and back (via Search/Favorites, the BACK button, or even a fresh load), pressing right from the nav rail does nothing — focus is stuck on the rail.

## Root cause
`onNavDestinationSelected` (`MainActivity.kt`) sets `nextFocusRightId` on all nav-rail buttons when navigating to Search/Favorites/Downloads (pointing at e.g. `main_search`), but the `navigation_home` case was commented out with a stale id. Two problems followed:
1. Returning Home never re-pointed the rail — it stayed aimed at the destroyed Search fragment's `main_search`, so right-focus from the rail was dead.
2. That hook only fires on rail-button navigation anyway: the BACK-button return path and first app load never ran it at all.

## Fix
Move the Home rail-focus handling into `addOnDestinationChangedListener` (`MainActivity.kt`), which fires on **every** arrival at Home regardless of how you got there:

```kotlin
if (navDestination.matchDestination(R.id.navigation_home)) {
    ...
    navRailView.nextFocusRightId = R.id.home_change_api
    // ... and for each nav rail button
}
```

so on every visit to Home, all rail buttons' `nextFocusRightId` point at the provider-selector button.

## Verification
Reproduced on the TV emulator (blank home, "None" selected) across all three arrival paths:
- Cold start: LEFT → RIGHT returns to "None"
- Search → BACK → Home: RIGHT reaches `home_change_api`
- Search → rail Home: RIGHT reaches `home_change_api`

Also verified with a plugin + home content loaded (StreamPlay): LEFT/RIGHT between rail and provider button works, and DOWN from the provider button drops into content as before.

## Notes
- The old commented mapping used a stale id (`home_preview_change_api`); the live id is `home_change_api`, which exists in both `fragment_home.xml` and `fragment_home_tv.xml`.
- Initial focus still lands on the provider button via `<requestFocus />` in `fragment_home_tv.xml`; this change additionally makes the rail→provider path work on every arrival.

AI note: this change was written with AI assistance per AI-POLICY.md; the logic was reproduced and verified on the TV emulator and every line is understood.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -763,7 +763,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
             // navigation to home for the second time as onNavDestinationSelected will not get called
             // when first loading up the app
 
-            R.id.navigation_home -> R.id.home_change_api
+            // R.id.navigation_home -> R.id.home_preview_change_api
             R.id.navigation_search -> R.id.main_search
             R.id.navigation_library -> R.id.main_search
             R.id.navigation_downloads -> R.id.download_appbar
@@ -1680,6 +1680,24 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
             if (navDestination.matchDestination(R.id.navigation_home)) {
                 attachBackPressedCallback("MainActivity") {
                     showConfirmExitDialog(settingsManager)
+                }
+                // Re-point nav rail focus at the provider button on every arrival
+                // (rail button, back stack pop, first load), so D-pad right always
+                // reaches the "None" / provider selector button.
+                if (isLayout(TV or EMULATOR)) {
+                    val fromView = binding?.navRailView
+                    if (fromView != null) {
+                        fromView.nextFocusRightId = R.id.home_change_api
+                        for (focusView in arrayOf(
+                            R.id.navigation_downloads,
+                            R.id.navigation_home,
+                            R.id.navigation_search,
+                            R.id.navigation_library,
+                            R.id.navigation_settings,
+                        )) {
+                            fromView.findViewById<View?>(focusView)?.nextFocusRightId = R.id.home_change_api
+                        }
+                    }
                 }
             } else detachBackPressedCallback("MainActivity")
         }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
```diff
@@ -1681,6 +1681,24 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
                 attachBackPressedCallback("MainActivity") {
                     showConfirmExitDialog(settingsManager)
                 }
+                // Re-point nav rail focus at the provider button on every arrival
+                // (rail button, back stack pop, first load), so D-pad right always
+                // reaches the "None" / provider selector button.
+                if (isLayout(TV or EMULATOR)) {
```
fire-light42: This pull request works and it really is an important bug fix, but right now it breaks D-pad behavior when pressing right on a populated home screen. It always changes focus to `home_change_api` even if `home_change_api` is not visible on the screen when scrolled down. 

I am not sure how to best solve this issue. I can not trade one bug for another. Please see if you can resolve this new bug.

fire-light42 (changes requested):
Great change. I would have merged if not for the new bug.

RjBiermann:
The scrolled-down case is fixed: the rail re-point now lives in HomeFragment instead of MainActivity, targeting home_change_api only when the home list is at scroll top. When scrolled, the rail buttons get no explicit right-target, so RIGHT from the rail uses default focus search and enters content, matching the old populated-home behavior. Triggers: the existing homeMasterRecycler onScrolled listener plus doOnLayout, which covers first load and returning to a home fragment recreated at a saved scroll offset. The MainActivity destination-listener block was reverted.

RjBiermann:
> The scrolled-down case is fixed: the rail re-point now lives in HomeFragment instead of MainActivity, targeting home_change_api only when the home list is at scroll top. When scrolled, the rail buttons get no explicit right-target, so RIGHT from the rail uses default focus search and enters content, matching the old populated-home behavior. Triggers: the existing homeMasterRecycler onScrolled listener plus doOnLayout, which covers first load and returning to a home fragment recreated at a saved scroll offset. The MainActivity destination-listener block was reverted.

tested. when there is homepage loaded with plugin and has rows. the dpad doesn't focus on the `home_change_api` instead to any poster.

PR #3127, opened 2026-08-17
fix: group subtitles by language

refirestream:
- Group subtitle tracks by language name derived from IETF tag instead of raw originalName (e.g. "fr" and "French [SUB]" now land in the same "French" group)
- Fall back to "Unknown" group when language cannot be resolved
- Label tracks within a group by originalName instead of nameSuffix
- Filter subtitles by IETF tag match instead of name substring, matching primary subtag so regional variants (pt-br) match (pt)
- Keep subtitles with unresolvable language instead of dropping them (goes into the Unknown group)


refirestream:
Doing this pull request, I discovered that there is a setting to filter out "useless" languages in :
Settings > Player > Subtitles > "Filter by preferred media language".
I believe that should be enabled by default. What do you think ? 

People that have not selected the correct languages during setup might be annoyed though. If we enable the  "Filter by preferred media language" setting now, will it change the value for existing users ?

refirestream:
I tested the app manually and its working well

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1195,20 +1197,26 @@ class GeneratorPlayer : FullScreenPlayer() {
                     ArrayAdapter<Spanned>(ctx, R.layout.sort_bottom_single_choice)
                 subsArrayAdapter.add(ctx.getString(R.string.no_subtitles).html())
 
+                val unknownGroupName = ctx.getString(R.string.subtitles_group_unknown)
+                fun groupName(sub: SubtitleData): String {
+                    return fromTagToLanguageName(sub.getIETF_tag())?.takeIf { it.isNotBlank() }
```
fire-light42, line 1202: I would prefer if you save the results to a list instead of calling this function since:

1. Calling the function again creates an assumption that the function will always return the same value, which can lead to bugs when refactoring in the future.

2. I dislike calling potentially expensive functions unnecessarily.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
```diff
@@ -1232,8 +1240,8 @@ class GeneratorPlayer : FullScreenPlayer() {
                     val subtitleOptions =
                         subtitlesGroupedList
                             .getOrNull(subtitleGroupIndex - 1)?.value?.map { subtitle ->
-                                val nameSuffix = subtitle.nameSuffix.html()
-                                nameSuffix.ifBlank {
+                                val label = subtitle.originalName.html()
```
fire-light42, line 1243: This completely breaks the subtitle suffix system and makes it very confusing to precisely select between different subtitles of the same language. Is there any reason for this change?

fire-light42 (changes requested):
Good idea and minimal code changes, but needs some small changes.

fire-light42:
> Doing this pull request, I discovered that there is a setting to filter out "useless" languages in :
Settings > Player > Subtitles > "Filter by preferred media language".
I believe that should be enabled by default. What do you think ?

Changing default behavior needs a lot of motivation, since users have come to expect that this is how CloudStream works. I think it would be too confusing. I would rather recommend a more subtle and visible solution, for example always placing preferred languages at the top of subtitle languages. However, this is best done as a separate pull request.

> People that have not selected the correct languages during setup might be annoyed though. If we enable the "Filter by preferred media language" setting now, will it change the value for existing users ?

It would change the value for all users who have not previously changed the value, unless it is explicitly migrated using code.

app/src/main/java/com/lagradost/cloudstream3/ui/player/GeneratorPlayer.kt
Sarlay: You are right, there is an ambiguity when 2 subtitles have the same exact name that needs to be fixed.
I believe that their are 2 possibilities for the name:
1) Original name (+ number when duplicate) 
IMG_20260908_144933
This allows to display additional information for the subtitle track (distinguish forced subs, full subs, or audio description), but it is less standardized
2) Number only
IMG_20260908_151718
More standardized design but loss of information.


I am leaning more towards option 1, what is your opinion ?

refirestream: I implemented option 1 for now

Issue #2771, opened 2026-05-07
Alignment error in the upcoming post


ByAyzen:
### Steps to reproduce

Contents which is not released is showing with "coming soon" but it has an alignment problem.

### Expected behavior

It should be higher than actors.

### Actual behavior

Its under the actors, which can cause users couldnt see it

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/98077585-b164-441c-98af-e17b9a72186b" />

### Cloudstream version and commit hash

Prerelease latest

### Android version

Android 14

### Logcat

```java

```

### Acknowledgements

- [x] I am sure my issue is related to the app and **NOT some extension**.
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [x] I will fill out all of the requested information in this form.

PR #3136, opened 2026-08-25
Fix: Upcoming post alignment issue

ByAyzen:
**Make upcoming text balanced in center. Fixing and closing** **#2771** 

It might not be visible immediately, but when users scroll down to the description, it is now clearly visible. I also thought placing it above the actors and genres, but it looked really bad.

**Before**

resim

**After**

resim


fire-light42 (approved):
Thank you!

Issue #2376, opened 2025-12-25
Open in browser


ByAyzen:
### Steps to reproduce

Click on extension
Extension had some issue when loading so it gave error
Click on "open in browser"

### Expected behavior

It should directly open the website

### Actual behavior

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/ddc8e00f-2efa-4104-8133-d81a18570b12" />

This menu shows up.

### Cloudstream version and commit hash

4.6.1 Prerelease/Latest

### Android version

A13 Tiramisu

### Logcat

```java
No need for logcat.
```

### Acknowledgements

- [x] I am sure my issue is related to the app and **NOT some extension**.
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [x] I will fill out all of the requested information in this form.

ByAyzen:
Btw its not listed by alphabetical thats why its a complete mess.

rockhero1234:
I agree

Abodabodd:
> ### Steps to reproduce
> Click on extension Extension had some issue when loading so it gave error Click on "open in browser"
> 
> ### Expected behavior
> It should directly open the website
> 
> ### Actual behavior
[...]

PR #3137, opened 2026-08-25
Feat: Open in browser is now directing immedialetly.

ByAyzen:
**Fixing and changing #2376** 

When an error occurred on the home page, clicking "Open in browser" previously opened an unorganized and non-alphabetical menu listing all extensions instead of directly opening the website. With this PR it directly opens the current extensions website in the browser.

Note: I wasn't sure about completely removing the menu. I thought keeping it and sorting it alphabetically, but decided to remove it for a simple user experience. Alternatively i can revert these changes and just sort the menu alphabetically.


fire-light42 (approved):
Excellent! I have been thinking about this, but never got to implementing it. Thank you!

PR #3140, opened 2026-08-26
Fix: Search menu focus and suggestions

ByAyzen:
**I tried to fix 2 problems which is happen on TV UI.**

**1.Suggestions**

When suggestions appear, the user closes them, opens content and comes back, but suggestions show up again.

**2.Focus**

When a user searches for content and presses the down button to focus on the poster, focus jumps to the left sidebar instead. Similarly, pressing the right button from the left sidebar to reach the poster sends the focus back up to the search bar. 

**Video for Focus;**

https://github.com/user-attachments/assets/2c10183d-e987-4bdc-8c0e-6f7b321ee7bd

**Screenshot for Suggestions (It also happend to me so its a very common problem.);**

resim

**AI Usage:**

I used AI Guidance for focus issue.

**Tests;**

I tested these two problems with the prerelease and my debug version. All problems are fixed in the debug version. If someone wants to reproduce them, suggestions can easily be reproduced, but search might not be.

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/MainActivity.kt
@@ -764,7 +764,7 @@ class MainActivity : AppCompatActivity(), ColorPickerDialogListener, BiometricCa
             // when first loading up the app
 
             // R.id.navigation_home -> R.id.home_preview_change_api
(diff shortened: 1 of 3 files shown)


ByAyzen:
I'm closing this PR because focus issues will likely be fixed with compose redesign. Suggestion fix will be here #3202

PR #3143, opened 2026-08-27
feat: update home banner size and match its shimmer

iAm-an-iA:
# Summary
Fixes the initial loading skeleton on the Home screen to properly match the full-bleed Hero Banner redesign instead of the legacy 3-item carousel skeleton. This is crucial for improving user experience by eliminating jarring layout shifts and visual flickering, especially when content is cached or loads instantly—ensuring a seamless and smooth visual transition on app launch and provider switching.

### TV
Screen_recording_20260828_151920

### Phone
Screen_recording_20260828_152100

### DIFF
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapterPreview.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapterPreview.kt
@@ -655,7 +655,7 @@ class HomeParentItemAdapterPreview(
         }
 
         private fun updatePreview(preview: Resource<Pair<Boolean, List<LoadResponse>>>) {
-            if (preview is Resource.Success) {
+            if (preview is Resource.Success || isLayout(TV or EMULATOR)) {
                 homeNonePadding.apply {
                     val params = layoutParams
                     params.height = 0
--- a/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapterPreview.kt
+++ b/app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapterPreview.kt
@@ -310,6 +310,9 @@ class HomeParentItemAdapterPreview(
         private val previewViewpager: ViewPager2 =
             itemView.findViewById(R.id.home_preview_viewpager)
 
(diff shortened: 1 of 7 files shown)


iAm-an-iA:
Now that the catalog loads ahead of the hero banner, we have a chance to improve UX in Android TV by shrinking both the banner and its shimmer just enough to expose the first catalog row immediately. Does adjusting the height make sense here, or do we prefer keeping the full-bleed banner as is?

iAm-an-iA:
https://github.com/user-attachments/assets/fe166748-36f3-4205-af50-3aab7d400709

Shimmer + Lazy Banner + Provider Cache example

app/src/main/res/layout/fragment_home_head_tv.xml
```diff
@@ -34,7 +34,7 @@
             <androidx.cardview.widget.CardView
                 android:id="@+id/home_preview_info_btt"
                 android:layout_width="match_parent"
-                android:layout_height="400dp"
+                android:layout_height="260dp"
```
fire-light42: Please keep UI changes in a separate pull request.

app/src/main/java/com/lagradost/cloudstream3/ui/home/HomeParentItemAdapterPreview.kt
```diff
@@ -652,7 +655,7 @@ class HomeParentItemAdapterPreview(
         }
 
         private fun updatePreview(preview: Resource<Pair<Boolean, List<LoadResponse>>>) {
-            if (preview is Resource.Success) {
+            if (preview is Resource.Success || isLayout(TV or EMULATOR)) {
```
fire-light42: Better to keep this logic offloaded to `getStatusBarHeight` rather than define the behavior in both places.

fire-light42 (changes requested):
Everything looks good, except for the regular UI changes.

It is better to split unrelated features into multiple pull requests. Otherwise discussions about one feature may block the other feature. It also makes it more difficult to review.

app/src/main/res/layout/fragment_home_head_tv.xml
iAm-an-iA: I've reverted the TV banner so this PR stays purely focused on the shimmer updates

PR #3144, opened 2026-08-27
Feat: Horizontal posters for search and recommendations.

ByAyzen:
Since horizontal posters were only available on the main page, I wanted to bring this support to the search and recommendations sections as well. To not break multi search, I disabled it there and only applied it to extension search. I also used @Prerelease tag to be sure it doesn't break the stable version.

**Parameters for implementing to extensions:**

```
@Prerelease
open val hasHorizontalSearch: Boolean = false
@Prerelease
open val hasHorizontalRecommendations: Boolean = false
```

I'm not completely sure about these two parameters because they can feel confusing; I can agree to change them. I could place them directly into search and new*****LoadResponse similar to how mainpage does it.

**AI Usage**
 
I used AI for implementing this idea. After that, I checked and cleaned up the issues.

**Tests**

I tested on both prerelease and debug versions. Tested for 2 days and didn't have any problems.

**Two comparison images below;

[Search](https://imgslider.com/17407c22-6bf4-4006-b5c2-6de91ffd4686)
[Recommendation](https://imgslider.com/e015e703-5404-454a-9c9c-d86bf8147227)


library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -626,6 +633,15 @@ abstract class MainAPI {
     open val vpnStatus = VPNStatus.None
     open val providerType = ProviderType.DirectProvider
 
+
+    /**
+     * These horizontal options are only affecting extension search and load recommendations.
+     * */
+    @Prerelease
+    open val hasHorizontalSearch: Boolean = false
```
fire-light42: These should rather be a property of [SearchResponseList](https://github.com/recloudstream/cloudstream/blob/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt#L1292-L1297) akin to [HomePageList](https://github.com/recloudstream/cloudstream/blob/master/library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt#L1285), to be consistent and allow for flexible results.

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
```diff
@@ -138,6 +138,13 @@ object APIHolder {
         initMap(true)
     }
 
+    fun isApiHorizontal(name: String?): Boolean {
```
fire-light42, line 142: Do not forget to also annotate new functions with `@Prerelease`. The best way to spot compatibility issues is by looking at library.api to see all ABI changes.

fire-light42 (changes requested):
Good feature, but needs more `@Prerelease` annotations and a small change in MainAPI 👍

library/src/commonMain/kotlin/com/lagradost/cloudstream3/MainAPI.kt
fire-light42: This can be made backwards compatible since it uses a builder function.

ByAyzen: I wasn't quite sure about these two properties either. Thanks for the feedback.

Issue #2665, opened 2026-04-15
The clear button in the download queue doesn't work.


AkshayKanathur:
### Steps to reproduce

1. Start downloading something
2. Open the Download queue
3. Press the clear button at the top right 

### Expected behavior

The download queue should be cleared. 

### Actual behavior

The download queue doesn't clear.

https://github.com/user-attachments/assets/9276da41-ae8e-40ab-8f91-08e174e29316

### Cloudstream version and commit hash

4.7.0-PRE c31c576

### Android version

Android 12

### Logcat

```java

```

### Acknowledgements

- [x] I am sure my issue is related to the app and **NOT some extension**.
- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to pre-release version **[Latest](https://github.com/recloudstream/cloudstream/releases)**.
- [x] I will fill out all of the requested information in this form.

PR #3147, opened 2026-08-29
Fix: Cancel and clear downloads.

ByAyzen:
**Small fix.** Fixing this issue #2665


fire-light42:
I think this issue should rather be fixed by fixing the core problem which causes downloads to be stuck. This is sadly a bandage solution and I do not want users to accidentally cancel their active downloads.

ByAyzen:
> I think this issue should rather be fixed by fixing the core problem which causes downloads to be stuck. This is sadly a bandage solution and I do not want users to accidentally cancel their active downloads.

Agreed. Now it will only delete paused downloads.

ByAyzen:
My last commit was unnecessary. I tried to fix some stuck download issue, but apparently it wasn't an issue, it was just for a minute. Prerelease had no problem like that, so I reverted these changes.

PR #3158, opened 2026-09-06
feat(player): implement dual subtitles, sync offset, and side-by-side comparison viewer

nutine:
Adds dual subtitle support to the player:

- Long-pressing any subtitle in the selector marks it as secondary (`[2nd]`), which renders at the top of the screen while the primary subtitle stays at the bottom.
- Custom `DualDefaultTrackSelector` so two embedded tracks (or embedded + external) can be decoded simultaneously in ExoPlayer.
- Separate sync/delay offset controls for the secondary subtitle in the sync dialog.
- Side-by-side comparison dialog (opened by long-pressing the top-right corner of the player) that pairs both subtitle timelines together. Tapping any line seeks to that timestamp. Playback auto-resumes on dialog close.
- Holding the right side of the screen pauses and reveals the secondary subtitle if not set to always visible.
- Setting in player preferences to always keep secondary subtitles visible.

Tested on phone and Android TV with both external URL subtitles (.srt / .vtt) and embedded MKV/MP4 tracks.

---
*Note on AI usage (per AI Policy): An LLM was used for assistance with boilerplate and initial drafting, but the implementation, track selector logic, and player integration were refined and tested manually on device.*


fire-light42:
This is *a lot* of different changes. Please split them into separate pull requests which can be reviewed and merged separately. Not every single feature needs to be separate, but dual subtitles, general UI improvements and AI translations all require their own independent review.

nutine:
Thanks for the feedback! I have scoped this pull request down strictly to **Dual Subtitles & Comparison Viewer**. I am opening separate, independent pull requests for the general UI/gesture improvements and the AI subtitle translation feature.

PR #3160, opened 2026-09-08
Make the sync subtitle delay feature intuitive

refirestream:
I've always felt like this subtitle delay feature was counter intuitive, the + button on the right should actually make the white slider go forward (meaning that the subtitle was lagging too much behind the audio), as opposed to what is currently implemented.
Screenshot_20260908-165702_CloudStream Debug
This PR implements that.


fire-light42 (changes requested):
Great feature! I do agree that it is much more intuitive now! 

However it seems like this pull request accidentally included additional code which changes subtitle groupings

fire-light42 (approved):

refirestream:
Thanks for merging ! And sorry about that mix-up with the accidentally included code

PR #3169, opened 2026-09-12
Add preference to start videos paused

BlipBlob:
off by default


fire-light42 (approved):
I do not want to add unnecessary settings, but this is something I might use myself 😀

PR #3172, opened 2026-09-13
Added compose subtitle settings

LagradOst:


LagradOst:
The main work here is to re-create the color picker we had from a lib. While not being 1:1 it should be expressive enough so ppl dont complain. If they do I guess we have to re-add the color-wheel. However that would be a pain for TV-support ngl.

LagradOst:
Another main issue is that we (somehow?) used fontres for fonts!!! and serialized it!!! How in the hell did ppl not crash?

PR #3175, opened 2026-09-13
feat(player): add two-finger subtitle jump gestures

nutine:
Adds two-finger gestures to skip between subtitle lines during playback:

- Two-finger tap on the left or right side jumps to the previous or next subtitle cue.
- Two-finger horizontal swipe as an alternative way to jump.
- Shows a brief HUD snippet with the sentence text when jumping.
- If playback is more than 1 second into the active line, rewinding jumps back to the beginning of the current sentence instead of skipping to the previous one.

Makes dialogue navigation quick when watching subtitled shows without having to bring up the full UI controls.

Tested on a physical phone (Android 13).

---
*Note on AI usage (per AI Policy): An LLM was used to assist with boilerplate and initial drafting, but the gesture logic and player integration were written, tuned, and tested manually on hardware.*


Bnyro:
Please read the [AI Policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md). You didn't even write the PR description yourself...

nutine:
Fair point, sorry about the generic template format. I've updated the description in my own words and added the AI disclosure per the policy.

The code itself is straightforward gesture detection in `PlayerGestureHelper` and subtitle cue seeking in `FullScreenPlayer` — it just lets you jump between subtitle lines with two fingers and shows a quick HUD snippet. Tested and tuned on physical hardware so it doesn't conflict with pinch-to-zoom or seek swipes.

Let me know if you have any feedback on the implementation or if anything should be adjusted.

PR #3176, opened 2026-09-13
feat(subtitles): implement subtitle translation proxy support

nutine:
Adds support for on-the-fly subtitle translation using an external translation proxy server.

Instead of embedding heavy translation logic, vendor APIs, or keys into the app, this keeps the client completely decoupled:
- Users can set a translation proxy URL in Player Subtitles settings.
- Selecting "Translate subtitles" in the player lets you pick a target language.
- The app builds a proxied URL (`<proxy>/translate?url=<sub_url>&target=<lang>`) and passes it directly to the player as a standard WebVTT subtitle track.
- The proxy server downloads the subtitle, translates it, and streams back WebVTT.

An open-source reference proxy implementation is available here:
https://github.com/nutine/cloudstream-subtitle-proxy

Tested with local and remote proxy servers on Android 13.

---
*AI Policy disclosure: An LLM was used to assist in drafting boilerplate, but the architecture, code refactoring, and integration were implemented, verified, and tested manually.*


Bnyro:
Please read the [AI Policy](https://github.com/recloudstream/cloudstream/blob/master/AI-POLICY.md). You didn't even write the PR description yourself...

I also don't think that this should be implemented directly on the client, that looks extremely ugly. The only solid way to implement this if there's a dedicated server / proxy that can translate subtitles so that the proxied URL can just be passed to the player instead of the original subtitle URL.

nutine:
Thanks for the review and perspective @Bnyro. You make a solid point about keeping the player client clean — handling translation on a dedicated proxy/server makes much more sense architecturally than baking LLM API calls directly into the app. Closing this PR. Also apologies for the AI-templated description.

nutine:
You're completely right about decoupling the client from direct LLM logic — putting vendor API calls and prompt batching directly into the Android app was messy.

I've refactored this PR according to your suggestion:
1. The client now only knows about a configurable subtitle translation proxy URL (in Subtitles settings).
2. When a user requests translation in the player, it simply builds the proxied URL (`<proxy>/translate?url=<sub_url>&target=<lang>`) and hands it to `addAndSelectSubtitles` as a regular WebVTT track.
3. The client diff is now minimal (~140 lines total, no LLM SDKs, no batching or parsing hacks in the player).
4. I've also published a reference open-source proxy implementation here that anyone can run/host: https://github.com/nutine/cloudstream-subtitle-proxy

Also updated the description and added the AI disclosure per the policy. Let me know if this looks cleaner!

PR #3179, opened 2026-09-14
feat(extractors): support stream mirror selection and add firestream.site domain

Bnyro:
e.g. https://firestream.site/e/XYjWUQDl?0001 has 0001 as query arg so the stream link extraction would fail before these changes.


Bnyro:
Could somebody please help with the Prerelease annotation?

I had to change the class signature (made it open to inheritance), so I annotated it with Prerelease, but apparently that's not right either. No clue how to fix it.

same in https://github.com/recloudstream/cloudstream/pull/3180

fire-light42:
There is no perfect solution for making classes open, but the main goal of the ABI check is to keep the library backwards compatible. Making a class open does break any existing extensions, but can cause future breakage. It is an acceptable compromise.

PR #3183, opened 2026-09-14
Fixed slow regex

LagradOst:


LagradOst:
From profiling I saw that just constructing the languages we want to select took wayyyy too long. This was caused in part due to regex and also sorting. Now subtitle fragment first few frames should be less laggy on open.

PR #3201, opened 2026-09-20
Fixed some settings bugs

LagradOst:


LagradOst:
This was tested on weird tv devices and emulator, but it looks like the keyboard appearing bug is hard to fix before navigation is also in compose.

PR #3205, opened 2026-09-21
feat(extractors): add StreamRuby extractor

mehdigm4life:
Adds support for the StreamRuby / rubyvidhub.com embed hosts, which are used as a video source by a number of streaming sites.

The extractor:
- fetches the embed page first to seed the session cookie required by the site
- posts to the /dl endpoint to get the player script
- unpacks the packed JS and pulls the HLS stream out of the jwplayer sources

Also ships a StreamRubyCom variant that targets the streamruby.com mirror domain. The file code is parsed so that both /e/{code} and /embed-{code}.html URLs work.


mehdigm4life:
Updated the library abi dump to match the JDK 17 toolchain used by CI. The previous dump had a few JDK 21-only List methods (addFirst/addLast/removeFirst/removeLast) that the JDK 17 build doesn't emit, which made checkKotlinAbi fail. The new run may need reviewers to approve the workflow run before it starts.

PR #3206, opened 2026-09-21
feat(extractors): add StreamRuby extractor

mehdigm4life:
Adds support for the StreamRuby / rubyvidhub.com embed hosts, which are used as a video source by a number of streaming sites.

The extractor:
- fetches the embed page first to seed the session cookie required by the site
- posts to the /dl endpoint to get the player script
- unpacks the packed JS and pulls the HLS stream out of the jwplayer sources

Also ships a StreamRubyCom variant that targets the streamruby.com mirror domain. The file code is parsed so that both /e/{code} and /embed-{code}.html URLs work.

Includes a library abi dump refresh so the binary compatibility check matches the JDK 17 toolchain used by CI.


library/src/commonMain/kotlin/com/lagradost/cloudstream3/extractors/StreamRuby.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.lagradost.cloudstream3.utils.Qualities.Unknown
+import com.lagradost.cloudstream3.utils.newExtractorLink
+
+@Prerelease
+class StreamRubyCom : StreamRuby() {
+    override var mainUrl = "https://streamruby.com"
+}
+
+@Prerelease
+open class StreamRuby : ExtractorApi() {
```
fire-light42: This looks like the same logic as LuluStream, please inherit from that one if possible

fire-light42:
No, I mean literally:

```kt
class StreamRuby : LuluStream() {
    override val name = "StreamRuby"
    override val mainUrl = "https://rubyvidhub.com"
}
```

Does that not work?

mehdigm4life:
It can't be a plain subclass because the site needs more than LuluStream's flow: the filecode comes from URLs like /embed-{code}.html (and /e/{code}), the embed page has to be fetched first to seed the session cookie before posting to /dl, and the /dl response is packed JS that has to be unpacked before the stream URL can be pulled out. That's why the extractor overrides just those three steps while reusing LuluStream's overall flow (getFileCode → getPlayerScript → parsePlayerScript).

Happy to drop it down to a plain subclass if it turns out the plain flow works on the live sites.

fire-light42:
Okay I understand :+1: 

Please share a demo video link for testing and to see if any other extractor implements the required logic

mehdigm4life:
Here are some working demo links to test with — they cover both URL formats the extractor parses:

https://streamruby.com/e/sp03sbfzp0tc (filecode sp03sbfzp0tc)
https://streamruby.com/embed-sp03sbfzp0tc.html

Both work fine, and the extractor grabs the HLS stream correctly.

mehdigm4life:
@fire-light42

